PR #1, opened 2015-11-26
Init commit of travis config and scripts

130s:
See https://github.com/ros-industrial/ros_industrial_meetings/issues/1 for general idea for this repo.

Usage is summarized in [README.md](https://github.com/130s/industrial_ci/blob/add/travisconf/README.md). Hope it would answer all questions, but if not please let me know (maybe the content of the tests need documented?).

Once this PR gets merged, I'll open another PR in a **client** repository (I'm thinking of industrial_core but it can be any other repository).


130s:
See https://github.com/ros-industrial/ros_industrial_meetings/issues/1 for general idea for this repo.

Usage is summarized in [README.md](https://github.com/130s/industrial_ci/blob/add/travisconf/README.md). Hope it would answer all questions, but if not please let me know (maybe the content of the tests need documented?).

Once this PR gets merged, I'll open another PR in a **client** repository (I'm thinking of industrial_core but it can be any other repository).

check_metapackage.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Author: Yuki Furuta <furushchev@jsk.imi.i.u-tokyo.ac.jp>
```
shaun-edwards: This file should contain a license header.  Since the author is not @130s, we must make sure we use the whatever license file is with the original file.  I realize this is a trivial source file, but at the very least we should give credit where credit is due.

travis.sh
```diff
@@ -0,0 +1,243 @@
+#!/bin/bash
+
+## This is a "common" script that can be run on travis CI at a downstream github repository.
+## 
+## Input
+##  .travis.rosintall: A .rosintall file that's used to download the resource of the dependended packages that need to be built from source.
+## 
+## Greatly inspired by JSK travis with additional comments for maintainability: https://github.com/jsk-ros-pkg/jsk_travis 
+## Author: Isaac I. Y. Saito (TORK)
```
shaun-edwards: Does anybody know if we should include licenses in bash scripts?

shaun-edwards:
Just a couple of license questions.  I'm sure I'll have some additional comments once we start using travis, but for now I'm good to merge this.  +1 after the license issues are discussed.

@gavanderhoorn, do you want to have a look?

check_metapackage.py
130s: All I can tell from the publicly available resource is the [repository the original file is copied from is BSD](https://github.com/jsk-ros-pkg/jsk_travis/blob/1dc06eea19e9ac479c786066c92423eaf4c9ce04/package.xml#L5), and the name of its organization.

For BSD, looks like it's better to add license and copyrights header per each file ([referred to this](http://stackoverflow.com/a/3053725/577001)).

Therefore I added the license (BSD) and copyright header (under U-Tokyo).

travis.sh
130s: Not an answer but if we use BSD, then we might as well [add license and copyrights header per file](http://stackoverflow.com/a/3053725/577001).

Jmeyer1292:
Tagging this discussion, as I'm interested in updating Descartes as soon as its ready.

130s:
Updating commits unfortunately hid the comments from @shaun-edwards so if anyone is interested please expand the `outdated diff`above.

Issue #6, opened 2015-12-17
Update README to "git clone" workflow


130s:
Mentioned at https://github.com/ros-industrial/industrial_ci/issues/3#issuecomment-165312102.

130s:
Continue discussion in #9.

PR #9, opened 2015-12-24
Update readme (addresses #6)

130s:
Commit messages should be self-explanatory. Copying them for convenience:

- Explain env variables 	d3870de
- Replace "git submodule" with "git clone". 	25328d2
- Add example pkgs 004f1f2


shaun-edwards:
+1

PR #10, opened 2015-12-29
[doc] Elaborate env vars and what are tested and what can be empty.

130s:


shaun-edwards:
+1...thanks for the detailed docs...this helps a lot!

ipa-mdl:
> I would be curious how you achieve to run the tests against the install space. Can you describe this a little bit. 

All installed *.test files in configured downstream packages get executed with rostest:
https://github.com/ros-industrial/industrial_ci/blob/master/travis.sh#L238-L256

PR #12, opened 2016-01-01
changelog for 0.1.1

130s:


shaun-edwards:
@130s, sorry for the delay.  I assume this change is related to a bloom release?  In which case, we generally don't review "release artifacts" because this really slows down the release.  +1.

gavanderhoorn:
Are we actually bloom-releasing this?

130s:
>  I assume this change is related to a bloom release?

I thought it's worthwhile for a discussion. I'll open a ticket for it.

Whether or not bloom-release this repo, I want to tag/version snapshots of the moments; this allows users of this repo to use specific snapshot in a definite way if needed; pointing to a specific commit also works, but version is more palpable and less confusing IMO.

And as a bonus for this repo to be made as a catkin package, we can utilize helper tools (`catkin_prepare_release` and `catkin_generate_changelog`).

I do this with other non-release packages of mine at our team, and it works well. At least it gives us a peace of mind ;)

> we generally don't review "release artifacts" because this really slows down the release.

I got it. Then from next time I'll `catkin_generate_changelog` and `catkin_prepare_release` without opening PRs.

Issue #14, opened 2016-01-08
Allow custom setup rules


130s:
Currently in client repo you can add your own script (as documented) like this:
```
:
script: 
  - source .ci_config/travis.sh
  - source ./your_travis.sh
```

`./your_travis.sh` always runs once `.ci_config/travis.sh` finishes. With this you cannot customize the installation setup.

For example, when a device driver package X requires a prorietary library installed to run. This library is available online, but not via `apt` or through package management system and you need to install it in a classic way (unzip, run installer etc.) 
Even if you defined this manual installation in ./your_travis.sh, the library gets installed after `run_tests` and all that already ran.

130s:
Sorry, posted in the middle by mistake.

----

Currently in client repo you can add your own script (as documented) like this:
```
:
script: 
  - source .ci_config/travis.sh
  - source ./your_travis.sh
```

`./your_travis.sh` always runs once `.ci_config/travis.sh` finishes. With this you cannot customize the installation setup.

For example, when a device driver package X requires a prorietary library installed to run. This library is available online, but not via `apt` or through package management system and you need to install it in a classic way (unzip, run installer etc.) 
Even if you defined this manual installation in ./your_travis.sh, the library gets installed after `run_tests` and all that already ran.

130s:
Closed via #15

PR #15, opened 2016-01-08
Add how to call pre-process script (addresses #14).

130s:
To the question #14 I found that we actually don't need additional implementation. As added to the readme in this PR, we can simply call pre-processing script(s).

This PR also adds tests for the proposed concept; install a [`sl` command](http://unix.stackexchange.com/a/341/14968) in the pre-process, and calling it in the post-process is successful as you see [in this output](https://travis-ci.org/130s/industrial_ci/jobs/101186280#L3309) (because of Travis' time-elapsed view, the output looks broken but we can tell the program is installed).

Lastly, only as a reference, I've experimented another way to address #14; See https://github.com/130s/industrial_ci/pull/5 where `travis.sh` is modified to accept custom script if specified. I found it redundant and overkill.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -234,6 +234,24 @@ The jobs that run Prerelease Test may usually take longer than the tests defined
 
 Then open a pull request using this branch against the branch that the change is subject to be merged. You do not want to actually merge this branch no matter what the Travis result is. This branch is solely for Prerelease Test purpose.
 
+Run pre-install custom commands
+-----------------------------------------
+
+You may want to add custom steps prior to the setup defined in `./travis.sh <./travis.sh>`_. Example:
+
+* A device driver package X in your repository or in your repository's dependency requires a prorietary library installed. This library is publicly available, but not via apt or any package management system and thus the only way you can install it is in a classic way (unzip, run installer etc.) (`More discussion <<https://github.com/ros-industrial/industrial_ci/issues/14>`_).
+
+In this case, add `source`d scripts before `travis.sh` gets called (see below for an example). 
+
+::
+
+  script: 
+    - source ./your_custom_PREprocess.sh
+    - source .ci_config/travis.sh
+    - source ./your_custom_POSTprocess.sh
+
+In the above case, in both `.ci_config/travis.sh` and `your_custom_POSTprocess.sh` the environment is kept from previous script(s), so whatever is done in previous scripts remains. 
+
 For maintainers of industrial_ci repository
 ================================================
 
--- a/post-process.sh
+++ b/post-process.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2015, Isaac I. Y. Saito
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#       * Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#       * Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in the
+#       documentation and/or other materials provided with the distribution.
+#       * Neither the name of the Isaac I. Y. Saito, nor the names
+#       of its contributors may be used to endorse or promote products derived
+#       from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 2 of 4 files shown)


130s:
For some reason Travis isn't triggered. Re-triggering it by close-reopening.

130s:
Please review @shaun-edwards @gavanderhoorn.
Application example https://github.com/InstitutMaupertuis/institut_maupertuis_robots_descriptions/pull/1

README.rst
```diff
@@ -190,6 +190,23 @@ Maintainers of client repos are responsible for applying the changes that happen
 
 2. Don't forget to commit the changes the command above makes.
 
+Run pre-install custom commands
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+You may want to add custom steps prior to the setup defined in `./travis.sh <./travis.sh>`_. Example:
+
+* A device driver package X in your repository or in your repository's dependency requires a prorietary library installed. This library is publicly available, but not via apt or any package management system and thus the only way you can install it is in a classic way (unzip, run installer etc.) (`More discussion <<https://github.com/ros-industrial/industrial_ci/issues/14>`_)
```
gavanderhoorn: Is this line missing an ending 'dot'?

README.rst
```diff
@@ -190,6 +190,23 @@ Maintainers of client repos are responsible for applying the changes that happen
...
+
+In this case, add `source`d scripts before travis.sh gets called (see below for an example).
```
gavanderhoorn: travis.sh not monospaced (ie: missing backticks). More important to be consistent though, so if you haven't done that in the rest of the document, don't do it here.

README.rst
```diff
@@ -190,6 +190,23 @@ Maintainers of client repos are responsible for applying the changes that happen
...
+In this case, add `source`d scripts before travis.sh gets called (see below for an example). 
+
+::
+
+  script: 
+    - source ./your_custom_PREprocess.sh
+    - source .ci_config/travis.sh
+    - source ./your_custom_POSTprocess.sh
+
+In above case, in both `.ci_config/travis.sh` and `your_custom_POSTprocess.sh` the environment is kept from previous script(s), so whatever done in previous scripts remain.
```
gavanderhoorn: Only language (but not a native speaker myself): "In *the* above case, ..".

and: "whatever *is* done in previous scripts remain*s*".

gavanderhoorn:
This looks ok to me in principle.

One minor point would be that I would not expect a script that is *sourced* to then invoke commands that change the system (ie: `source`ing is typically used to update a shell's environment, not to run things like `apt-get`).

This may just be me though: the real idea behind `source` is to run a script in the current `bash` instance, so it seems perfectly acceptable use.

130s:
@gavanderhoorn thank you, all inline comments should be addressed.

If the usage of `source` could puzzle (especially) experienced linux users, we might want to change it to more appropriate command. 
But because calling the `.ci_config/travis.sh` by `source` is already recommended and I believe many client repos do that, changing how to call scripts from upstream repo needs to be done carefully. That's worth another ticket IMO.

gavanderhoorn:
> If the usage of `source` could puzzle (especially) experienced linux users, we might want to change it to more appropriate command.

No, as I wrote, it's really just me. `source` is fine.

README.rst
```diff
@@ -190,6 +190,23 @@ Maintainers of client repos are responsible for applying the changes that happen
...
+In this case, add `source`d scripts before `travis.sh` gets called (see below for an example). 
+
+::
+
+  script: 
+    - source ./your_custom_PREprocess.sh
+    - source .ci_config/travis.sh
+    - source ./your_custom_POSTprocess.sh
+
+In the above case, in both `.ci_config/travis.sh` and `your_custom_POSTprocess.sh` the environment is kept from previous script(s), so whatever is done in previous scripts remain.
```
gavanderhoorn: `remain` still missing an `s`?

post-process.sh
```diff
@@ -0,0 +1,37 @@
...
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+set -x
+
+echo "post_travis.sh BEGIN";
+echo $PATH;
+/usr/games/sl;  # For some reason, relative path becomes error on Travis.
```
gavanderhoorn: Might want to add that this is *just an example*, and that users should place their own commands here.

pre-process.sh
```diff
@@ -0,0 +1,39 @@
...
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+set -x
+
+echo "custom_setup.sh BEGIN";
+sudo apt-get update
+sudo apt-get install -y -q -qq sl
+echo $PATH;
+/usr/games/sl;  # For some reason, relative path becomes error on Travis.
```
gavanderhoorn: Same comment here about it being an example.

130s:
Thanks again, done.

130s:
@gavanderhoorn Does this look ok now?

shaun-edwards:
@gavanderhoorn, ping?

shaun-edwards:
@gavanderhoorn, ping?

130s:
Rebased to resolve conflict. It's been a while but can we move(it!) on? @gavanderhoorn

130s:
Let me go ahead merge since [I already got "ok" in the beginning](https://github.com/ros-industrial/industrial_ci/pull/15#issuecomment-172903024).

gavanderhoorn:
Sorry about this. Just looked up my email history and I can't find any of the pings of @shaun-edwards, not sure what happened there.

A bit late, but a :+1: from me.

PR #16, opened 2016-01-09
Add option to not test 

130s:
There might a case where you don't want to run the tests, neither from build nor install spaces.

Example; device driver packages where emulation is not yet available, but you still want to check other things defined in the CI config (build etc.)

Since `NOT_TEST_INSTALL` is already implemented, in this PR I'm adding `NOT_TEST_BUILD`.


shaun-edwards:
+1

PR #19, opened 2016-02-07
Fix folding typo (traivs->travis).

gavanderhoorn:
As per title.

Looking at [travis/build/shell/dsl.rb](https://github.com/travis-ci/travis-build/blob/91bf066/lib/travis/build/shell/dsl.rb#L58-L63) it's `travis_*`, not `traivs_*`.


130s:
Wow, thanks for the fix and the reference.

PR #20, opened 2016-02-08
Move a patch that becomes available via DEB to older ROS distro only section

130s:
to keep the code cleaner. Looks like the patch in `rostest` in question is already included in [deb version of 1.11.16 that's already out there](http://repositories.ros.org/status_page/ros_indigo_default.html?q=rostest).

This PR will resolve a patch error like https://github.com/ros-industrial/staubli/pull/20#issuecomment-18076221.

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -203,8 +203,8 @@ if [ "$NOT_TEST_BUILD" != "true" ]; then
     
     # Patches for rostest that are only available in newer codes.
     # Some are already available via DEBs so that patches for them are not needed, but because EOLed distros (e.g. Hydro) where those patches are not released into may be still tested, all known patches are applied here.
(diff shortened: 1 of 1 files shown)


130s:
The same patch error was already happening in [a PR in Dec 2015](https://github.com/ros-industrial/industrial_core/pull/128) (in particular [here](https://travis-ci.org/ros-industrial/industrial_core/jobs/96340855#L3598)). That said we haven't utilized this particular patch at all from the beginning of this repo. So there shouldn't be an issue to remove it.

shaun-edwards:
+1

130s:
This is a functional change (should not effect almost all the cases) so I just created [a new tag](https://github.com/ros-industrial/industrial_ci/releases/tag/0.1.2).

PR #22, opened 2016-02-15
Fix undeclared args for the number parallel job

130s:
Fixing that some args for the number of parallel job are not documented nor set default values.


shaun-edwards:
+1...in need of this for `industrial_core`.  Running test jobs in parallel causes socket connections to fail.

130s:
@shaun-edwards if this does not solve your issue, please let me know.

PR #23, opened 2016-02-15
Add support for catkin_make_isolated. Enable source build using .rosinstall

130s:
DO NOT MERGE YET.

This addresses #21.

<s>This should be considered as a work-in-progress; I might need to ass more commits to let it pass the travis test of this repository.</s>

PR #27, opened 2016-04-12
Workaround for rospack 2.3.0 (fix #26).

### LINKED ISSUE #26: Nodelet dependency
The `ros_canopen` travis jobs currently work if tested against the indigo repo, but fail against the shadow-fixed repo (https://travis-ci.org/ros-industrial/ros_canopen/builds/122520707).

It's all good until the test install space step, which fails with
````
+++rospack plugins --attrib=plugin nodelet
[rospack] Error: no such package nodelet
```
The corresponding line is https://github.com/ros-industrial/industrial_ci/blob/master/travis.sh#L233
It introduces a dependency on the `nodelet` package.

For some reason it worked without having the `nodelet` package before (transitive dependency?). However, this seems to have changed with new changes in shadow-fixed.

Where does this line come from? What is its purpose?
I would suggest to remove it.

130s:
https://github.com/ros-industrial/industrial_ci/compare/master...130s:fix/rospackplugins#diff-f9a62833538ab14fff849c2be16ab986L233

```
rospack plugins --attrib=plugin nodelet
```

This line is nice to check if the nodelet plugin is properly configured or not when the software tested contains at least one plugin. But this particular line is so agressive that even repositories where no nodelet is used turns to failure, with `rospack` version 2.3.0.

So I'm kind of relaxing this line of test by only printing warning message regardless of the check result, without ending the entire build on Travis.

I'm now running test using a downstream repo, [my forked ros_canopen](https://github.com/130s/ros_canopen/pull/1).

(I've already implemented a feature to specify downstream repos utilizing `.rosinstall`, but it's in https://github.com/ros-industrial/industrial_ci/pull/23 and even the specific commit is mixed with irrelevant change (my bad) that's why I couldn't use .rosinstall in this PR.)

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -64,6 +64,9 @@ function error {
     exit 1
 }
 
+CHARCOLOR_YELLOW='\033[1;33m'
+CHARCOLOR_NOCOLOR='\033[0m'
+
 BUILDER=catkin
 ROSWS=wstool
 CI_PARENT_DIR=.ci_config  # This is the folder name that is used in downstream repositories in order to point to this repo.
@@ -230,7 +233,8 @@ if [ "$NOT_TEST_INSTALL" != "true" ]; then
         catkin build -i -v --summarize --no-status $BUILD_PKGS $CATKIN_PARALLEL_JOBS --make-args $ROS_PARALLEL_JOBS
         source install/setup.bash
         rospack profile
-        rospack plugins --attrib=plugin nodelet
+        # 
+        rospack plugins --attrib=plugin nodelet || echo -e "${CHARCOLOR_YELLOW}No nodelet plugin found.${CHARCOLOR_NOCOLOR} If any of the software in your repository does not utilize nodelet, ignore this. ${CHARCOLOR_YELLOW}But if your software depends on nodelet but you're seeing this warning, something is wrong.${CHARCOLOR_NOCOLOR}"
     fi
 
     travis_time_end  # catkin_install_build


130s:
[Test at my forked ros_canopen repo passed](https://github.com/130s/ros_canopen/issues/1). 
Please review @ipa-mdl

shaun-edwards:
Is this also the issue with ros-industrial-consortium/descartes#154 (among others).  @Jmeyer1292(ping)

ipa-mdl:
@shaun-edwards: yes, it is the same error.

Why do we need this line at all?
Which ROS-I packages really need this test?
(I wanted to discuss this in https://github.com/ros-industrial/industrial_ci/issues/26)

pjreed:
We just discovered that the change in rospack is causing https://github.com/swri-robotics/marti_messages to fail to build against the ros-shadow-fixed repository; there are no nodelets or plugins in there, so `rospack plugins --attrib=plugin nodelet` is now returning an error code and making the build fail.  It'd be nice to have this merged in soon.

shaun-edwards:
@pjreed, I believe this should have been fixed in #28.

@130s, thanks for your help with debugging this issue and the work-around.  I appreciate all of your effort.

130s:
Yes, #28 should solve the issue.

Yet I'll keep this PR open since there's a discussion going on in #26 and I like to add an enhancement in the end.

ipa-mdl:
`plugins --attrib=plugin nodelet` could be an `AFTER_SCRIPT`.

ipa-mdl:
`rospack` does not get installed anymore -> closing

PR #28, opened 2016-04-13
Temporarilly disable rospack plugin line (fixes #26).

### LINKED ISSUE #26: Nodelet dependency

130s:
With this #26 should disappear.

(This change isn't necessarily the permanent fix (discussion is ongoing in #26)).

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -230,7 +230,6 @@ if [ "$NOT_TEST_INSTALL" != "true" ]; then
         catkin build -i -v --summarize --no-status $BUILD_PKGS $CATKIN_PARALLEL_JOBS --make-args $ROS_PARALLEL_JOBS
         source install/setup.bash
         rospack profile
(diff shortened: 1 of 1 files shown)


ipa-mdl:
+1

Issue #30, opened 2016-04-19
catkin clean -a


Jmeyer1292:
After merging a new Descartes pull request, I received a notification that the build had failed. A quick search reveals:

```
+++catkin clean -a

usage: catkin [-h] [-a] [--test-colors] [--version]

              [--force-color | --no-color]

              [build | clean | config | create | env | init | list | locate | profile]

              ...

catkin: error: unrecognized arguments: -a
```
See the following for the full log:
https://travis-ci.org/ros-industrial-consortium/descartes/jobs/124093593

Is there something I'm not setting? I'll keep looking, but I wanted to throw this up here in case someone else has an idea.

130s:
Thanks for the report. I'll look into this.
Seems related https://github.com/catkin/catkin_tools/issues/320.

130s:
Should be resolved by #31. @Jmeyer1292 @VictorLamoine if it isn't please let me know on this ticket.

130s:
You can simply re-run the Travis jobs that failed.

PR #31, opened 2016-04-19
Adjust to catkin_tools 0.4.0

130s:
Initialized by #30, with the Travis build of this PR I'm trying to find if there's any difference introduced by [catkin_tools 0.4.0 that was released today](http://lists.ros.org/pipermail/ros-users/2016-April/069968.html).

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -10,6 +10,8 @@
 
   <build_depend>catkin</build_depend>
   <build_depend>python</build_depend>
+  <build_depend version_gte="0.4.0">python-catkin-tools</build_depend>
+  <run_depend version_gte="0.4.0">python-catkin-tools</run_depend>
   <run_depend>rospy_tutorials</run_depend>
   <run_depend>rostest</run_depend>
   <export/>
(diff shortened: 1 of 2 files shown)


130s:
Travis tests are all green.
I'm inclined to merge this and see how #30 goes, but I need +1 to do so.

VictorLamoine:
:+1: I came up with the exact same fix

130s:
@VictorLamoine thanks!

This should fix #30.

package.xml
```diff
@@ -10,6 +10,8 @@
 
   <build_depend>catkin</build_depend>
   <build_depend>python</build_depend>
+  <build_depend version_gte="0.4.0">python-catkin-tools</build_depend>
+  <run_depend version_gte="0.4.0">python-catkin-tools</run_depend>
```
ipa-mdl: If I remember it correctly, the version are ignored by rosdep and are just for the release packaging.
However, they are good indicators for people that are debugging problems.

130s: Yeah, I guess only systematic reaction is that catkin [prints warning in a certain condition](http://answers.ros.org/question/136526/how-versioned-build_depend-works/). 
I added it more for human purpose, not for machine.

PR #33, opened 2016-04-19
Fix apt-get quiet option

VictorLamoine:
The `-q` and `-qq` switches must be appended before the apt-get command to work.
Providing `-q` and `-qq` is useless as `-qq` overrides `-q`.

man:
```
-q, --quiet
           Quiet; produces output suitable for logging, omitting progress indicators. More q's will produce more quiet up to a maximum of 2. You can also use -q=# to set the quiet
           level, overriding the configuration file. Note that quiet level 2 implies -y; you should never use -qq without a no-action modifier such as -d, --print-uris or -s as
           APT may decide to do something you did not expect. Configuration Item: quiet.
```


130s:
Sounds about right. +1

PR #42, opened 2016-05-09
Update ros.key acquision.

130s:
[How to acquire apt key has changed since Indigo](http://wiki.ros.org/indigo/Installation/Ubuntu#indigo.2BAC8-Installation.2BAC8-Sources.Set_up_your_keys). 

Not sure how relevant this is since the older method has been working so far for Indigo, but Travis today fails (e.g. [this](https://travis-ci.org/130s/moveit_core/jobs/128791932)) at key acquisition. So it's better to update to the latest method.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,8 @@ env:
     - ROS_DISTRO=indigo CATKIN_PARALLEL_JOBS='-p1' ROS_PARALLEL_JOBS='-j1'  # Intend build on low-power platform
     # - env: ROS_DISTRO=indigo PRERELEASE=true  ## Comment out because this is meaningless for always failing without prerelease testable contents in industrial_ci.
     - ROS_DISTRO=indigo PRERELEASE=true PRERELEASE_REPONAME=industrial_core PRERELEASE_DOWNSTREAM_DEPTH=0
+    - ROS_DISTRO=indigo APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet  # Passing wrong SKS URL as a break test. Should still pass.
+    - ROS_DISTRO=jade   APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet  # Passing wrong SKS URL as a break test. Should still pass.
     - ROS_DISTRO=jade   PRERELEASE=true PRERELEASE_REPONAME=warehouse_ros PRERELEASE_DOWNSTREAM_DEPTH=1
 matrix:
   allow_failures:
--- a/travis.sh
+++ b/travis.sh
@@ -60,14 +60,16 @@ if [ ! "$ROS_PARALLEL_JOBS" ]; then export ROS_PARALLEL_JOBS="-j8"; fi
 if [ ! "$ROS_PARALLEL_TEST_JOBS" ]; then export ROS_PARALLEL_TEST_JOBS="$ROS_PARALLEL_JOBS"; fi
 # If not specified, use ROS Shadow repository http://wiki.ros.org/ShadowRepository
 if [ ! "$ROS_REPOSITORY_PATH" ]; then export ROS_REPOSITORY_PATH="http://packages.ros.org/ros-shadow-fixed/ubuntu"; fi
+# For apt key stores
+if [ ! "$APTKEY_STORE_HTTPS" ]; then export APTKEY_STORE_HTTPS="https://raw.githubusercontent.com/ros/rosdistro/master/ros.key"; fi
+if [ ! "$APTKEY_STORE_SKS" ]; then export APTKEY_STORE_SKS="hkp://ha.pool.sks-keyservers.net"; fi  # Export a variable for SKS URL for break-testing purpose.
+if [ ! "$HASHKEY_SKS" ]; then export HASHKEY_SKS="0xB01FA116"; fi
+
 echo "Testing branch $TRAVIS_BRANCH of $DOWNSTREAM_REPO_NAME"
 # Set apt repo
 sudo -E sh -c 'echo "deb $ROS_REPOSITORY_PATH `lsb_release -cs` main" > /etc/apt/sources.list.d/ros-latest.list'
 # Common ROS install preparation
 # apt key acquisition. Since keyserver may often become accessible, backup method is added.
-if [ ! "$APTKEY_STORE_SKS" ]; then export APTKEY_STORE_SKS="hkp://ha.pool.sks-keyservers.net"; fi  # Export a variable for SKS URL for break-testing purpose.
-if [ ! "$HASHKEY_SKS" ]; then export HASHKEY_SKS="0xB01FA116"; fi
-if [ ! "$APTKEY_STORE_HTTPS" ]; then export APTKEY_STORE_HTTPS="https://raw.githubusercontent.com/ros/rosdistro/master/ros.key"; fi
 sudo apt-key adv --keyserver $APTKEY_STORE_SKS --recv-key $HASHKEY_SKS || ((echo 'Fetching apt key from SKS keyserver somehow failed. Trying to get one from alternative.\n'; wget $APTKEY_STORE_HTTPS -O - | sudo apt-key add -) || (echo 'Fetching apt key by an alternative method failed too. Exiting since ROS cannot be installed.'; error))
 lsb_release -a
 sudo apt-get update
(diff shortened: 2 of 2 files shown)


130s:
This change might not be required but I guess recommended; there seems to be times where key server wasn't responsive with the old method. http://answers.ros.org/question/212254/ros-indigo-key-server-down/

shaun-edwards:
+1

gavanderhoorn:
@130s wrote:
> This change might not be required but I guess recommended; there seems to be times where key server wasn't responsive with the old method. http://answers.ros.org/question/212254/ros-indigo-key-server-down/

I'm not sure, but I think that issue actually complains about the fact that the *new* method doesn't work. The keyservers in the pool that was used were not always up/reachable, leading to import errors. I've answered many questions about that on RA.

> [..] but Travis today fails (e.g. this) at key acquisition.

I'm not sure why that build got a `403`, but today, `wget http://packages.ros.org/ros.key` actually works for me.

I may be a bit conservative, but if we want to move away from the 'old method', perhaps we should use `wget https://raw.githubusercontent.com/ros/rosdistro/master/ros.key`, as @tfoote wrote in the answer on RA. That is the same key as on the key server, but served over https instead of http, and should almost always work.

The keyservers have proven unstable in the recent past, but perhaps the HA (high availability) pool is stable enough.

---

Perhaps a bit of checking would be nice, something like (pseudo):
```bash
# try key servers
apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net [..]

# make sure it succeeded
if [ $? -ne 0 ]; then
  # try alternative
  wget https://raw.githubusercontent.com/ros/rosdistro/master/ros.key -O - | sudo apt-key add -

  # check again ..
  ..
fi
```

130s:
Great to have a ROS-A guru!

> I may be a bit conservative, but if we want to move away from the 'old method', perhaps we should use wget https://raw.githubusercontent.com/ros/rosdistro/master/ros.key, as @tfoote wrote in the answer on RA. That is the same key as on the key server, but served over https instead of http, and should almost always work.
> :
> Perhaps a bit of checking would be nice, something like (pseudo):

Sounds reasonable. Based on your comment, I made a change that does:
- Use `ha.pool.sks-keyservers.net` as a primary key source (as recommended in [ROS Indigo installation page](http://wiki.ros.org/indigo/Installation/Ubuntu#indigo.2BAC8-Installation.2BAC8-Sources.Set_up_your_keys))
- If fail, fall back on `https://raw.githubusercontent.com/ros/rosdistro/master/ros.key`
- If neither works then exit the entire script (since there's no way that ROS will be installed (other than building the entire ROS from source, which is out of scope here)).

Could I ask you to review @gavanderhoorn ?

.travis.yml
```diff
@@ -17,6 +17,8 @@ env:
     - ROS_DISTRO=indigo CATKIN_PARALLEL_JOBS='-p1' ROS_PARALLEL_JOBS='-j1'  # Intend build on low-power platform
     # - env: ROS_DISTRO=indigo PRERELEASE=true  ## Comment out because this is meaningless for always failing without prerelease testable contents in industrial_ci.
     - ROS_DISTRO=indigo PRERELEASE=true PRERELEASE_REPONAME=industrial_core PRERELEASE_DOWNSTREAM_DEPTH=0
+    - ROS_DISTRO=indigo APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet  # Passing wrong SKS URL as a break test. Should still pass.
+    - ROS_DISTRO=jade   APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet
```
gavanderhoorn: You probably already intended to, but don't forget to change this back to `.net` (from `.vet`).

travis.sh
```diff
@@ -64,7 +64,9 @@ echo "Testing branch $TRAVIS_BRANCH of $DOWNSTREAM_REPO_NAME"
 # Set apt repo
 sudo -E sh -c 'echo "deb $ROS_REPOSITORY_PATH `lsb_release -cs` main" > /etc/apt/sources.list.d/ros-latest.list'
 # Common ROS install preparation
-wget http://packages.ros.org/ros.key -O - | sudo apt-key add -
+# apt key acquisition. Since keyserver may often become accessible, backup method is added.
+if [ ! "$APTKEY_STORE_SKS" ]; then export APTKEY_STORE_SKS="hkp://ha.pool.sks-keyservers.net"; fi  # Export a variable for SKS URL for break-testing purpose.
+sudo apt-key adv --keyserver $APTKEY_STORE_SKS --recv-key 0xB01FA116 || ((echo 'Fetching apt key from SKS keyserver somehow failed. Trying to get one from alternative.\n'; wget https://raw.githubusercontent.com/ros/rosdistro/master/ros.key -O - | sudo apt-key add -) || (echo 'Fetching apt key by an alternative method failed too. Exitting since ROS cannot be installed.'; error))
```
gavanderhoorn: Not sure what style you follow with other 'constants', but it might be good to store the `githubusercontent.com` url in a variable as well.

And the key hash maybe?

gavanderhoorn: Typo: `exitting` -> `exiting`.

130s: The only reason why I exported SKS server URL as a variable is that I wanted to add a job to test the situation with network being unavailable, and I implemented that by passing wrong SKS server URL (that's what I'm doing in [this](https://github.com/ros-industrial/industrial_ci/pull/42/files/dbc496b1f2c3d1ee57d4a7bcc0ff059e09f35f1f#diff-354f30a63fb0907d4ad57269548329e3R20) and the following lines). That being said I have few reason to export `githubusercontent.com`. Does this make sense?

gavanderhoorn: Yeah, as I wrote, it depends on your style in the rest of the script. Personally I always create constants/variables, but you don't have to. It's up to you.

.travis.yml
130s: Yeah this `.vet` is intended; Wanted to check if the script falls back on to the alternative key location when SKS server isn't accessible (same as a line above). I've added an in-line comment.

travis.sh
130s: Oh like, export all constants as variables? If so, I support that idea. That allows customization (no matter how niche that is) in case there's a need.

Done.

travis.sh
```diff
@@ -64,7 +64,11 @@ echo "Testing branch $TRAVIS_BRANCH of $DOWNSTREAM_REPO_NAME"
 # Set apt repo
 sudo -E sh -c 'echo "deb $ROS_REPOSITORY_PATH `lsb_release -cs` main" > /etc/apt/sources.list.d/ros-latest.list'
 # Common ROS install preparation
-wget http://packages.ros.org/ros.key -O - | sudo apt-key add -
+# apt key acquisition. Since keyserver may often become accessible, backup method is added.
+if [ ! "$APTKEY_STORE_SKS" ]; then export APTKEY_STORE_SKS="hkp://ha.pool.sks-keyservers.net"; fi  # Export a variable for SKS URL for break-testing purpose.
+if [ ! "$HASHKEY_SKS" ]; then export HASHKEY_SKS="0xB01FA116"; fi
+if [ ! "$APTKEY_STORE_HTTPS" ]; then export APTKEY_STORE_HTTPS="https://raw.githubusercontent.com/ros/rosdistro/master/ros.key"; fi
```
gavanderhoorn: More of a general comment (and again a style thing I guess), but any reason not to check and export all variables at the top of the script?

Related: do you need to `export` those variables? If not, `LOCAL_VAR=${VAR:-SOME_DEFAULT}` would be much more concise.

130s: > variables at the top of the script?

Again I agree!

> do you need to export those variables?

Ha, yes, I don't know why I'm exporting those. Need to go to bed now I think.

gavanderhoorn: > Ha, yes, I don't know why I'm exporting those. Need to go to bed now I think.

:). Maybe you should :).

130s: Actually I changed my mind and leave variables exported;it isn't harmful, and extends customizability (e.g. One can inherit env vars if s/he wants to run a custom script after `travis.sh`. Need to run like `. travis.sh` though to preserve env vars).

130s:
@gavanderhoorn Does this look good for you?

gavanderhoorn:
Yes, looks good.

Don't forget to change `.vet` to `.net` :).

130s:
Thanks!

> Don't forget to change .vet to .net :).

Please see the comment https://github.com/ros-industrial/industrial_ci/pull/42#discussion_r62631156. 
Merging but if that's still an issue feel free to add a comment.

gavanderhoorn:
>> Don't forget to change .vet to .net :).

> Please see the comment https://github.com/ros-industrial/industrial_ci/pull/42#discussion_r62631156. 
Merging but if that's still an issue feel free to add a comment. 

o wait. I think I understand now: that `.vet` is part of the tests for this repository.

PR #46, opened 2016-05-14
enable USE_DEB

rhaschke:
Was there a particular reason that USE_DEB functionality was disabled?
For MoveIt, it would be very helpful to build and test a whole workspace because there are strong interdependencies between individual repositories.

The only issue I could see, was that the actual downstream should be taken from travis but not from `.rosinstall`. I managed to ensure that by removing the downstream from .rosinstall and then using the existing link-in of the actual source folder.

I also added support for an url directly specified in USE_DEB.


rhaschke:
Replaced by individual PR #47.

PR #47, opened 2016-05-14
enable USE_DEB

rhaschke:
Was there a particular reason that USE_DEB functionality was disabled?
For MoveIt, it would be very helpful to build and test a whole workspace because there are strong interdependencies between individual repositories.

The only issue I could see, was that the actual downstream should be taken from travis but not from `.rosinstall`. I managed to ensure that by removing the downstream from .rosinstall and then using the existing link-in of the actual source folder.

I also added support for an url directly specified in USE_DEB.

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -311,6 +311,8 @@ fi
 travis_time_start after_script
 
 ## BEGIN: travis' after_script
+
+set +x
 PATH=/usr/local/bin:$PATH  # for installed catkin_test_results
 PYTHONPATH=/usr/local/lib/python2.7/dist-packages:$PYTHONPATH
 if [ "${ROS_LOG_DIR// }" == "" ]; then export ROS_LOG_DIR=~/.ros/test_results; fi # http://wiki.ros.org/ROS/EnvironmentVariables#ROS_LOG_DIR
@@ -320,6 +322,6 @@ if [ "$BUILDER" == catkin -a -e ~/.ros/test_results/ ]; then catkin_test_results
 
 travis_time_end  # after_script
 
+set +x
 cd $TRAVIS_BUILD_DIR  # cd back to the repository's home directory with travis
 pwd
-
--- a/travis.sh
+++ b/travis.sh
@@ -29,7 +29,7 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 #
-## Greatly inspired by JSK travis https://github.com/jsk-ros-pkg/jsk_travis 
+## Greatly inspired by JSK travis https://github.com/jsk-ros-pkg/jsk_travis
 ## Author: Isaac I. Y. Saito
 
 ## This is a "common" script that can be run on travis CI at a downstream github repository.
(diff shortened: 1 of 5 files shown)


130s:
> Was there a particular reason that USE_DEB functionality was disabled?

Honestly, not really. It's just when I was initiating the package there wasn't a need for this feature. You enabling it is truly appreciated! Let me review in depth tomorrow.

travis.sh
```diff
@@ -121,24 +121,39 @@ travis_time_start setup_rosws
...
-###    fi
-###    if [ -e $CI_SOURCE_PATH/.travis.rosinstall.$ROS_DISTRO ]; then
-###        # install (maybe unreleased version) dependencies from source for specific ros version
-###        $ROSWS merge file://$CI_SOURCE_PATH/.travis.rosinstall.$ROS_DISTRO
-###    fi
-###    $ROSWS update
-###    $ROSWS set $DOWNSTREAM_REPO_NAME http://github.com/$TRAVIS_REPO_SLUG --git -y
-###fi
+case "$USE_DEB" in
+true) # When USE_DEB is true, the dependended packages that need to be built from source are downloaded based on .travis.rosinstall file.
```
130s: Shouldn't this be `false`?

130s: Shouldn't this be false?
You may have referred to the comment in my old commit that said `true`, but I think my comment was wrong.

travis.sh
```diff
@@ -121,24 +121,39 @@ travis_time_start setup_rosws
...
+   fi
+   if [ -e $CI_SOURCE_PATH/.travis.rosinstall.$ROS_DISTRO ]; then
+       # install (maybe unreleased version) dependencies from source for specific ros version
+       $ROSWS merge file://$CI_SOURCE_PATH/.travis.rosinstall.$ROS_DISTRO
+   fi
+   ;;
+
+source)
+   $ROSWS init .
+   $DOWNSTREAM_REPO_NAME/setup_upstream.sh -w ~/ros/ws_$DOWNSTREAM_REPO_NAME
```
130s: Could you provide an example of `setup_upstream.sh`? I'm a bit lost for the entire `source` case.

130s:
Looks good with some comments already added.
I've addressed some and also added a further change, targeting your branch at https://github.com/ubi-agni/industrial_ci/pull/1 (which seems to have already merged another PR #48, so you may not want to merge https://github.com/ubi-agni/industrial_ci/pull/1 but instead cherry-pick my commits?).

travis.sh
rhaschke: Indeed, I simply followed the original code and comments. Changing true to false, is not clear to me. What is the intended meaning of USE_DEB? Is it "Use Debian packages"? In that case, the values "source" and <url> are not plausible anymore.
I was thinking of that variable indicating the source of the upstream workspace. What about changing USE_DEB to UPSTREAM_WORKSPACE with values "file", "source", and <url>?

travis.sh
rhaschke: No I can't. This is code was taken over from the disabled code. My thinking of this was, that setup_upstream.sh sets up the .rosinstall file.

travis.sh
```diff
@@ -121,24 +121,39 @@ travis_time_start setup_rosws
...
+
+# download upstream packages into workspace
+if [ -e .rosinstall ]; then
+   # ensure that the downstream is not in .rosinstall
+   $ROSWS rm $DOWNSTREAM_REPO_NAME || true
+   $ROSWS update
+fi
 # CI_SOURCE_PATH is the path of the downstream repository that we are testing. Link it to the catkin workspace
 ln -s $CI_SOURCE_PATH .
-####if [ "$USE_DEB" == source -a -e $DOWNSTREAM_REPO_NAME/setup_upstream.sh ]; then $ROSWS init .; $DOWNSTREAM_REPO_NAME/setup_upstream.sh -w ~/ros/ws_$DOWNSTREAM_REPO_NAME ; $ROSWS update; fi
```
rhaschke: This was the original location of the `setup_upstream.sh` code.

travis.sh
130s: I'll have to come back later for the review, but a short comment:

> What is the intended meaning of USE_DEB? Is it "Use Debian packages"?

Yes, with a little more elaboration, "Use debian binary packages for the upstream (ie. packages that the package being tested depends on)". 

> What about changing USE_DEB to 

This is my very bad, but when I first created this package I already included this variable in an example `.travis.yml` so that there are a number of packages out there that already have this defined. So as a "backward" compatibility, I hope we can keep using the same variable name...

rhaschke: As the code using this variable wasn't enabled yet and you changed the semantics of this variable (from true to false enabling the .rosinstall), it's a good point in time to change the name and clarify the semantics ;-)

130s: > As the code using this variable wasn't enabled yet and you changed the semantics of this variable (from true to false enabling the .rosinstall)

No, I haven't changed the semantics. I think that `USE_DEB=false` makes sense as an option that utilizes .rosinstall file, like so:

- `USE_DEB=false`: use source from .rosinstall.
- `USE_DEB=true`: use debian binary.

travis.sh
130s: Sorry, I should've known myself what this option was.
For now this feature with `source`, I can't think of a good usecase (in the repo I originally took the entire script from, [there's an example](https://github.com/jsk-ros-pkg/jsk_travis/commit/1e62bd53db6b1be819b59196055370b1462ee621#diff-04c6e90faac2675aa89e2176d2eec7d8R67) though). I'm inclined to omit this option.

travis.sh
rhaschke: You are right. The code was implementing what you say. However, the comment stated the opposite. In any case, I suggest to replace USE_DEB by some more verbose variable as suggested above:
UPSTREAM_WORKSPACE with values
- "debian": use debian binaries (default)
- "file": consider .travis.rosinstall
- "source": setup using setup_upstream.sh script

130s:
@rhaschke Could you rebase?

rhaschke:
@130s I currently don't have the time to look after this PR as well. Please feel free to take over. Most of the changes are from your side anyway.

130s:
Got it, I'm closing in favor of #59.

Issue #49, opened 2016-05-16
[Prerelease Test] Test often disregards the actual result from Catkin


130s:
For example https://travis-ci.org/anqixu/ueye_cam/jobs/130600243#L9129 in https://github.com/anqixu/ueye_cam/pull/44
```
+catkin_test_results --verbose
Summary: 0 tests, 0 errors, 0 failures
+echo 'ROS Prerelease Test went successful.'
ROS Prerelease Test went successful.
+exit 0
+travis_time_end
+set +x
<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
Function show_testresult took 0 min 0 sec
+cd /home/travis/build/anqixu/ueye_cam
+pwd
/home/travis/build/anqixu/ueye_cam
+++catkin_test_results build
Test results directory "/home/travis/ros/ws_ueye_cam/src/build" does not exist
+++error
+++travis_time_end 31
+++set +x
<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
Function prerelease_from_travis_sh took 18 min 25 sec
```

Looks similar to https://github.com/ros-industrial/industrial_ci/pull/35#issuecomment-216399634 but I'm not sure.

I'm afraid that this issue makes Prerelease Test feature of `industrial_ci` hard to use, since it makes its result unstable regardless the actual test result from Catkin.

ipa-mdl:
Why is the output checked in  https://github.com/ros-industrial/industrial_ci/blob/master/ros_pre-release.sh#L74 AND https://github.com/ros-industrial/industrial_ci/blob/master/travis.sh#L175?
The latter incovation is wrong because it includes the non-existing build directory.

130s:
Ok, for some reason the Travis job for [the PR](https://github.com/ros-industrial/industrial_ci/pull/35) where Prerelease feature was initially added [didn't run the right code and the job passed without error](https://s3.amazonaws.com/archive.travis-ci.org/jobs/128151566/log.txt) -- it didn't run "[catkin_test_results build](https://github.com/ros-industrial/industrial_ci/pull/35/files#diff-f9a62833538ab14fff849c2be16ab986R177)" that WAS included in the PR.

Anyways, I'm working of a fix.

PR #50, opened 2016-05-19
Fix for #49 (prerelease test result failure)

130s:
<s>**DO NOT Review nor Merge yet.**</s>
Ready to be reviewed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,11 @@ dist: trusty
 language: generic
 compiler:
   - gcc
-#notifications:
-#  Whoever receives success and/or failure via email
+notifications:
+  #  Whoever receives success and/or failure via email
+  email:
+    recipients:
+      - gm130s@gmail.com  # Change this to yours should you copy this file.
 env:
   matrix:
     - ROS_DISTRO=indigo NOT_TEST_BUILD='true'
@@ -17,6 +20,7 @@ env:
     - ROS_DISTRO=indigo CATKIN_PARALLEL_JOBS='-p1' ROS_PARALLEL_JOBS='-j1'  # Intend build on low-power platform
     # - env: ROS_DISTRO=indigo PRERELEASE=true  ## Comment out because this is meaningless for always failing without prerelease testable contents in industrial_ci.
     - ROS_DISTRO=indigo PRERELEASE=true PRERELEASE_REPONAME=industrial_core PRERELEASE_DOWNSTREAM_DEPTH=0
+    - ROS_DISTRO=indigo PRERELEASE=true PRERELEASE_REPONAME=ros_canopen
     - ROS_DISTRO=indigo APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet  # Passing wrong SKS URL as a break test. Should still pass.
     - ROS_DISTRO=jade   APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet  # Passing wrong SKS URL as a break test. Should still pass.
     - ROS_DISTRO=jade   PRERELEASE=true PRERELEASE_REPONAME=warehouse_ros PRERELEASE_DOWNSTREAM_DEPTH=1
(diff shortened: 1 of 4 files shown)


130s:
Ready to be reviewed. @ipa-mdl Could you take a look? This solves #49.

travis.sh
```diff
@@ -170,11 +170,12 @@ travis_time_end  # rosdep_install
...
-if [ "$PRERELEASE" == true ] && [ -e ${CI_SOURCE_PATH}/$CI_PARENT_DIR/ros_pre-release.sh ]; then 
-  ${CI_SOURCE_PATH}/$CI_PARENT_DIR/ros_pre-release.sh
-  catkin_test_results build && (echo 'ROS Prerelease Test went successful.'; exit 0) || error
+if [ "$PRERELEASE" == true ] && [ -e ${CI_SOURCE_PATH}/$CI_PARENT_DIR/ros_pre-release.sh ]; then
+  ${CI_SOURCE_PATH}/${CI_PARENT_DIR}/ros_pre-release.sh
+  retval_prerelease=$?
+  if [ $retval_prerelease -eq 0 ]; then success; else error; fi  # Internally called travis_time_end for prerelease_from_travis_sh
+  # With Prerelease option, we want to stop here without running the rest of the code.
 fi
-travis_time_end  # prerelease_from_travis_sh
```
ipa-mdl: did you intend to delete this line? Should the start time (lin2 172) be deleted as well?

.travis.yml
```diff
@@ -8,8 +8,10 @@ dist: trusty
 language: generic
 compiler:
   - gcc
-#notifications:
-#  Whoever receives success and/or failure via email
+notifications:
+  email:
+    recipients:
+      - gm130s@gmail.com
```
ipa-mdl: I would add a comment to change this line if copied.

travis.sh
130s: Removing `travis_time_end` in this line is intended. I now let [`success` and `error` functions do the work internally](https://github.com/ros-industrial/industrial_ci/pull/50/files/bb704c7af752fa389e1e9e4a5ad990eab3b6812b#diff-a188fc26ef16ad3194c8d957e9e0cd3dR78).

Prior to this PR, `error` function already ran `travis_time_end`, but only when non-zero result occurred. `error` also run `exit 1`, which was called only with non-zero. So it was kind of redundant and incomplete. I hope this PR is cleaner and complete, in terms of fold and exit.

.travis.yml
130s: Sorry, I don't understand your comment. I reverted back the comment line in my updated commit. Is that what you meant?

ros_pre-release.sh
```diff
@@ -71,9 +72,10 @@ travis_time_start run_prerelease
 travis_time_end  # run_prerelease
 
 travis_time_start show_testresult
-catkin_test_results --verbose && (echo 'ROS Prerelease Test went successful.'; exit 0) || error
+catkin_test_results --verbose && { echo 'ROS Prerelease Test went successful.'; RETVAL=0; success; } || { RETVAL=1; error; }
 travis_time_end  # show_testresult
```
ipa-mdl: So `travis_time_end` should not be used if block is terminated by `success` and `error`?

130s: Good point, and my answer is no. `return` [in this line](https://github.com/ros-industrial/industrial_ci/pull/50/files#diff-69501095267bf26a6434210905896aa4R81) is important for the parent script to be able to capture the result. So we don't want this script to exit at this line yet and thus don't want to call `success`.

Hope this variation is not unnecessarily complicated, but maybe it is. I appreciate your opinion.

.travis.yml
ipa-mdl: You have provided your mail address here, so adding a hint to change it for users that copied this config might prevent you from spam ;)

130s: Very thoughtful of you :) Addressed.

130s:
@ipa-mdl thank you for taking a look many times. If you don't have more to say I'll go ahead merge.

130s:
Wait Prerelease Tests aren't passing. I'll look.

130s:
Ready for the review.

.travis.yml
```diff
@@ -8,22 +8,27 @@ dist: trusty
 language: generic
 compiler:
   - gcc
-#notifications:
-#  Whoever receives success and/or failure via email
+notifications:
+  #  Whoever receives success and/or failure via email
+  email:
+    recipients:
+      - gm130s@gmail.com  # Change this to yours you copy this file.
```
ipa-mdl: +if ? ;)

ipa-mdl:
I think this PR is a litte bit mixed up, since it includes fixes and unrelated(?) enhancements, but the code it is much cleaner now, so +1 for merge :)

130s:
Thanks. Indeed the most of commits may seem unrelated but all of them contribute to the fix.
The last comment is addressed now.

Issue #53, opened 2016-05-27
Hash sum mismatch


VictorLamoine:
This is external to `industrial_ci` but all current builds fails because of a server problem:

```
W: Failed to fetch http://dl.hhvm.com/ubuntu/dists/trusty/main/binary-amd64/Packages  Hash Sum mismatch
E: Some index files failed to download. They have been ignored, or old ones used instead.
++++error
++++travis_time_end 31
```

130s:
Could you give us a link to some example results? I can't seem to find one.

VictorLamoine:
https://travis-ci.org/InstitutMaupertuis/universal_robot_motion/jobs/133313579#L242

VictorLamoine:
Closing as the builds are clear from this temporary bug now.
This issue was just to warn other users.

gavanderhoorn:
Docker is having some issues (see docker/docker#23203), so builds failing due to hash mismatches will happen again I reckon.

PR #56, opened 2016-05-27
[Workaround] "apt-get update" stops when there's an unaccessible repo

130s:
Work around #53 

I can't find an official way for `apt-get update` command to continue when it encounter an unaccessible repository. This change may not seem intelligent, but looking at the tests it's been working fine.

@VictorLamoine would you want to take a look?

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -108,7 +108,7 @@ sudo -E sh -c 'echo "deb $ROS_REPOSITORY_PATH `lsb_release -cs` main" > /etc/apt
 # apt key acquisition. Since keyserver may often become accessible, backup method is added.
 sudo apt-key adv --keyserver $APTKEY_STORE_SKS --recv-key $HASHKEY_SKS || ((echo 'Fetching apt key from SKS keyserver somehow failed. Trying to get one from alternative.\n'; wget $APTKEY_STORE_HTTPS -O - | sudo apt-key add -) || (echo 'Fetching apt key by an alternative method failed too. Exiting since ROS cannot be installed.'; error))
 lsb_release -a
-sudo apt-get update
+sudo apt-get update || (echo "ERROR: apt server not responding. This is a rare situation, and usually just waiting for a while clears this. See https://github.com/ros-industrial/industrial_ci/pull/56 for more of the discussion"; error)
 sudo apt-get -qq install -y python-catkin-tools python-rosdep python-wstool ros-$ROS_DISTRO-rosbash ros-$ROS_DISTRO-rospack
 # If more DEBs needed during preparation, define ADDITIONAL_DEBS variable where you list the name of DEB(S, delimitted by whitespace)
 if [ "$ADDITIONAL_DEBS" ]; then sudo apt-get install -q -qq -y $ADDITIONAL_DEBS;  fi


130s:
Any admins, if there'll be +1s, please go ahead merge -- I'll be down for hours but #53 seems to be making this package unusable.

gavanderhoorn:
If the `apt-get update` fails, the next lines trying to install the ROS packages will also presumably fail, right? Not sure this would be an improvement then, it would only make the build fail later.

VictorLamoine:
Test launched here:
https://travis-ci.org/InstitutMaupertuis/universal_robot_motion/builds/133345071

VictorLamoine:
:+1:

130s:
@gavanderhoorn 

> If the apt-get update fails, the next lines trying to install the ROS packages will also presumably fail, right? 

Yes, that's a fair question.
Thing is, when the fallback added in this PR is reached, that means `apt-get update` failed, which means we're already not in normal situation. So my idea was to at least move forward ignoring the rest of apt repositories (after a small research, looks like it's not just me who's surprised by `apt-get` missing an option to continue to run ignoring unaccessible repos) crossing fingers.

Apparently in the previous test of this PR where the issue #53 was still happening (looks like it's backed up now at the remote server. So "waiting" can be kind of a permanent option as [suggested ttoo](http://answers.ros.org/question/228493/hash-sum-mismatch-while-update-package-indexindigoubuntu-140404/) on answers), [this worked](https://travis-ci.org/ros-industrial/industrial_ci/jobs/133343463#L247), meaning, we were just lucky (this doesn't necessarily mean that the workaround in this PR is always stable).

gavanderhoorn:
> Thing is, when the fallback added in this PR is reached, that means `apt-get update` failed, which means we're already not in normal situation. So my idea was to at least move forward ignoring the rest of apt repositories

Exactly. So if building then fails later, does that then mean the changes introduced to the repository / by the PR are bad, or was it caused by the build script 'trying to move forward'? Sometimes building succeeds, but run dependencies are not present, leading to failures in tests. I don't think this is something that is uniformly desirable.

ipa-mdl:
I think it is okay for the CI tests to fail in these (rare?) situations.

Since we don't want to update all, but add the ROS repo it might be an option to just update this one:
```
apt-get update -o Dir::Etc::sourcelist=sources.list.d/ros-latest.list -o Dir::Etc::sourceparts=- -o APT::Get::List-Cleanup=0
```
(Taken from http://askubuntu.com/questions/65245/apt-get-update-only-for-a-specific-repository)

130s:
Trying to summarize up:
- No action is ok; just wait for a few hours and apt server usually backs up.
- But want to be friendly to users. Let's do something even the smallest for them.

I guess I agree with Gijs; "The smallest" should not confuse the cause of any future test results. So I updated the commit to just print a message and exit the entire job.

travis.sh
```diff
@@ -108,7 +108,7 @@ sudo -E sh -c 'echo "deb $ROS_REPOSITORY_PATH `lsb_release -cs` main" > /etc/apt
 # apt key acquisition. Since keyserver may often become accessible, backup method is added.
 sudo apt-key adv --keyserver $APTKEY_STORE_SKS --recv-key $HASHKEY_SKS || ((echo 'Fetching apt key from SKS keyserver somehow failed. Trying to get one from alternative.\n'; wget $APTKEY_STORE_HTTPS -O - | sudo apt-key add -) || (echo 'Fetching apt key by an alternative method failed too. Exiting since ROS cannot be installed.'; error))
 lsb_release -a
-sudo apt-get update
+sudo apt-get update || (echo "ERROR: apt server not resonding. This is a rare situation but usually you have to wait for awhile (typically a few hours) then try again. Sorry for the inconvenience. See https://github.com/ros-industrial/industrial_ci/pull/56 for more of the discussion"; error)
```
gavanderhoorn: Suggestion: "a rare situation but usually" -> "a rare situation, and usually just waiting for a while clears this."

gavanderhoorn: Typo: resonding -> responding

130s: Thx, updated.

VictorLamoine:
:+1: with this, it's simple and human readable.

gavanderhoorn:
The latest changes make sense, if you have nothing else to add I'd be in favour of merging this.

130s:
All right, thanks all!

PR #59, opened 2016-06-04
Enable UPSTREAM_WORKSPACE option

130s:
Just a rebased and squashed version of #47 that @rhaschke initiated.

PR #61, opened 2016-06-04
[Urgent] Fix wrong github raw content URL.

130s:
tl;dr URL for a .rosintall file is wrong so that the Travis always fail for now. This PR should fix it.

@gavanderhoorn could you review and merge if this LGT you? I have to go to bed :/

Detailed storyline (*warning*: boring):
- In https://github.com/ros-industrial/industrial_ci/pull/59 I added .rosintall files along with the logic that utilize them, and also wanted to check if the logic worked in the same PR.
- Initially I used the URL of .rosintall files on my forked repo so that Travis could find them.
- Once Travis passed, I swapped URL with [the one in `ros-industrial` organization](https://github.com/ros-industrial/industrial_ci/pull/59/commits/a68ddaba00a57eeaa020a68defc1c3373ef10faf#diff-354f30a63fb0907d4ad57269548329e3R27 ). And merge the PR without making sure the Travis passes (it shouldn't pass because the files never existed unless PR was merged). Turned out I used a wrong URL.


gavanderhoorn:
Not sure I completely understand what is going on, but the new URL does resolve and the old one doesn't, so I'll merge this for you.

130s:
Thank you for the quick review-merge!

PR #74, opened 2016-07-11
Better handling a cornercase where USE_DEB=true and UPSTREAM_WORKSPACE=file

130s:
Found that the current document might be confusing for users (at https://github.com/ros-industrial/staubli_experimental/pull/12).

`USE_DEB` variable was already there since when this package was initially introduced but recently it's gone deprecated https://github.com/ros-industrial/industrial_ci/pull/59, and having it unchanged while adding .rosinstall would collide.

@gavanderhoorn Would you find this change clear enough?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ env:
     - ROS_DISTRO=indigo APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet  # Passing wrong SKS URL as a break test. Should still pass.
     - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=debian
     - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=file  # Using default file name for ROSINSTALL_FILENAME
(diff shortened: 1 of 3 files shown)


gavanderhoorn:
Well, it gets the point across to not set `USE_DEB`, but maybe a sentence about whether that will mean that *all* dependencies will have to be supplied by the `.rosinstall` file (or that it will still use `rosdep`) would be good.

130s:
@gavanderhoorn Thank you. I ended up updating broader portion in the same section but hope it answers to your question.

gavanderhoorn:
Thanks. This clarifies it for me.

I'd already gone ahead and assumed that it worked as you describe, but this is good information to have in the readme.

+1 from me.

130s:
In fact I thought only document might not be helpful enough, so I added a code.
Now, both `USE_DEB=true` and `UPSTREAM_WORKSPACE=file` are defined together, the entire script fails with a warning message. Let's see if [this Travis job](https://travis-ci.org/ros-industrial/industrial_ci/jobs/143973401) just added fails as intended.

130s:
Worked as intended. Let me move on merging.

PR #77, opened 2016-09-03
Only use catkin_test_results --verbose if it exists

evenator:
Although industrial_ci isn't targeted at ROS hydro, I still maintain a hydro branch of some repositories to support legacy projects. The hydro version of catkin_test_results has no --verbose option, which causes the test to fail when used on hydro.

This PR checks to see if the --verbose flag exists before using it.

This was tested on a hydro branch of swri_robotics/mapviz: [before (failing)](https://travis-ci.org/swri-robotics/mapviz/builds/156967693) and [after (passing)](https://travis-ci.org/swri-robotics/mapviz/builds/157317100).

### DIFF
--- a/industrial_ci/ci_main.sh
+++ b/industrial_ci/ci_main.sh
@@ -87,6 +87,10 @@ travis_time_start init_travis_environment
 # Define more env vars
 BUILDER=catkin
 ROSWS=wstool
+# For compatibilility with hydro catkin, which has no --verbose flag
+CATKIN_TEST_RESULTS_CMD="catkin_test_results"
+if [ catkin_test_results --verbose 1>/dev/null 2>/dev/null; then CATKIN_TEST_RESULTS_CMD="catkin_test_results --verbose"; fi
+
 if [ ! "$CATKIN_PARALLEL_JOBS" ]; then export CATKIN_PARALLEL_JOBS="-p4"; fi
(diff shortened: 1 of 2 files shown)


industrial_ci/ci_main.sh
```diff
@@ -321,10 +321,14 @@ travis_time_start after_script
 ## BEGIN: travis' after_script
 PATH=/usr/local/bin:$PATH  # for installed catkin_test_results
 PYTHONPATH=/usr/local/lib/python2.7/dist-packages:$PYTHONPATH
+# For compatibilility with hydro catkin, which has no --verbose flag
```
130s: A matter of taste, but we're trying to define (environment) variables earlier as possible and collect at fewer places (around [here](https://github.com/ros-industrial/industrial_ci/blob/6f2e70627bfa86bd7fab8c6821d6ddfe35f3314c/industrial_ci/ci_main.sh#L87)). Would you want to move this there too?

130s:
Thanks for the addition. Please consider the comment I added. Also I'll open a PR against your branch to add a document for usage with older distros.

Btw, your initial Hydro build passed without error https://github.com/swri-robotics/mapviz/pull/415? Interesting.

evenator:
I'll move the environment variable to where you suggested, and I'll take a documentation PR whenever you have that ready.

The pass on #415 is actually an erroneous pass--it failed on the install step looking for a non-existent opencv3 package. Come to think of it, that might be a bug in industrial-ci...

evenator:
I merged the doc PR

Issue #78, opened 2016-09-04
Error at additional deb installation not captured


130s:
Pointed out at https://github.com/ros-industrial/industrial_ci/pull/77#issuecomment-244568034

Installation of `ros-indigo-opencv3` at https://travis-ci.org/swri-robotics/mapviz/builds/150746675#L679-L682 failed, which is defined [here](https://github.com/swri-robotics/mapviz/pull/415/files#diff-354f30a63fb0907d4ad57269548329e3R18) falsely for ROS Hydro build. However Travis CI succeeded. Obvious false-positive.

PR #79, opened 2016-09-04
Catch apt error for ADDITIONAL_DEB

130s:
Fix for #78

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,8 @@ env:
     - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=https://raw.githubusercontent.com/ros-industrial/industrial_ci/master/.travis.rosinstall
     - ROS_DISTRO=indigo USE_DEB=true  # Checking backup compatibility with UPSTREAM_WORKSPACE
     - ROS_DISTRO=indigo USE_DEB=false # Checking backup compatibility with UPSTREAM_WORKSPACE
+    - ROS_DISTRO=indigo ADDITIONAL_DEBS="ros-indigo-opencv3"
+    - ROS_DISTRO=indigo ADDITIONAL_DEBS="ros-hydro-opencv3"  # This should fail (trying from a wrong distro).
     - ROS_DISTRO=jade   APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet  # Passing wrong SKS URL as a break test. Should still pass.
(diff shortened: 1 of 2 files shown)


130s:
Could I ask @evenator for the review since you originally found the issue #78?

A caveat with this PR; I added a job to check true-negative (i.e. I want the job to fail upon error from the run), which I don't know yet how to capture on Travis CI. Right now the added jos is labeled `allow_failure` so that the entire jobs can pass but we just don't get notified for T-N. I [asked](http://stackoverflow.com/questions/39314067/to-capture-true-negative-on-travis-ci) if there's a way.

That said this PR is ready for review and merge if it's good.

evenator:
LGTM

Regarding intended failures, is there a way you could invert the output of the script for that particular job so that a fail is a pass and vice versa?

130s:
> Regarding intended failures, is there a way you could invert the output of the script for that particular job so that a fail is a pass and vice versa?

Added a commit https://github.com/ros-industrial/industrial_ci/pull/79/commits/f7d09fa441237c7a3b648fb4907bd4509f4dfaea for that. Not elegant at all though.

130s:
> Regarding intended failures, is there a way you could invert the output of the script for that particular job so that a fail is a pass and vice versa?

Added a commit https://github.com/ros-industrial/industrial_ci/pull/79/commits/2b89d8e2d85ab55ff5d4a4836101e1c590887683 for that. Not elegant at all though.

130s:
Sorry for some debug commits; those unnecessary ones are now removed.

Travis passed for [the job in question](https://travis-ci.org/ros-industrial/industrial_ci/jobs/157803108).

130s:
@evenator could you verify again? Sorry to bother you but much have changed since your last +1.

evenator:
This looks like it should work, but I don't like that the `_CI_INVERTERROR` option is only implemented for this particular case. I think it would be better to invert at the highest possible level (i.e. invert the output of the whole script).

Also, `_CI_INVERTERROR` isn't a very clear name to me. Maybe `_CI_EXPECT_FAIL`?

130s:
> I don't like that the _CI_INVERTERROR option is only implemented for this particular case. I think it would be better to invert at the highest possible level (i.e. invert the output of the whole script).

Actually that makes sense and I also thought that handling only particular case is too local. I opened a separate PR https://github.com/ros-industrial/industrial_ci/pull/80 and moved all associated changes there because implimenting the entire script sounds like a big project. 

Merging since the only remaining commit here is +1ed before.

PR #80, opened 2016-09-08
[WIP] Invert all commands result

130s:
**DO NOT MERGE YET**

Following the suggestion at https://github.com/ros-industrial/industrial_ci/pull/79#issuecomment-245067327, I open a separate PR.


ipa-mdl:
It might be easy to override bash exit function with an inverted variant:

```bash
function exit{
    code= ${1-0}
    if [ "$_CI_INVERTERROR" == true ] # better name: JOB_SHOULD_FAIL
    then 
        if [ "${1-0}"=="0" ]
        then
            code=1
        else
            code=0
        fi
   fi
   builtin exit $code
}
```

PR #81, opened 2016-10-22
Fix usermod error on docker-based ROS Prerelease

130s:
Without knowing, [`docker` install instruction](https://docs.docker.com/engine/installation/linux/ubuntulinux/) might have been updated to NOT use the user `ubuntu`. They [surely updated URL of the page](http://wiki.ros.org/action/info/regression_tests?action=diff&rev2=58&rev1=57).

This fixes issues like https://github.com/ros-industrial/ros_canopen/pull/193#issuecomment-254575036 (confirmed on my fork that with this change Travis passed https://github.com/130s/ros_canopen/pull/3).

### DIFF
--- a/industrial_ci/ros_pre-release.sh
+++ b/industrial_ci/ros_pre-release.sh
@@ -52,7 +52,7 @@ function install_ros() {
 }
 
 function setup_docker() {
-    sudo usermod -aG docker ubuntu
+    sudo usermod -aG docker $(whoami)
     # ROS Buildfarm for prerelease http://wiki.ros.org/regression_tests#How_do_I_setup_my_system_to_run_a_prerelease.3F
     sudo -E sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros-latest.list'
     sudo -E apt-key adv --keyserver hkp://pool.sks-keyservers.net --recv-key 0xB01FA116


ipa-mdl:
I would guess that Travis just migrated their VMs and just dropped the `ubuntu` user in favor of something else.

`$USER` (or even better `$(whoami)`) does not assume a specific username.

+1 for merge :)

130s:
> even better `$(whoami)`

Agreed. Switched to it. Once Travis passes it's mergeable.

BTW, the section of the code that caused the error https://github.com/ros-industrial/ros_canopen/pull/193#issuecomment-254575036 should be reached only when you enable `"$PRERELEASE" == true`, which at the time of the PR [is not enabled in `.travis.yml`](https://github.com/ros-industrial/ros_canopen/blob/cc20987898e149cbd60728d64dee612795c0087e/.travis.yml). Indeed in a successful [Travis build](https://travis-ci.org/130s/ros_canopen/builds/169791721) in my [forked repo](https://github.com/130s/ros_canopen/pull/3) it's not reached. I'm not sure if there's s uncovered bug.

130s:
> even better `$(whoami)`

Agreed ([ref](http://stackoverflow.com/questions/19306771/get-current-users-username-in-bash)). Switched to it. Once Travis passes it's mergeable.

BTW, the section of the code that caused the error https://github.com/ros-industrial/ros_canopen/pull/193#issuecomment-254575036 should be reached only when you enable `"$PRERELEASE" == true`, which at the time of the PR [is not enabled in `.travis.yml`](https://github.com/ros-industrial/ros_canopen/blob/cc20987898e149cbd60728d64dee612795c0087e/.travis.yml). Indeed in a successful [Travis build](https://travis-ci.org/130s/ros_canopen/builds/169791721) in my [forked repo](https://github.com/130s/ros_canopen/pull/3) it's not reached. I'm not sure if there's s uncovered bug.

ipa-mdl:
>  I'm not sure if there's s uncovered bug.

I have a dedicated test-release branch which the PR was targeted at:
https://github.com/ros-industrial/ros_canopen/blob/test-release/.travis.yml

130s:
> I have a dedicated test-release branch which the PR was targeted at:
> https://github.com/ros-industrial/ros_canopen/blob/test-release/.travis.yml

:relieved:

PR #82, opened 2016-12-02
fix environment variables passed to docker container

ipa-bnm:
`UPSTREAM_WORKSPACE` and `ROSINSTALL_FILENAME` was missing in `docker run` command. Using rosinstall files with kinetic is not possible without this.


130s:
+1. Thank you for picking it up, and also for removing an invalid argument `USE_DEBROS_DISTRO`.

Issue #64, opened 2016-06-14
ROS Prerelease Test option may not test what we want


130s:
Reported upstream https://github.com/ros-infrastructure/prerelease_website/issues/31
Can someone confirm this issue? If this holds then it's a shame that I didn't notice this...

130s:
[Confirmed in the upstream](https://github.com/ros-infrastructure/prerelease_website/issues/31):

> Currently, I believe it will always checkout code as defined in the `devel` section of the rosdistro distribution.yaml file.

This means the pull requests are not used in the prerelease test.

130s:
> This means the pull requests are not used in the prerelease test.

Add a temporary note https://github.com/ros-industrial/industrial_ci/pull/60/commits/a2a0e218416bcf58a7f414b9c3a594b3dde24204 in #60

ipa-mdl:
`generate_prerelease_script.py` has optional `--custom-branch` and `--custom-repo` parameters.

What do we want to test?
* for branch build: the branch
* for PR build: the remote branch?

ipa-mdl:
https://github.com/ipa-mdl/industrial_ci/commit/c875eacdc5ef05fdc6c4445c446065f939ae05d1 should do the trick, needs a source entry in ROS distribution.
Example: https://travis-ci.org/ipa-mdl/ros_canopen/builds/184373698

130s:
> `generate_prerelease_script.py` has optional --custom-branch and --custom-repo parameters.

Very good to know this! 

> https://github.com/ipa-mdl/industrial_ci/commit/c875eacdc5ef05fdc6c4445c446065f939ae05d1 should do the trick, needs a source entry in ROS distribution.

Would you open a PR?
Although I believe upstream https://github.com/ros-infrastructure/prerelease_website/issues/31 needs fixed, your workaround seems like a minimum hack against the existing upstream api that industrial_ci can take advantage from (and I think advantage exceeds disadvantage here).

> What do we want to test?
> 
> -    for branch build: the branch
> -    for PR build: the remote branch?

I've never thought of testing branches without PRs. Unless there's a need we can stick to PR only but I'm note opinionated here.

----

[...]

PR #83, opened 2016-12-15
fix kinetic prelease test

ipa-mdl:
Since kinetic tests are run in docker and the prerelease test uses docker as well, this resulted in a docker-in-docker configuration that did not work.

First I tried to make docker-in-docker work with travis, but I had to modify the xenial container a lot because of non-root user issues.

The simpler solution was to not run the prerelease test within docker.
(ros-jade-catkin must be installed for `catkin_test_results` in kinetic test)

Before: https://travis-ci.org/ros-industrial/ros_canopen/builds/184098605
After: https://travis-ci.org/ros-industrial/ros_canopen/builds/184216135

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ env:
     - ROS_DISTRO=indigo CATKIN_PARALLEL_JOBS='-p1' ROS_PARALLEL_JOBS='-j1'  # Intend build on low-power platform
     # - env: ROS_DISTRO=indigo PRERELEASE=true  ## Comment out because this is meaningless for always failing without prerelease testable contents in industrial_ci.
     - ROS_DISTRO=indigo PRERELEASE=true PRERELEASE_REPONAME=industrial_core PRERELEASE_DOWNSTREAM_DEPTH=0
-    - ROS_DISTRO=indigo PRERELEASE=true PRERELEASE_REPONAME=ros_canopen
+    - ROS_DISTRO=kinetic PRERELEASE=true PRERELEASE_REPONAME=ros_canopen
     - ROS_DISTRO=indigo APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet  # Passing wrong SKS URL as a break test. Should still pass.
     - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=debian
     - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=file  # Using default file name for ROSINSTALL_FILENAME
@@ -42,7 +42,6 @@ env:
 matrix:
   allow_failures:
     - env: ROS_DISTRO=indigo PRERELEASE=true PRERELEASE_REPONAME=industrial_core PRERELEASE_DOWNSTREAM_DEPTH=0
-    - env: ROS_DISTRO=indigo PRERELEASE=true PRERELEASE_REPONAME=ros_canopen
     - env: ROS_DISTRO=indigo UPSTREAM_WORKSPACE=file USE_DEB=true  # Expected to fail. See https://github.com/ros-industrial/industrial_ci/pull/74
     - env: ROS_DISTRO=indigo ADDITIONAL_DEBS="ros-hydro-opencv3"  # This should fail (trying from a wrong distro).
     - env: ROS_DISTRO=jade   PRERELEASE=true PRERELEASE_REPONAME=warehouse_ros PRERELEASE_DOWNSTREAM_DEPTH=1  # Intended to fail (as of Apr 2016), to test the capability of capturing Prerelease Test failure.
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ env:
     - ROS_DISTRO=indigo CATKIN_PARALLEL_JOBS='-p1' ROS_PARALLEL_JOBS='-j1'  # Intend build on low-power platform
     # - env: ROS_DISTRO=indigo PRERELEASE=true  ## Comment out because this is meaningless for always failing without prerelease testable contents in industrial_ci.
     - ROS_DISTRO=indigo PRERELEASE=true PRERELEASE_REPONAME=industrial_core PRERELEASE_DOWNSTREAM_DEPTH=0
-    - ROS_DISTRO=kinetic PRERELEASE=true PRERELEASE_REPONAME=ros_canopen
+    - ROS_DISTRO=kinetic PRERELEASE=true PRERELEASE_REPONAME=std_msgs
     - ROS_DISTRO=indigo APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet  # Passing wrong SKS URL as a break test. Should still pass.
     - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=debian
     - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=file  # Using default file name for ROSINSTALL_FILENAME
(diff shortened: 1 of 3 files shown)


ipa-mdl:
For now all non-prerelease test fail, like for the master branch.
I tried to fix this in #84

.travis.yml
```diff
@@ -42,7 +44,6 @@ env:
 matrix:
   allow_failures:
     - env: ROS_DISTRO=indigo PRERELEASE=true PRERELEASE_REPONAME=industrial_core PRERELEASE_DOWNSTREAM_DEPTH=0
-    - env: ROS_DISTRO=indigo PRERELEASE=true PRERELEASE_REPONAME=ros_canopen
```
130s: Because there's an unintuitive limitation with prerelease test feature https://github.com/ros-industrial/industrial_ci/issues/64 (I've been wanting to send a patch to the upstream but that has never occurred yet..), I kind of want to keep all prerelease jobs in `allow_failures` until it gets resolved. Thoughts?

130s:
> Since kinetic tests are run in docker and the prerelease test uses docker as well, this resulted in a docker-in-docker configuration that did not work.

U, oh! Thanks for catching that. This should definitely need to be pulled soon.

ipa-mdl:
I have remove the indigo and jade prerelease tests as they don't cover anything special.
This should save 30 minutes of test duration..

industrial_ci/ros_pre-release.sh
```diff
@@ -71,7 +81,8 @@ function run_ros_prerelease() {
     travis_time_end  # setup_docker
 
     travis_time_start setup_prerelease_scripts
-    mkdir -p /tmp/prerelease_job; cd /tmp/prerelease_job; generate_prerelease_script.py https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/production/index.yaml $ROS_DISTRO default ubuntu trusty amd64 ${PRERELEASE_REPONAME} --level $PRERELEASE_DOWNSTREAM_DEPTH --output-dir ./
+
```
ipa-mdl: What do you mean?

industrial_ci/ros_pre-release.sh
```diff
@@ -40,38 +40,40 @@ if [ ! "$PRERELEASE_DOWNSTREAM_DEPTH" ]; then export PRERELEASE_DOWNSTREAM_DEPTH
...
-    travis_time_start setup_docker
-    setup_docker
-    travis_time_end  # setup_docker
+    travis_time_start setup_environment
+    setup_environment
+    travis_time_end  # setup_environment
 
     travis_time_start setup_prerelease_scripts
-    mkdir -p /tmp/prerelease_job; cd /tmp/prerelease_job; generate_prerelease_script.py https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/production/index.yaml $ROS_DISTRO default ubuntu trusty amd64 ${PRERELEASE_REPONAME} --level $PRERELEASE_DOWNSTREAM_DEPTH --output-dir ./
+
```
130s: This line might be spurious (originally commented [at the older commit](https://github.com/ros-industrial/industrial_ci/pull/83#discussion_r92954580) but still seems to remain in the latest).

.travis.yml
130s: Aside from the recent discussion development in #64, the commit that gets used in the ros prerelease test is the latest in the development branch listed at `rosdistro`, which means that ros prerelease test as of now does not test your pull request by default. So until #64 is resolved we better not use prereleae for checking regression, no?

130s:
LGTM except for minor comments.
And could you tell us why the following was needed?

> (ros-jade-catkin must be installed for catkin_test_results in kinetic test)

.travis.yml
ipa-mdl: If it fails `industrial_ci` is broken, so I don't see the point in allowing the failure.

ipa-mdl:
> And could you tell us why the following was needed?
(ros-jade-catkin must be installed for catkin_test_results in kinetic test)

The prerelease tests needs `catkin_test_results` to check the results.
Since travis runs ubuntu trusty the `ros-kinetic-catkin` could not be installed.
http://wiki.ros.org/regression_tests#How_do_I_setup_my_system_to_run_a_prerelease.3F recommends to install `ros-jade-catkin` instead.

For long term support `ros-indigo-catkin`might be an alternative.

industrial_ci/ros_pre-release.sh
ipa-mdl: What's wrong with this line?

.travis.yml
130s: First and foremost, I was wrong in starting this discussion at this line. So +1 for removing this line.
(This particular line is removed only because [the corresponding line is also removed](https://github.com/ros-industrial/industrial_ci/pull/83/files#diff-354f30a63fb0907d4ad57269548329e3L27), totally irelevant from the discussion above. My bad.)

ipa-mdl: I removed the first line only recently ;)
Instead of testing `ros_canopen` we might test something more stable like `std_msgs`.
I doesn't really matter which package to prerelase-test in order to test our scripts.

industrial_ci/ros_pre-release.sh
130s: Yes, isn't it?

ipa-mdl: Did not notices that this patch chunk was all about the empty line, will remove it :)

130s:
Continued from https://github.com/ros-industrial/industrial_ci/pull/83#discussion_r92981349
For this PR, this isn't the case so we can move on without taking any action for this. 

> If it fails `industrial_ci` is broken, so I don't see the point in allowing the failure.

Because of the issue #64, the tested branch in the prerelease test upon a PR is different from the one in PR. While I agree that prerelease test shouldn't fail against any tip in a devel branch or HEAD, I feel it's a bit counter-intuitive that prerelease and non-prerelease tests in the same PR use different branches. Particularly, when an error occurs in prerelease, I don't want to force the one who opens the PR to fix an issue that s/he may have no idea of.

And there can be direct commits (see example below) that can introduce un-managed failure to HEAD (I've experienced that when I added corrupted change in a changelog commit). In this sense we might want to consider checking not only PRs (just discussed at https://github.com/ros-industrial/industrial_ci/issues/64#issuecomment-267904198).

Direct commit examples:
- `catkin_prepare_release` directly commits
- some/most maintainers directly commit changelog without opening a PR. 
- Also some (e.g. MoveIt! maintainers) chose to cherry-pick commits directly from a merged PR to another branches where applicable.

ipa-mdl:
full ack!

> Instead of testing ros_canopen we might test something more stable like std_msgs.

deal?

130s:
>>    Instead of testing ros_canopen we might test something more stable like std_msgs.
>
> deal?

Well, I thought testing a repo where industrial_ci is already used from makes it easier to narrow down issues if any (e.g. when a job that tests against for example ros_canopen in a PR on industrial_ci fails, then it's likely the PR is problematic, not the tested repo).

ipa-mdl:
> Well, I thought testing a repo where industrial_ci is already used from makes it easier to narrow down issues if any 

That's right for the source code tests with extra source dependencies etc..
But for the prerelease test we don't have to test the prerelease itself, we just have to make sure that we start it properly.

130s:
> for the prerelease test we don't have to test the prerelease itself, we just have to make sure that we start it properly.

Ah, you mean in my example we don't need to test if prerelease passes for ros_canopen? That makes sense!
If you could add that change to this PR we can merge it. Otherwise I'll open one later.

ipa-mdl:
I have now removed the empty line and switched to the prerelease test on std_msgs.

130s:
+1 
Once test passes you can merge @ipa-mdl (I need to go now).
Thanks a lot for iterating this!

PR #84, opened 2016-12-15
Fix tests and remove hard-coded script path

ipa-mdl:
The tests fails for master because somehow `catkin-tools` takes .ci_config as a package location.
(might be a bug..)

I took the chance to remove the need to name the directory `.ci_config`, which is the reason for copying the  files. In addition `ci_main.sh` does not rely on `travis.sh` anymore.


130s:
I'd like to test and see how things work by myself (later today hopefully), but the issue raised only occurs on Kinetic jobs doesn't it? If that's the case I have to admit that I might have not been as attentive as IJ jobs when I was writing the code...

ipa-mdl:
> but the issue raised here only occurs on Kinetic jobs doesn't it?

No, travis build fails for the master:
https://travis-ci.org/ros-industrial/industrial_ci/builds/180789707

ipa-mdl:
Now the kinetic non-prerelease test fail, I missed something..

ipa-mdl:
should be fixed now, we will see in 2 hours ;)

ipa-mdl:
I have extracted all(?) to-be-passed environment variables to docker.env.
The TRAVIS_* ones should be replaced by generic ones later..

ipa-mdl:
@130s: some jobs failed to download packages, can you restart them? 
https://travis-ci.org/ros-industrial/industrial_ci/jobs/184463311
https://travis-ci.org/ros-industrial/industrial_ci/jobs/184463312

https://travis-ci.org/ros-industrial/industrial_ci/jobs/184463320 is supposed to fail?

130s:
> some jobs failed to download packages, can you restart them?
> https://travis-ci.org/ros-industrial/industrial_ci/jobs/184463311
> https://travis-ci.org/ros-industrial/industrial_ci/jobs/184463312

Done.

@shaun-edwards Can we add @ipa-mdl admin (/ whatever necessary) privilege so that he can restart Travis jobs?

130s:
Seems reasonable, merging.

Except one thing that I'm curious; 

> The tests fails for master because somehow catkin-tools takes .ci_config as a package location. (might be a bug..)

I want to see how this happened. I doesn't find the associated SHA in https://travis-ci.org/ros-industrial/industrial_ci/pull_requests. If you could show the link to the job page that'll be great
(If it's something in the upstream we gotta fix in the upstream).

ipa-mdl:
https://travis-ci.org/ros-industrial/industrial_ci/builds/180789707
It cannot be related to the last PR.
I guess catkin was synced and therefore behaved differently.

shaun-edwards:
I've added @ipa-mdl as an admin (sorry for the delay).  For what it's worth, I think @130s has the ability to do this as well.  Go to  `Settings` -> `Collaborators & Teams`

ipa-mdl:
@shaun-edwards: Thanks!

PR #85, opened 2016-12-19
test prerelease from travis source

### LINKED ISSUE #64: ROS Prerelease Test option may not test what we want

ipa-mdl:
This patch runs prerelease tests on the job-bundled code (branch or merge) instead of the ROS distro devel branch.
For obvious reasons this does not work if another repository needs to be tested (i.e. `PRERELEASE_REPONAME` is set)
In this case the devel branch will be tested (as for the prerelease website).

Please note: The devel-job entry in ROS distro is still mandatory.

This might fix #64


130s:
+1

130s:
@ipa-mdl please consider https://github.com/ipa-mdl/industrial_ci/pull/1 that add a note in readme.

PR #88, opened 2016-12-29
Abort rosdep-install.sh on error

ipa-mdl:
This patch forces rosdep-install.sh to exit on errors.
Without -e the script exit code was taken from the manifest handling code.
(https://github.com/ros-industrial/universal_robot/pull/277#issuecomment-269709029)

### DIFF
--- a/industrial_ci/rosdep-install.sh
+++ b/industrial_ci/rosdep-install.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -x
+#!/bin/bash -ex
 
 # Software License Agreement (BSD License)
 #


industrial_ci/rosdep-install.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -x
+#!/bin/bash -ex
```
130s: If I were adding this change, I would add a comment why we need this because I can foresee I will easily forget why.
But having `-x` isn't common already, so it might be obvious enough.

ipa-mdl: Not sure how to properly comment a shebang..
Perhaps we should set it explicitly as in https://github.com/ros-industrial/industrial_ci/blob/master/industrial_ci/ci_main.sh#L38

130s: I opened https://github.com/ipa-mdl/industrial_ci/issues/3 to show what I will probably want to see in the future :)

ipa-mdl: That's rather confusing as "-ex" might look like "exit".
I will prepare another commit :)

130s:
Ok, then I'll wait for another PR from you for https://github.com/ros-industrial/industrial_ci/pull/88#discussion_r94221687. I'm merging because in case there are PRs that could be blocked by the issue this PR addresses. Thanks!

PR #90, opened 2017-01-03
[ci_main.sh] Fix to not terminate.

130s:
Thanks to https://github.com/ros-industrial/industrial_ci/commit/cec7f53ed5dc665534a887ef8a228dfcf3a584b7#commitcomment-20338023, I found that the jobs on Travis always ends when `ci_main.sh` finishes, due to which post process never occurs. This PR fixes it.

Also added a couple of more minor commits.


VictorLamoine:
:+1: all in for simplification!

industrial_ci/ci_main.sh
```diff
@@ -325,4 +325,4 @@ cd $TARGET_REPO_PATH  # cd back to the repository's home directory with travis
 pwd
 
 HIT_ENDOFSCRIPT=true
-success 0
```
ipa-mdl: I had to think twice why this would terminate the travis job. Then I noticed that we source all scripts..

VictorLamoine: I had similar issues in my GitLab script, you never want to call `exit` but instead `return` if you source scripts!

industrial_ci/test/example_ros.test
```diff
@@ -1,13 +0,0 @@
-<launch>
```
ipa-mdl: File is used in https://github.com/ros-industrial/industrial_ci/blob/master/industrial_ci/CMakeLists.txt#L9

130s: Thanks, fixed.

ipa-mdl: soI guess it should be moved instead of deteleted

industrial_ci/ci_main.sh
130s: Thanks for the headsup. If there any necessary change let us know further via the comments or PRs. IMO utility functions need to be called within the current process rather than in sub process, so `source`-ing is still the best way to use `util.sh`.

PR #92, opened 2017-01-11
Support private github repositories on kinetic

ipa-bnm:
To be able to clone private repositories inside the docker container, the ssh agent needs to be forwarded into the docker container.

### DIFF
--- a/industrial_ci/ci_main.sh
+++ b/industrial_ci/ci_main.sh
@@ -86,9 +86,6 @@ if [[ "$ROS_DISTRO" == "kinetic" ]] && ! [ "$IN_DOCKER" ]; then
 fi
 
 travis_time_start init_travis_environment
-#setup github.com ssh key
-mkdir -p ~/.ssh
-ssh-keyscan github.com >> ~/.ssh/known_hosts 2> /dev/null
 # Define more env vars
 BUILDER=catkin
 ROSWS=wstool
--- a/industrial_ci/ci_main.sh
+++ b/industrial_ci/ci_main.sh
@@ -75,12 +75,15 @@ if [[ "$ROS_DISTRO" == "kinetic" ]] && ! [ "$IN_DOCKER" ]; then
 
   docker_target_repo_path=/root/ci_src
   docker_ici_pkg_path=${ICI_PKG_PATH/$TARGET_REPO_PATH/$docker_target_repo_path}
-  docker run \
+  docker create \
+      --name run-industrial-ci \
       --env-file ${ICI_PKG_PATH}/docker.env \
       -e TARGET_REPO_PATH=$docker_target_repo_path \
       $SSH_DOCKER_CMD \
       -v $TARGET_REPO_PATH/:$docker_target_repo_path industrial-ci/xenial \
       /bin/bash -c "cd $docker_ici_pkg_path; source ./ci_main.sh;"
+  docker cp ~/.ssh run-industrial-ci:/root/ # pass SSH settings to container
+  docker start run-industrial-ci
   retval=$?
   if [ $retval -eq 0 ]; then HIT_ENDOFSCRIPT=true; success 0; else exit; fi  # Call  travis_time_end  run_travissh_docker
 fi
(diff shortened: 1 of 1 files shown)


industrial_ci/ci_main.sh
```diff
@@ -65,18 +65,30 @@ if [[ "$ROS_DISTRO" == "kinetic" ]] && ! [ "$IN_DOCKER" ]; then
   travis_time_end  # build_docker_image
 
   travis_time_start run_travissh_docker
+
+  #forward ssh agent into docker container
+  if [ "$SSH_AUTH_SOCK" ]; then
+      export SSH_DOCKER_CMD="-t -i -v $(dirname $SSH_AUTH_SOCK):$(dirname $SSH_AUTH_SOCK) -e SSH_AUTH_SOCK=$SSH_AUTH_SOCK"
```
ipa-mdl: Do you really need `-i` and -`t`?

industrial_ci/ci_main.sh
```diff
@@ -65,18 +65,30 @@ if [[ "$ROS_DISTRO" == "kinetic" ]] && ! [ "$IN_DOCKER" ]; then
...
       -e TARGET_REPO_PATH=$docker_target_repo_path \
+      $SSH_DOCKER_CMD \
       -v $TARGET_REPO_PATH/:$docker_target_repo_path industrial-ci/xenial \
       /bin/bash -c "cd $docker_ici_pkg_path; source ./ci_main.sh;"
   retval=$?
   if [ $retval -eq 0 ]; then HIT_ENDOFSCRIPT=true; success 0; else exit; fi  # Call  travis_time_end  run_travissh_docker
 fi
 
 travis_time_start init_travis_environment
+#setup github.com ssh key
```
ipa-mdl: Not sure if we should hardcode it here.
It is user specific and might be added via `.travis.yml` or `BEFORE_SCRIPT`.
However, adding `github.com` per default might be reasonable for many users.
@130s: any opinion on this?

industrial_ci/ci_main.sh
ipa-bnm: I've tried it without `-i` and `-t` without any luck. Seems to be necessary.

industrial_ci/ci_main.sh
130s: >  However, adding github.com per default might be reasonable for many users.

+1. It might be a good addition for private repo usecases.

industrial_ci/ci_main.sh
```diff
@@ -65,18 +65,30 @@ if [[ "$ROS_DISTRO" == "kinetic" ]] && ! [ "$IN_DOCKER" ]; then
...
+mkdir -p ~/.ssh
+ssh-keyscan github.com >> ~/.ssh/known_hosts
```
130s: Doesn't `ssh-keyscan` print the key content if it finds any? If so I'd suggest to make the output quiet.

ipa-mdl: The output gets piped into known hosts..

ipa-bnm: Yes, ssh-keyscan is gathering the public ssh host key. Output should already be quiet. The output of ssh-keyscan goes directly into `.ssh/known_hosts`.

130s: Ok, confirmed that the key is not printed (I should've done that first).

> ```
> ssh-keyscan github.com >> ~/.ssh/known_hosts 2> /dev/null
> ```

+1
I'm not a security expert so just prefer sticking to not disclosing network-related info as much as possible.

ipa-mdl: this line add the public key to the known hosts.
ssh will ask for interactive confirmation otherwise.

industrial_ci/ci_main.sh
ipa-mdl: I will try it again.

ipa-mdl:
https://github.com/ipa-bnm/industrial_ci/pull/2 needs to be merged first

PR #93, opened 2017-01-11
added flag to enable/disable bash trace output

ipa-bnm:


130s:
+1
@ipa-bnm please consider merging https://github.com/ipa-bnm/industrial_ci/pull/1. I think it's good to add this option to some of existing jobs to see how this works with `false`. 
Since there have already been many jobs in the .travis.yml I didn't want to add more.

industrial_ci/util.sh
```diff
@@ -35,13 +35,13 @@
 ## This is a script where the functions commonly used within the industrial_ci repo are defined.
 
 function travis_time_start {
-    set +x
+    if [ "$DEBUG_BASH" ] && [ "$DEBUG_BASH" == true ]; then set +x; fi
```
130s: Could this line be just `set +x` without if clause, because if `"$DEBUG_BASH" != true` then `set -x` must have not been done?

PR #94, opened 2017-01-11
added flag to enable/disable verbose output

ipa-bnm:


.travis.yml
```diff
@@ -49,7 +51,7 @@ matrix:
 before_script:
   - CI_DIR=.
   - export BEFORE_SCRIPT="rm -fr common_travis/CATKIN_IGNORE; ${BEFORE_SCRIPT}"
-  - if [ "${USE_CATKIN_MAKE}" == "true" ] ;then sed -i 's@catkin build -i -v --limit-status-rate 0.001@catkin_make@' $CI_DIR/travis.sh; fi
```
ipa-mdl: I would omit this chunk, `USE_CATKIN_MAKE` is broken anyway.

industrial_ci/ci_main.sh
```diff
@@ -38,6 +38,15 @@
 set -e # exit script on errors
 set -x # print trace
 
+#Define some verbose env vars
+if [ "$VERBOSE_OUTPUT" ] && [ "$VERBOSE_OUTPUT" == true ]; then
+    export OPT_V="-v"
+    export OPT_I="-i"
```
ipa-mdl: if you set both togethe then there is no need to split it

industrial_ci/ci_main.sh
```diff
@@ -70,6 +79,8 @@ if [[ "$ROS_DISTRO" == "kinetic" ]] && ! [ "$IN_DOCKER" ]; then
   docker run \
       --env-file ${ICI_PKG_PATH}/docker.env \
       -e TARGET_REPO_PATH=$docker_target_repo_path \
+      -e OPT_I=$OPT_I \
```
ipa-mdl: these should not be listed here

industrial_ci/ci_main.sh
```diff
@@ -82,16 +93,18 @@ BUILDER=catkin
 ROSWS=wstool
 # For compatibilility with hydro catkin, which has no --verbose flag
 CATKIN_TEST_RESULTS_CMD="catkin_test_results"
-if [ catkin_test_results --verbose 1>/dev/null 2>/dev/null; then CATKIN_TEST_RESULTS_CMD="catkin_test_results --verbose"; fi
```
ipa-mdl: IMHO catkin_test_results should always be run in verbose mode. It is the main result of the CI service.

ipa-mdl: If you really want to turn off verbosity, introduce a `VERBOSE_TEST_RESULTS` variable which defaults to `true`

industrial_ci/ci_main.sh
ipa-bnm: I've just would to give the possibility to use them desperately. But I see your point.

industrial_ci/ci_main.sh
ipa-bnm: Why not? We need to access the variables inside the docker container.

ipa-mdl: no, they get derived from `VERBOSE_OUTPUT` anyway

ipa-mdl:
@ipa-bnm: please rebase on master and squash the fixes

industrial_ci/ci_main.sh
```diff
@@ -38,6 +38,13 @@
 set -e # exit script on errors
 if [ "$DEBUG_BASH" ]; then set -x; fi # print trace if DEBUG
 
+#Define some verbose env vars
```
130s: Thanks for adding a comment. shell is too magic to me even now (and forever) so explanation per certain chunk of code is really helpful (I should've done that too...will enforce as my rule).

industrial_ci/ci_main.sh
```diff
@@ -129,7 +136,7 @@ sudo -E sh -c 'echo "deb $ROS_REPOSITORY_PATH `lsb_release -cs` main" > /etc/apt
 # apt key acquisition. Since keyserver may often become accessible, backup method is added.
 sudo apt-key adv --keyserver $APTKEY_STORE_SKS --recv-key $HASHKEY_SKS || ((echo 'Fetching apt key from SKS keyserver somehow failed. Trying to get one from alternative.\n'; wget $APTKEY_STORE_HTTPS -O - | sudo apt-key add -) || (echo 'Fetching apt key by an alternative method failed too. Exiting since ROS cannot be installed.'; error))
 lsb_release -a
-sudo apt-get update || (echo "ERROR: apt server not responding. This is a rare situation, and usually just waiting for a while clears this. See https://github.com/ros-industrial/industrial_ci/pull/56 for more of the discussion"; error)
+sudo apt-get -qq update || (echo "ERROR: apt server not responding. This is a rare situation, and usually just waiting for a while clears this. See https://github.com/ros-industrial/industrial_ci/pull/56 for more of the discussion"; error)
```
130s: +1. Just fyi this still prints the problematic URLs in case of failure so that the readers can address the issue if they want to.

130s:
Thank you for the change and iterating @ipa-bnm and @ipa-mdl. I'm doing squash merge through github feature.

Issue #73, opened 2016-07-08
Should 'rosdep install' be run without '-r'?


gavanderhoorn:
I guess this is again a question of 'when to fail': `rosdep` is run with the `-r` argument by `rosdep-install.sh` at the moment. That makes it continue when it encounters a problem while resolving / installing dependencies, which seems like a good thing (sometimes).

In the case of `staubli_experimental` (example), this means that the build will fail at a (much) later point, as `staubli_experimental` depends on packages in `staubli`, but those haven't been released, and the `staubli` source repository is not included in the workspace setup for `industrial_ci`. `rosdep` cannot resolve the packages, but continues.

I don't know whether fail-early would be good here: how many times has `-r` saved a build that would've otherwise failed?

See [ros-industrial/staubli_experimental/builds/143354933](https://travis-ci.org/ros-industrial/staubli_experimental/builds/143354933) for a failed build.

130s:
Here are my subjective view for supporting fail-later (i.e. `rosdep` continues regardless resource-not-found error):
- IMHO, first 2 things we might want to check with CI among many others are build and test.
 - Wrong dependency definition doesn't always stop the system; for example, you can define infinitely many inexistent, unused run_depend in your package.xml. While `rosdep` fails to resolve those, build process won't be affected as long as necessary dependencies are resolved.
- Once resource issue is fixed then resource error won't be likely to happen (until new dependency definition is added).
- Missing resource is almost guranteed to be captured somewhere in the entire test process (rosdep, build, test).

Downside includes that jobs last longer, and there can be multiple errors so spotting them becomes harder (I guess this is the case with `staubli`?).

As a workaround(? not sure if this is correct wording here), how about this; 

- When there was resolution errors during `rosdep-install.sh`, the "fold" section turns to red, but the whole script goes on. This hopefully makes locating if error occurred during rosdep process a little easier.

Hm, not very good. Ideas welcomed.

ipa-mdl:
TL;DR: rosdep failures should stop the test immediately.

> you can define infinitely many inexistent, unused run_depend in your package.xml

This contradicts all ROS-I quality demands.
Most likely inexistent dependencies are the result of typos or outdated keys etc.

> Missing resource is almost guranteed to be captured somewhere in the entire test process 

The CI tests should better give a good indication which parts of the chain did fail.
If I made an error with the rosdep keys, I expect rosdep to fail. Not cmake and definitely not the compilation.

[...]

PR #95, opened 2017-01-12
Simplify rosdep, do not use robust flag

ipa-mdl:
As pointed out in #73, rosdep should not use `-r` for CI builds.
This PR even removes the manifest handling code.
Not sure why this was added at all.


ipa-mdl:
> Not sure why this was added at all.

I will add a testcase for this.

ipa-mdl:
I have tested it with a simple mockup (#96) test.

With this PR: https://travis-ci.org/ipa-mdl/industrial_ci/jobs/191319013:
* rosdep install dependencies from manifest.xml
* catkin detects manifest.xml and fallbacks to rosbuild, this fails because `ROS_PACKAGE_PATH` is not set-up properly

Wthout this PR: https://travis-ci.org/ipa-mdl/industrial_ci/jobs/191347515:
* rosdep uses `package.xml`
* manifest handling seems to fail anyway, I guess the script does not handle the symlink properly

Hybrid manifest package with master: https://travis-ci.org/ipa-mdl/industrial_ci/jobs/191351917:
* rosdep uses `package.xml`
* catkin behaves like in first case

manifest.xml breaks the tests in any case, this is not related to rosdep

So, I am going to merge this for the sake of simplicity.
And open a new issue for manifest.xml handling.

130s:
Thank you @ipa-mdl for the experiment https://github.com/ros-industrial/industrial_ci/pull/95#issuecomment-272205293! I didn't know `manifest.xml` wasn't handled as intended.

I'll make a new tag (yes I've been doing so) since now we removed a feature that regardless was not functioning, and send a brief update announce to swri ML.

ipa-mdl:
> and send a brief update announce to swri ML.

 Please point to https://github.com/ros-industrial/industrial_ci/issues/97 as well.

130s:
Sent [an announcement](https://goo.gl/bmXGsE) but sorry @ipa-mdl, I mispelled your name...

PR #96, opened 2017-01-12
implemented mockup tests feature

ipa-mdl:
This feature implements a generic mockup feature.
Mockup packages can used to test certain aspect and even might be valid packages.
This can be used together with `allowed_failures` or (even better) #80 
So mockup packages are ignores by default can be enabled by `USE_MOCKUP` varibale which points to the mockup directoy.

Per default  `industrial_ci_testpkg`is added a mockup (it was moved accordingly).


.travis.yml
```diff
@@ -16,7 +16,10 @@ notifications:
 env:
   global:
     - DEBUG_BASH='true' # Print bash trace
+    - USE_MOCKUP='mockups/industrial_ci_testpkg' # add a test package by default
   matrix:
+    - ROS_DISTRO=indigo USE_MOCKUP=''
+    - ROS_DISTRO=indigo NOT_TEST_BUILD='true'
```
130s: Is this line duplicating a line below?

130s:
So, you're adding `mockup` directory (and move industrial_ci_testpkg there) to clarify that it's a location for the packages that are used for testing `industrial_ci`.If that's correct +1.

For the naming of the directory, I've seen examples where packages of the same kind are stored under `test` folders (e.g. https://github.com/ros/ros_comm/tree/kinetic-devel/test). So I thought naming `test` might be more intuitive and common. However, because in our case it's not located within a package, it might not matter. I don't have a strong opinion but what do you think?

Also, why not add packages you used in https://github.com/ros-industrial/industrial_ci/pull/95#issuecomment-272205293 and maybe add the job as allow_failure?

>  This can be used together with ... (even better) #80

I'm not sure what you mean here.

130s:
> Also, why not add packages you used in https://github.com/ros-industrial/industrial_ci/pull/95#issuecomment-272205293 and maybe add the job as allow_failure?

I take that back since it doesn't do anything worthy now that we've removed `manifest.xml` handling.

ipa-mdl:
>> This can be used together with ... (even better) #80

> I'm not sure what you mean here.

If some mockup packages should fail, it might be an option to allow them fail.
However, it might be even better to check that they fail.

> So, you're adding mockup directory (and move industrial_ci_testpkg there) to clarify that it's a location for the packages that are used for testing industrial_ci.

yes, that's correct

> For the naming of the directory, I've seen examples where packages of the same kind are stored under test

`test` is okay for me as well.
I chose `mockup`, because `test` seems a little bit misleading in a CI package, but this was spontaneous decision.
Just tell me, what name you prefer. I will update it and resolve the conflicts.

> > Also, why not add packages you used in #95 (comment) and maybe add the job as allow_failure?

> I take that back since it doesn't do anything worthy now that we've removed manifest.xml handling.

If it would not have failed, I would have added it.
We need more unit tests for sure!
(I will add a CI fusion roadmapping issue soon, which will address this as well)

.travis.yml
ipa-mdl: fixed in latest version

ipa-mdl:
> I chose mockup, because test seems a little bit misleading in a CI package, but this was spontaneous decision.

In addition `mockup` corresponds to `USE_MOCKUP`.
Perhaps `mockups` would be better or `_mockups`.

130s:
All make sense. Thank you!

PR #100, opened 2017-01-14
[WiP] fix for exit code issues

### LINKED ISSUE #99: travis.sh terminates travis job, no post-process script is run
I figured that the postprocess is never run because we `exit` from a sourced script.
Why do we source travis.sh instead of calling it?

ipa-mdl:
fixes #99


ipa-mdl:
This was a little bit tricky..

`success` seems not to work as expected, but I was not able to find the error.
Normal builds work quite well without calling `success`at the end,  but `set +x` should better be called..

I have removed the process scripts as they don't really test something, but add 20 seconds to each run.
`sl` might be fun in a console, but it floods the travis logs.
As an replacement I have added a unit test that enforces the post step to fail.

With the post process working I noticed that the folding style was still around with them.
That's why I changed the prefix.

Finally `success` did not work properly for docker-based tests.
To work around I have factored out the actual test  out of `ci_main.sh`, this was planned for #98  anyway.

130s:
I assume it's no longer WIP despite the title, so merging.

ipa-mdl:
> I assume it's no longer WIP despite the title, so merging.

I just forgot to remove it..

PR #101, opened 2017-01-14
Speedup tests

ipa-mdl:
This PR simplifies the test candidates to ultra stable packages with only few dependencies.
Test job duration will decrease from >1h to ~15 minutes.

Please note: The travis job for this PR will take about 20 minutes because job ~~15~~16 still uses the rosinstall file from the master branch.

### DIFF
--- a/.ci.rosinstall.jade
+++ b/.ci.rosinstall.jade
@@ -1,4 +1,3 @@
 - git:
-    uri: https://github.com/ros-industrial/industrial_core.git
-    local-name: ros-industrial/industrial_core
-    version: jade
+    uri: https://github.com/ros/actionlib.git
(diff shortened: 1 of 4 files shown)


ipa-mdl:
rebased.

130s:
I already merged, but for the sake of argument; I used `industrial_core` as a downstream test for `industrial_ci`. Because client repos will be affected immediately once a PR is merged on industrial_ci, I thought we also want to spot potential bugs early before applying changes. This PR removes all downstream testings AFAIK. How do you think?

ipa-mdl:
> This PR removes all downstream testings AFAIK. How do you think?

We can add `industrial_core` to one test, but not for all.

PR #102, opened 2017-01-15
attach docker container to stdout/err again

ipa-mdl:
kinetic tests were run, but output was hidden.


ipa-mdl:
simple hot fix, all tests passed -> merged

PR #103, opened 2017-01-16
More fixes for error handling

### LINKED ISSUE #99: travis.sh terminates travis job, no post-process script is run

ipa-mdl:
It turned out that #100 did not fix #99 completely.
`error()` did exit the travis-generated script as well.
Lesson learned: never `exit` from a `source`d script.

So `ci_main.sh` does not get sourced anymore; environment is no longer exposed.
That's why I have added `AFTER_SCRIPT`.
Please note: Additional scripts added in .travis.yml do not get executed in docker container!

Travis executes all `script:` lines regardless of the exit code, but marks the build as failed if one line failed.
However, this might change: https://github.com/travis-ci/travis-ci/issues/1066#issuecomment-257441420


ipa-mdl:
now it was easy to implement an exit code check (supersedes #80)

ipa-mdl:
Perhaps `AFTER_SCRIPT` should only be executed in `source_tests.sh` (like `BEFORE_SCRIPT`).

ipa-mdl:
rebased

industrial_ci/util.sh
```diff
@@ -86,15 +86,14 @@ function _end_fold_script {
...
+    if ! [ -z $TRAVIS_FOLD_NAME ]; then
         ici_time_end $color_wrap
     else
 	echo "Previous Travis fold name not found. It might be either successful termination of the script, or wrong call. Skipping 'ici_time_end' anyway."
     fi
 
-    if [ $exit_code -eq "1" ]; then trap - ERR; fi  # Reset signal handler since the shell is about to exit.
     if [ "$DEBUG_BASH" ] && [ "$DEBUG_BASH" == true ]; then set -x; fi
-    if [ $exit_code -ne "-1" ]; then exit $exit_code; fi
+    if [ $exit_code -ne "-1" ]; then ici_exit $exit_code; fi
```
130s: Depending on situation, calling this function can fail where `ici_exit` is not yet loaded.

ipa-mdl: I cannot think of a situation like this, do you have an example.
I can move `ici_exit`to the top if you prefer it..

ipa-mdl:
Do not merge yet, I'd like to add error tracing and the error function needs to get modified for exit code propagation

industrial_ci/util.sh
```diff
@@ -86,15 +86,14 @@ function _end_fold_script {
     color_wrap=${2:-32}
 
     if [ $exit_code -eq "1" ]; then color_wrap=31; fi  # Red color
-    if [ -z $TRAVIS_FOLD_NAME ]; then
+    if ! [ -z $TRAVIS_FOLD_NAME ]; then
```
130s: If this change is intended, the following else block may no longer make sense.

industrial_ci/ci_main.sh
```diff
@@ -52,7 +52,7 @@ export HIT_ENDOFSCRIPT=false
 
 source ${ICI_PKG_PATH}/util.sh
 
-trap error ERR
+trap error EXIT
```
130s: Do we want to create a new function `exit`? Isn't trapping `EXIT` with `error` function misleading?

ipa-mdl: Good idea, will add this later

industrial_ci/util.sh
130s: I don't have a usecase for now, but only because this file is "util" we better address all the possibility we can think of.
Moving `ici_exit` sounds good to me.

industrial_ci/util.sh
ipa-mdl: It is intended, but the else block still makes sense.
` if ! [ -z $TRAVIS_FOLD_NAME ]` -> `if $TRAVIS_FOLD_NAME is not empty` -> `if fold is open`
This is the bug I did not find in #100

industrial_ci/util.sh
ipa-mdl: I guess `ici_exit` will be refactored to become the exit handler

ipa-mdl:
> Does this change require to change .travis.yml in users' repos, from like source .ci_config/travis.sh to .ci_config/travis.sh

No, `source travis.sh` is still supported. We might add a warning in this case since the enviroment is not exported anymore

> Sorry but I can't remember why BEFORE_SCRIPT is run from source_tests.sh

This way it is called after ROS and workspace set-up.

> Perhaps AFTER_SCRIPT should only be executed in source_tests.sh (like BEFORE_SCRIPT).

I guess this will change with #98 anyway: It works for now, we should not touch it for now.

industrial_ci/util.sh
130s: Ah, I see. Turned out I've been consistently misunderstanding `if -z` option.

130s:
I see. +1 from me.
Minor todos can be addressed in the future in you like to merge this one first @ipa-mdl.

industrial_ci/util.sh
ipa-mdl: `if [ -n $TRAVIS_FOLD_NAME ]` ("not empty") might be easier to understand ^^

ipa-mdl:
Error-code related refactoring completed :)
Now every fold lists an exit code and `error` forwards the exit code if possible.
`ici_exit` is the default exit handler and closes open folds

industrial_ci/util.sh
```diff
@@ -51,93 +51,76 @@ function ici_time_start {
...
-    	if [ $_exit_code -eq 0 ]; then
-    	    echo "${_FUNC_MSG_PREFIX} Arg HIT_ENDOFSCRIPT must be true when this function exit with 0. Turn _exit_code to 1."; _exit_code=1;
-    	else
-    	    echo "${_FUNC_MSG_PREFIX} _exit_code cannot be 0 for this func. Make sure you are calling this in a right context."; _exit_code=1;
-    	fi
+    local exit_code=${2:-$?} #
+    if [ -n $1 ]; then
+        echo "\e[31m$1\e0m" # print error in red
+    fi
+    if [ "$exit_code" == "0" ]; then # 0 is not error
```
130s: I assume this if block is added to handle mis-use? If so printing so would be friendly to the users?

industrial_ci/util.sh
```diff
@@ -51,93 +51,76 @@ function ici_time_start {
...
+        if [ $exit_code -ne "0" ]; then color_wrap=31; fi  # Red color for errors
+        ici_time_end "$color_wrap" "$exit_code"
     fi
 
-    if [ $exit_code -eq "1" ]; then trap - ERR; fi  # Reset signal handler since the shell is about to exit.
-    if [ "$DEBUG_BASH" ] && [ "$DEBUG_BASH" == true ]; then set -x; fi
-    if [ $exit_code -ne "-1" ]; then exit $exit_code; fi
+    if [ "$exit_code" == "${EXPECT_EXIT_CODE:-0}" ]; then
+        exit 0
+    elif [ "$exit_code" == "0" ]; then # 0 was not expected
```
130s: Same as https://github.com/ros-industrial/industrial_ci/pull/103/files#r96508040

130s:
Left a minor comment.

>  Now every fold lists an exit code and error forwards the exit code if possible.

That looks convenient!

industrial_ci/util.sh
ipa-mdl: exit_code is read from $?, which might be 0 in some cases, like `if ! false ; then echo "$?"; fi`.
So the following code works now as expected:
```bash
if ! [ -e "$my_path" ]; then
  error "path does not exist: '$my_path'"
fi
```
I think, the documentation needs to be updated..

industrial_ci/util.sh
ipa-mdl: No, this is another use case.

If exit_code != EXPECT_EXIT_CODE  then the exit_code will be used.
However, If exit_code is 0, but EXPECT_EXIT_CODE is not 0, the exit_code can not be returned as 0 would mean success.

industrial_ci/util.sh
ipa-mdl: Alternative implementation, with exit_code passing:
```
test -e "$my_path" || error "path does not exist: '$my_path'"
```

ipa-mdl:
If you do not object, I will merge this PR tomorrow (= in 10 hours) :)

PR #86, opened 2016-12-21
[WIP][prerelease] Test all supported Ubuntu distros for the specifid ROS distro.

130s:
This PR is based on #85 so once it's merged this needs rebased to master branch.

There are cases where ROS buildfarm failure occurs on certain Ubuntu distros (e.g. https://github.com/ros-planning/moveit/issues/18#issuecomment-260867116). For more complete coverage the prerelease feature better tests all supported distros (at least Ubuntu) in [REP-0003](http://www.ros.org/reps/rep-0003.html).

PR #104, opened 2017-01-17
Add maintainer

130s:
<s>See https://github.com/ros-industrial/industrial_ci/pull/101#issuecomment-272899368
I'm adding back `industrial_core` to a job that uses .rosinstall file on Kinetic. Rationale:
- Now that `industrial_core` is released into Kinetic. Kinetic will be maintained longer, and we want more Kinetic samples.
- It's allow_failure so that if in hurry we can<s>'t</s> still ignore this particular job.
</s>

Also added is the maintainer addition (picked from https://github.com/ros-industrial/industrial_ci/pull/86#issuecomment-268507950, which needs reconsideration and therefore on a hiatus now).

### DIFF
--- a/industrial_ci/package.xml
+++ b/industrial_ci/package.xml
@@ -4,7 +4,9 @@
   <description>common travis script suite</description>
   <license>BSD</license>
   <maintainer email="iisaito@kinugarage.com">Isaac I. Y. Saito</maintainer>
-  <author email="iiysaito@opensource-robotics.tokyo.jp">Isaac I. Y. Saito</author>
+  <maintainer email="mathias.luedtke@ipa.fraunhofer.de">Mathias Lüdtke</maintainer>
+  <author>Isaac I. Y. Saito</author>
+  <author email="mathias.luedtke@ipa.fraunhofer.de">Mathias Lüdtke</author>
 
   <buildtool_depend>catkin</buildtool_depend>
 


130s:
> - It's allow_failure so that if in hurry we can't still ignore this particular job.

CORRECTION; we **can** ignore

.travis.rosinstall
```diff
@@ -1,3 +1,4 @@
 - git:
-    uri: https://github.com/ros/std_msgs.git
```
ipa-mdl: This will be used on indigo as well.
Please add a new `.travis.rosinstall.kinetic` file for this purpose, this will test this feature as well :)

.travis.yml
```diff
@@ -45,13 +45,15 @@ env:
     - ROS_DISTRO=jade   PRERELEASE=true PRERELEASE_REPONAME=i_do_not_exist PRERELEASE_DOWNSTREAM_DEPTH=1  # Intended to fail, to test the capability of capturing Prerelease Test failure.
     - ROS_DISTRO=kinetic
     - ROS_DISTRO=kinetic NOT_TEST_BUILD='true' DEBUG_BASH='false' VERBOSE_OUTPUT='false'
+    - ROS_DISTRO=kinetic UPSTREAM_WORKSPACE=https://raw.githubusercontent.com/ros-industrial/industrial_ci/master/.travis.rosinstall.kinetic
```
ipa-mdl: please use `file` instead to cover https://github.com/ros-industrial/industrial_ci/blob/master/industrial_ci/source_tests.sh#L165 in a test

ipa-mdl:
Please do not merge it as is.

130s:
Now this PR is gone WIP, will probably needs some more work before review-ready.

> For downstream tests the industrial_core packages (or at least candiate) should be added to `PKGS_DOWNSTREAM`

Good point. 

- Then I noticed that there's no way that the packages that are specified by `PKGS_DOWNSTREAM` get downloaded. So I'm trying to re-use the code that was for `UPSTREAM_WORKSPACE`.
- Also doc was outdated for some variables.

ipa-mdl:
IMHO `PKGS_DOWNSTREAM` should be kept as is, it is very powerfull as is might select only a subset of tests.
Even packages from the upstream might be downstream package with regard to tests.
However,  I like the idea of a `DOWNSTREAM_WORKSPACE` variable.

ipa-mdl:
I would suggest to mix the downstream package issues with this PR.
There are some more issues with the overall behaviour. We should discuss this in a separate issue.
(I will open one)

130s:
Ok, I've the most simplified this PR, even changed the subject :)

ipa-mdl:
Thanks :)

Issue #44, opened 2016-05-13
pkg-config Installed By Default in Travis Environment, not on OSRF Build Farm


evenator:
We recently had a package that required pkg-config as a build tool, but we failed to add it as a dependency in the package.xml. It built fine on developer machines (because we all have pkg-config) and on Travis (because it has pkg-config), but then failed on jenkins.ros.org (which does not have pkg-config installed).

Would it be possible to remove pkg-config from the Travis environment to more accurately reflect the OSRF farm environment?

ipa-mdl:
This dependency is pulled in by `rospack`, which is used several times for printing debug information throughout the script.
Have you tried the [prerelase test option] (https://github.com/ros-industrial/industrial_ci#optional-run-ros-prerelease-test)?

130s:
Out of curiosity (and possible future enhancement on industrial_ci), could you link to the build error on jenkins you saw?

pjreed:
The issue here was a little tricky for us to figure out.  Here's the build that failed; it was the swri_transform_util package in the marti_common repository:
http://build.ros.org/job/Jbin_uT64__swri_transform_util__ubuntu_trusty_amd64__binary/7/

The reason it failed is because the linker couldn't find various yaml-cpp references that were transitively required by swri_yaml_util.

Next, take a look at the build for swri_yaml_util:
http://build.ros.org/job/Jbin_uT64__swri_yaml_util__ubuntu_trusty_amd64__binary/5/

The build succeeded, but because pkg-config was not included in its dependencies, it wasn't installed, and so the appropriate linker flags did not get passed on to other packages that depended on it.

ipa-mdl:
This was fixed in #108

Issue #17, opened 2016-01-12
License of the package


130s:
I thought I would change the license in a trial to conform ROS-I recommendation (apache 2.0), but the repository I got much inspiration from (`jsk_travis`) is BSD (and some scripts from there are simply redistributed here), I like to hold back.

If there's any other opinions, please reopen and discuss.

130s:
Just happened to find [Apache2.0 LICENSE file](https://github.com/ros-industrial/industrial_ci/blob/master/LICENSE), which Shaun created when he initiated this repo. I/we gotta figure out what license we use.

gavanderhoorn:
IANAL, but AFAIK, you can relicense artefacts, but you'll have to get permission from the original authors.

Also: it's not a requirement to have all artefacts under the same license. It's perfectly possible for some bits to be BSD, with others under Apache 2. And apparently BSD and Apache 2 are very compatible with each other.

Perhaps @ipa-mirb can refer us to some relevant documents?

shaun-edwards:
I have the same understanding as @gavanderhoorn.  I would propose that we keep existing licenses as is and require that any new "code" be licensed under Apache 2.0

ipa-mirb:
Unfortunately I do not have specific documents to point you to, as last time we dealt with licensing at the "OSS - legal aspects" event here at IPA we discussed more general terms. But from what I understood mixing BSD and Apache should not be a problem (whereas e.g., LGPL and BSD could be, I am afraid).

130s:
As I noted in https://github.com/ros-industrial/industrial_ci/pull/108#issuecomment-274114279, now there are only files that were created within this repo so that we have full control of the license for this package/repo (INAL but I hope so). So in https://github.com/ros-industrial/industrial_ci/pull/113 I changed the package license to Apache 2.0.

PR #108, opened 2017-01-18
Clean up

ipa-mdl:
This PR address #98.
It removes a lot of outdated code and doubled line, but perhaps not all.
For now it is meant for discussion and probably will not be merged as is.

Some remarks:

> do not print bash commands per default

The +x setting makes the travis log unreadable, this PR disables it in general, but enables it for 2 tests.

> remove outdated USE_DEVEL_SPACE tests

`USE_DEVEL_SPACE` not handled anymore

> do not disable meta-packages

There's no reason to disable them. They should get tested as well (see https://github.com/ros-industrial/industrial_core/pull/169 as well)

> removed MongoDB hack

I think its broken and not needed anymore.

> stripped external dependencies

addresses #44 

> removed git diff code, worked only for PRs

Github can be used for it..

> remove hydro patches

I think the patches are jsk-specific and not needed.
IMHO the test enviroment should replicate a standard hydro envirment
Special handling could be done in a `BEFORE_SCRIPT`

> simplified printing of test results

Not sure about this.
But I believe that `catkin_test_results` tests all valid locations..

**We should add a hydro downsteam test**


ipa-mdl:
`simplified printing of test results` should not be included or refactored for hydro support.
The hydro version of `catkin_test_results` differs to much.

ipa-mdl:
somethin went wrong with my push, need to fix that tomorrow.

ipa-mdl:
fixed incorrect push and rebased

industrial_ci/CMakeLists.txt
```diff
@@ -2,5 +2,5 @@ cmake_minimum_required(VERSION 2.8.3)
 project(industrial_ci)
 find_package(catkin REQUIRED)
 catkin_package()
-install(FILES check_metapackage.py ci_main.sh
```
130s: metapackage removal was introduced for this thoughts https://github.com/jsk-ros-pkg/jsk_travis/issues/117, which honestly I'm not yet sure what is the problem. So, we *may* see the same issue with that ticket but let's face it.

industrial_ci/source_tests.sh
```diff
@@ -104,22 +101,13 @@ sudo apt-key adv --keyserver $APTKEY_STORE_SKS --recv-key $HASHKEY_SKS  \
...
 # If more DEBs needed during preparation, define ADDITIONAL_DEBS variable where you list the name of DEB(S, delimitted by whitespace)
 if [ "$ADDITIONAL_DEBS" ]; then
     sudo apt-get install -q -qq -y $ADDITIONAL_DEBS || error "One or more additional deb installation is failed. Exiting."
 fi
-
-# For compatibilility with hydro catkin, which has no --verbose flag
-CATKIN_TEST_RESULTS_CMD="catkin_test_results"
-if catkin_test_results --verbose 1>/dev/null 2>/dev/null; then CATKIN_TEST_RESULTS_CMD="catkin_test_results --verbose"; fi
-
-# MongoDB hack - I don't fully understand this but its for moveit_warehouse
```
130s: +1 for removing mongo hack, which I don't fully understand but there has been a restoration in MoveIt! last year https://github.com/ros-planning/warehouse_ros/issues/20 so hopefully it's no longer needed.

industrial_ci/source_tests.sh
```diff
@@ -104,17 +104,18 @@ sudo apt-key adv --keyserver $APTKEY_STORE_SKS --recv-key $HASHKEY_SKS  \
 
 sudo apt-get -qq update || error "ERROR: apt server not responding. This is a rare situation, and usually just waiting for a while clears this. See https://github.com/ros-industrial/industrial_ci/pull/56 for more of the discussion"
  
-sudo apt-get -qq install -y python-catkin-tools python-rosdep python-wstool ros-$ROS_DISTRO-rosbash ros-$ROS_DISTRO-rospack
+sudo apt-get -qq install --no-install-recommends -y build-essential python-catkin-tools python-rosdep python-wstool ros-$ROS_DISTRO-catkin
+
```
130s: So is the idea that follow the standard ROS source installation http://wiki.ros.org/indigo/Installation/Source and remove custom steps AMAP?

industrial_ci/source_tests.sh
```diff
@@ -246,7 +228,7 @@ if [ "$NOT_TEST_BUILD" != "true" ]; then
     fi
 
     if [ "$BUILDER" == catkin ]; then
-        source devel/setup.bash ; rospack profile # force to update ROS_PACKAGE_PATH for rostest
```
130s: Looks like this was originally added at https://github.com/jsk-ros-pkg/jsk_travis/pull/164
Trying to figure out why we no longer need this, but I'm not sure yet.

.travis.yml
```diff
@@ -15,7 +15,6 @@ notifications:
       - gm130s@gmail.com  # Change this to yours should you copy this file.
 env:
   global:
-    - DEBUG_BASH='true' # Print bash trace
```
130s: For now +0 since when errors occur users now see less output. But in general I prefer brevity too, s let's apply this change in and see how users react.

130s:
Left some comments, all of which are "let's see how we'll do with these", particularly because again I'm all for simplifying the scripts. So +1 for merging.

Btw, once metapackage.sh is removed, we would no longer have license-headeredd files that were copied from the original forked repo, which is why this package is BSD. Now we'll be also able look into https://github.com/ros-industrial/industrial_ci/issues/17

industrial_ci/source_tests.sh
ipa-mdl: I tried to remove all dependencies that are not required for the actual test.

industrial_ci/source_tests.sh
ipa-mdl: rospack is mostly used for rosbuild packages..

.travis.yml
ipa-mdl: I was not able to find the errors in the +x enabled code ;)
It easy to add for a certain test.

industrial_ci/CMakeLists.txt
ipa-mdl: I think metapackages should not have build_depends anyway.

PR #109, opened 2017-01-18
fix unset of TRAVIS_FOLD_NAME

ipa-mdl:
Current master leads to doubled `after_script` fold closings because `TRAVIS_FOLD_NAME` is not unset properly.


ipa-mdl:
minimal change ;), tests pass and output is correct now -> merge

PR #113, opened 2017-01-20
[doc] Split readme into quick start and detail.

130s:
<s>**DO NOT MERGE YET**.</s>

Addresses part of https://github.com/ros-industrial/industrial_ci/issues/105

PR #115, opened 2017-01-23
[doc] Correction; .ci_config is NO LONGER a required name.

130s:
Pointed out at https://github.com/ros-industrial/industrial_ci/pull/113#discussion_r97195126

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -45,7 +45,7 @@ With a few steps, you can start in your client repository using CI confiurations
   script:
     - .ci_config/travis.sh
 
-* Note that `.ci_config` is the REQUIRED name of the cloned folder; it is hardcoded on `industrial_ci` side so you must adhere to it.
+* Note: The name `.ci_config` is NO longer REQUIRED for the cloned folder starting version 0.3.2; you can pick any name as long as the folder is hidden (by being prepended by ".").
 
 Concrete examples of config files
 -------------------------------------


ipa-mdl:
> you can name it whatever you like.

-> "you can name it whatever you like, a hidden folder is recommended"

README.rst
```diff
@@ -45,7 +45,7 @@ With a few steps, you can start in your client repository using CI confiurations
   script:
     - .ci_config/travis.sh
 
-* Note that `.ci_config` is the REQUIRED name of the cloned folder; it is hardcoded on `industrial_ci` side so you must adhere to it.
+* Note: The name `.ci_config` is NO longer REQUIRED for the cloned folder since version 0.3.2; you can name it whatever you like.
```
130s: > -> "you can name it whatever you like, a hidden folder is recommended"

@ipa-mdl Actually, is there a reason for this system-wise?
I've chosen to use hidden folder as a cleaner solution, even though CI scripts only run temporarily during the CI jobs. But other than that I don't see a reason.

ipa-mdl: catkin does not follow hidden folders, so it is basically a `CATKIN_IGNORE`.
(I don't want to `CATKIN_IGNORE` `industrial_ci` as it is a valid catkin package, which should be released later on)
If the CI folder is not hidden, it gets included in the build (and the logs), which is not that bad.
As an alternative the CI folder might be placed outside of `TARGET_REPO_PATH`, e.g. in `/tmp/' oder '~/'.

130s: I see. Then why not making the folder hidden is a requirement because I don't see any benefit in client repositories to build industrial_ci? I've updated the commit so.

PR #116, opened 2017-01-23
Run all tests in docker

ipa-mdl:
This PR sits on top of #114 

It runs all source tests in docker, where as the docker container/image can be replaced.
For example `mluedtke/ci-test:indigo` is an uploaded version of the default image for indigo.
This enables local tests as well:

```
cd my_repository
rosrun industrial_ci run_ci ROS_DISTRO=kinetic OS_REPOSITORY_PATH=http://packages.ros.org/ros/ubuntu
``` 

or 

```
rosrun industrial_ci run_ci /path/to/my/repository ROS_DISTRO=kinetic OS_REPOSITORY_PATH=http://packages.ros.org/ros/ubuntu
```

The drawback is that it takes a little bit longer to build the default image, but we could speed-up the process by provided an automated build.
Not yet ready for merge.

PR #114, opened 2017-01-21
prerelease-test cleanup and module preparations

ipa-mdl:
This is the last stop before splitting the `source_tests.sh` into functions (#98).

All scripts are now contained in `src` folder. The actual test scripts are contained in `src/tests`.
I have removed the install directive from the package for the time being.

The prerelease tests script does not expose variables anymore, installs less dependencies and calls the individual test scripts in folds.

PR #117, opened 2017-01-24
[hotfix] install recommended ssh-client

ipa-mdl:
#114 removed ssh-client from the installed package as it is just a recommended package.
As a result #92 is not working anymore.

This is also addressed in #116


ipa-mdl:
tests pass -> merge

PR #124, opened 2017-02-14
make prerelease test fail on test failures

ipa-mdl:
This was the default before c1594f46900f40a00697b5669acb5d062ffbdaf0.
However, the generated scripts do not stop on errors: https://github.com/ros-infrastructure/ros_buildfarm/issues/367

### DIFF
--- a/industrial_ci/src/tests/ros_prerelease.sh
+++ b/industrial_ci/src/tests/ros_prerelease.sh
@@ -76,7 +76,7 @@ function run_ros_prerelease() {
     run_in_prerelease_docker generate_prerelease_script.py https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/production/index.yaml "$ROS_DISTRO" default ubuntu "$UBUNTU_OS_CODE_NAME" amd64 "${reponame}" --level "$downstream_depth" --output-dir .
     ici_time_end  # setup_prerelease_scripts
 
-    if [ "$clone_underlay" = true ]; then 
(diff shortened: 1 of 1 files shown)


130s:
So, shouldn't a job on this repo that uses prerelease test like this one https://travis-ci.org/ros-industrial/industrial_ci/jobs/201356133 with `PRERELEASE_REPO=i_donot_exist` fail? It's passing currently.

ipa-mdl:
The prerelease test fails for all errors except failed tests :-/

> So, shouldn't a job on this repo that uses prerelease test like this one https://travis-ci.org/ros-industrial/industrial_ci/jobs/201356133 with PRERELEASE_REPO=i_donot_exist fail? It's passing currently.

That's correct behavior, the tests fails with 1 (fold is red), but `EXPECTED_EXIT_CODE` is 1 as well, so the job is passing :)

130s:
> That's correct behavior, the tests fails with 1 (fold is red), but EXPECTED_EXIT_CODE is 1

Oh, I see! Nice.

So is this change temporary? When the behavior discussed in https://github.com/ros-infrastructure/ros_buildfarm/issues/367 improved so that prerelease test fails upon test failure, should we roll back the change in this PR?

130s:
Regardless the response to my question https://github.com/ros-industrial/industrial_ci/pull/124#issuecomment-279760056, we can merge this. +1

ipa-mdl:
> improved so that prerelease test fails upon test failure, should we roll back the change in this PR?

Yes, I think so.

PR #128, opened 2017-02-16
set TERM, run tests with tty

ipa-mdl:
pass `TERM` to docker, to get rid of the `NOTICE: Could not determine the width of the terminal. A default width of 80 will be used. This warning will only be printed once.` warnings.


ipa-mdl:
canceled the other tests

130s:
The warning can still be seen e.g. [this line](https://travis-ci.org/ros-industrial/industrial_ci/jobs/202266336#L905)?

ipa-mdl:
> The warning can still be seen e.g. this line?

Yes :-/
https://github.com/catkin/catkin_tools/blob/master/catkin_tools/common.py#L444 seems not to work, it requires a tty.

ipa-mdl:
This did the trick.
So `catkin` does not read `TERM` anymore, but we should keep it for other tools.

PR #131, opened 2017-02-20
support for ROS lunar

ipa-mdl:
too easy ;)


130s:
The [`lunar` job](https://travis-ci.org/ros-industrial/industrial_ci/jobs/203429152) passing already doesn't make sense to me, when there are only a couple of packages? And they are available only on shadow-fixed but the [job config in .travis.yml](https://github.com/ros-industrial/industrial_ci/pull/131/files#diff-354f30a63fb0907d4ad57269548329e3R45) doesn't specify shadow. At this moment don't we want it to fail?

ipa-mdl:
>  but the job config in .travis.yml doesn't specify shadow. 
For some reaseon, shadow is the [default](https://github.com/ros-industrial/industrial_ci/blob/master/industrial_ci/src/env.sh#L24)

130s:
@ipa-mdl 

>>  but the job config in .travis.yml doesn't specify shadow. 
>
> For some reason, shadow is the [default](https://github.com/ros-industrial/industrial_ci/blob/master/industrial_ci/src/env.sh#L24)

Hm, I missed that, and that's something I've always been wondering what's the best. I want to discuss that in #132 

Also please consider https://github.com/ipa-mdl/industrial_ci/pull/6

ipa-mdl:
I have udated the config to select shadow eplicitly.

PR #132, opened 2017-02-23
[doc] Clarify the semi-deprecated ROS_REPOSITORY_PATH env var usage

130s:
Found at https://github.com/ros-industrial/industrial_ci/pull/131#issuecomment-281181103

I don't know if this is what we wanted/want. Thoughts?

If we decide to merge this change then we might want to inform users because this changes the default behavior.


ipa-mdl:
I would not merge it by now, but I think shadow should not be the default.
I really like the `moveit_ci` `ROS_REPO` variable, it would be great to have some shortcuts like `shadow`, `shadow-fixed`, `stable`, `testing`, `building` etc.

`ROS_REPO` and `ROS_REPOSITORY_PATH` won't be used with custom Docker images anyway  (as the repo setup is part of the image).

We could implement the follwing logic:
If `ROS_REPOSITORY_PATH` is set, use it.
If not, try to derive it from `ROS_REPO` (which defaults to `stable`).

130s:
> I really like the `moveit_ci` `ROS_REPO` variable, it would be great to have some shortcuts like `shadow`, `shadow-fixed`, `stable`, `testing`, `building` etc.

You mean [these moveit_ci](https://github.com/ros-planning/moveit_ci/blob/65aaef7c443f47bf90322d3052db08b70bc3203d/.travis.yml#L16_L18). Indeed they look nice.

> If `ROS_REPOSITORY_PATH` is set, use it.
> If not, try to derive it from `ROS_REPO` (which defaults to `stable`).

+1

ipa-mdl:
Should we reissue this PR?

ipa-mdl:
Why did you replace ROS_REPOSITORY_PATH in the second commit?
This will fail!

ipa-mdl:
The new deafult can be set [there](https://github.com/ros-industrial/industrial_ci/blob/master/industrial_ci/src/env.sh#L36-L38)

130s:
Sorry, I was confused.

Added a bit more of the relationship of ROS_REPOSITORY_PATH and ROS_REPO to the document.

130s:
Conflict resolved.

130s:
> Please don't set ROS_REPOSITORY_PATH as this will disable the ROS_REPO feature.

I see, done (by simply resetting the commit that did wrong).
Title of the PR also updated now that this only changes a doc.

ipa-mdl:
Should I open another PR for the new default repo?

PR #133, opened 2017-02-24
[ROS Lunar] Use mockups pkg as a smoke test.

130s:
<s>**DO NOT MERGE YET** - this should be mergeable once core packages of ROS Lunar become available (so maybe when https://github.com/ros/rosdistro/issues/13550 gets closed).</s> -> Condition cleared.

This addresses the concern raised at https://github.com/ros-industrial/industrial_ci/pull/131#issuecomment-281162743

> The [`lunar` job](https://travis-ci.org/ros-industrial/industrial_ci/jobs/203429152) passing already doesn't make sense to me, when there are only a couple of packages (I forgot what they were back then. Today more core packages have been already [released](http://repositories.ros.org/status_page/ros_lunar_default.html))?

Looking at the commit history at [rosdistro/lunar/distribution.yaml](https://github.com/ros/rosdistro/commits/master/lunar/distribution.yaml) the two packages that were available for Lunar at the time of https://github.com/ros-industrial/industrial_ci/pull/131#issuecomment-281162743 were `catkin` and `gencpp`, which are not sufficient for any catkin package to pass the test IMO. So any Lunar test should've failed, but due to the lack of testing, it passed.

Using `industrial_ci_testpkg` for the test should address the situation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ env:
     - ROS_DISTRO=kinetic OS_NAME=debian OS_CODE_NAME=xenial EXPECT_EXIT_CODE=1
     - ROS_DISTRO=kinetic NOT_TEST_BUILD='true' DEBUG_BASH='true' VERBOSE_OUTPUT='false'
     - ROS_DISTRO=kinetic APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet  # Passing wrong SKS URL as a break test. Should still pass.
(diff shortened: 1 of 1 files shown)


ipa-mdl:
This test doesn't cover any additional code path, it just adds up another >2.5 minutes to the test time.
If you really want to the test package build for lunar, I would suggest to merge it with the other line.
Perhaps we should do the same for the hydro test.

PR #135, opened 2017-03-05
introduced ROS_REPO environment variable

ipa-mdl:
as discussed in #132 
alias names are based on the server paths.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ env:
     - ROS_DISTRO=jade   PRERELEASE=true PRERELEASE_REPONAME=i_do_not_exist  EXPECT_EXIT_CODE=1 PRERELEASE_DOWNSTREAM_DEPTH=1  # Intended to fail, to test the capability of capturing Prerelease Test failure.
     - ROS_DISTRO=kinetic AFTER_SCRIPT='catkin --version'
     - ROS_DISTRO=kinetic NOT_TEST_BUILD='true' DEBUG_BASH='true' VERBOSE_OUTPUT='false'
-    - ROS_DISTRO=lunar ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
+    - ROS_DISTRO=lunar ROS_REPO=ros-shadow-fixed
 matrix:
   allow_failures:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -42,7 +42,7 @@ env:
     - ROS_DISTRO=jade   PRERELEASE=true PRERELEASE_REPONAME=i_do_not_exist  EXPECT_EXIT_CODE=1 PRERELEASE_DOWNSTREAM_DEPTH=1  # Intended to fail, to test the capability of capturing Prerelease Test failure.
     - ROS_DISTRO=kinetic AFTER_SCRIPT='catkin --version'
     - ROS_DISTRO=kinetic NOT_TEST_BUILD='true' DEBUG_BASH='true' VERBOSE_OUTPUT='false'
-    - ROS_DISTRO=lunar ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
+    - ROS_DISTRO=lunar ROS_REPO=ros-shadow-fixed
```
130s: Is there any issue in even shortening the value as "shadow-fixed" (without `ros-`)?

ipa-mdl: my policy was to use the repository path names for now, but we can add more :)

industrial_ci/src/env.sh
```diff
@@ -35,6 +33,24 @@ if [ "$USE_DEB" ]; then  # USE_DEB is deprecated. See https://github.com/ros-ind
 fi
 if [ ! "$UPSTREAM_WORKSPACE" ]; then export UPSTREAM_WORKSPACE="debian"; fi
 
+# If not specified, use ROS Shadow repository http://wiki.ros.org/ShadowRepository
```
ipa-mdl: @130s: Do you want to add more aliases (apart form shadow-fixed)?

.travis.yml
130s: Fair enough.

industrial_ci/src/env.sh
130s: Yeah, adding `ros` will be great.

130s: Oope, `ros` is already an option. Then I don't know if there's any other binary repository worth being added here.
So +1 with as is.

PR #136, opened 2017-03-06
use absolute paths

ipa-mdl:
This is taken from #120.
I'd like to merge it separately as it is required for other development as well.

This commit prevents any directory changes within our scripts.

* does not assume the location to match a specific directoy
* prefix all commands (if needed)
* prefix all path

### DIFF
--- a/.travis.rosinstall
+++ b/.travis.rosinstall
@@ -1,3 +1,6 @@
 - git:
     uri: https://github.com/ros/std_msgs.git
     local-name: ros/std_msgs
+- git:
+    uri: https://github.com/ros-industrial/industrial_ci.git
+    local-name: industrial_ci
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -88,8 +88,8 @@ esac
 
 # download upstream packages into workspace
 if [ -e $CATKIN_WORKSPACE/src/.rosinstall ]; then
-    # ensure that the downstream is not in .rosinstall
-    (cd $CATKIN_WORKSPACE/src; $ROSWS rm $TARGET_REPO_NAME || true)
+    # ensure that the target is not in .rosinstall
+    (cd $CATKIN_WORKSPACE/src; $ROSWS rm $TARGET_REPO_NAME 2> /dev/null || true)
     $ROSWS update -t $CATKIN_WORKSPACE/src
 fi
 # TARGET_REPO_PATH is the path of the downstream repository that we are testing. Link it to the catkin workspace
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -89,7 +89,8 @@ esac
 # download upstream packages into workspace
 if [ -e $CATKIN_WORKSPACE/src/.rosinstall ]; then
     # ensure that the target is not in .rosinstall
-    (cd $CATKIN_WORKSPACE/src; $ROSWS rm $TARGET_REPO_NAME 2> /dev/null || true)
+    (cd $CATKIN_WORKSPACE/src; $ROSWS rm $TARGET_REPO_NAME 2> /dev/null \
+     && echo "$ROSWS ignored $TARGET_REPO_NAME found in $CATKIN_WORKSPACE/src/.rosinstall file. Its source fetched from your repository is used instead." || true) # TODO: add warn function
     $ROSWS update -t $CATKIN_WORKSPACE/src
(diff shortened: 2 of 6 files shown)


ipa-mdl:
I have fixed two bugs related to wstool updates, should work as expected now.

130s:
I assume the title can be changed as something like "fix wstool commands"?

It's kind of a nitpick but what is this error and where does it occur? 
https://travis-ci.org/ros-industrial/industrial_ci/jobs/208310848#L170
```
ERROR in config: Unknown elements '['industrial_ci']'
```
Though this is probably not affecting the behavior, error messages could mislead readers so we might want to remove the source of the error if this is irrelevant to any of the actual errors.

Other than that LGTM.

ipa-mdl:
This "ERROR" is the result of the convenience feature that filters the target repo out of the upstream workspace for shared rosinstall files.
This fails if the target repo is not present in the rosinstall file.
The exit code was already ignored. I have silenced the output as well.

130s:
> This "ERROR" is the result of the convenience feature that filters the target repo out of the upstream workspace for shared rosinstall files.
> This fails if the target repo is not present in the rosinstall file.
> The exit code was already ignored. I have silenced the output as well.

I see. That's a nice handling. So in the case of us https://github.com/ros-industrial/industrial_ci/pull/136#issuecomment-284717798, because [.travis.rosinstall](https://github.com/ros-industrial/industrial_ci/blob/638f4d767ed2ed140f9206d160e863a429a90d3b/.travis.rosinstall) didn't include `industrial_ci` element `wstool` failed.

Then, do we not want to print if there was an element removed? Otherwise users won't notice.
I can't think of any usecase where users want to use the source of the target repo that is downloaded via rosinstall, but there ***can*** be the case where the specified rosinstall file includes the target repo. So I would print something like:

```
$ROSWS ignored $TARGET_REPO_NAME found in $CATKIN_WORKSPACE/src/.rosinstall file. Its source fetched from your repository is used instead.
```

ipa-mdl:
locale issue is tracked in #141

ipa-mdl:
TODO: `AFTER_SCRIPT` and `BEFORE_SCRIPT` should be executed in (read-only) `TARGET_REPO_PATH` context to simplify using external scripts

ipa-mdl:
@130s: please review again

industrial_ci/src/tests/source_tests.sh
```diff
@@ -89,7 +89,8 @@ esac
 # download upstream packages into workspace
 if [ -e $CATKIN_WORKSPACE/src/.rosinstall ]; then
     # ensure that the target is not in .rosinstall
-    (cd $CATKIN_WORKSPACE/src; $ROSWS rm $TARGET_REPO_NAME 2> /dev/null || true)
+    (cd $CATKIN_WORKSPACE/src; $ROSWS rm $TARGET_REPO_NAME 2> /dev/null \
+     && echo "$ROSWS ignored $TARGET_REPO_NAME found in $CATKIN_WORKSPACE/src/.rosinstall file. Its source fetched from your repository is used instead." || true) # TODO: add warn function
```
130s: > TODO: add warn function

Is this now addressed with this commit, or is it something else?

.travis.rosinstall
```diff
@@ -1,3 +1,6 @@
 - git:
     uri: https://github.com/ros/std_msgs.git
     local-name: ros/std_msgs
+- git:
+    uri: https://github.com/ros-industrial/industrial_ci.git
```
130s: Hm, I expect this entry will be skipped with a warning message, but I can't find it. This should be [this job](https://travis-ci.org/ros-industrial/industrial_ci/jobs/209393510) isn't it?

industrial_ci/src/tests/source_tests.sh
ipa-mdl: This is just a hint for us to wrap it it with a to-be-implement (log_)warn function later.

.travis.rosinstall
ipa-mdl: This job is using the file from the master branch.

It gets skipped in 
https://travis-ci.org/ros-industrial/industrial_ci/jobs/209393509
https://travis-ci.org/ros-industrial/industrial_ci/jobs/209393512
https://travis-ci.org/ros-industrial/industrial_ci/jobs/209393523

130s: > This job is using the file from the master branch.

Ah, you're absolutely right. We can't test unless you change the URL of the file.
Ok, we can see how this goes after merge.

Issue #106, opened 2017-01-17
Upstream/downstream logic design


ipa-mdl:
(branched off from https://github.com/ros-industrial/industrial_ci/pull/104#issuecomment-273306222)

The current logic needs to be improved, as it might not work as expected.

We have
* target repo with contained packages.
* a list of target to be tested (`TARGET_PKGS`)
* a rosinstall workspace (`UPSTREAM_WORKSPACE`)
* a list of downstream packages (`DOWNSTREAM_PKGS`)

We want to:
* install all dependencies that we do not have sources for (target or upstream or downstream)
* build all upstream dependencies (splitted from targets to identify problems with them)
* build all `TARGET_PKGS` (without dependencies) 
* run tests for `TARGET_PKGS`
* run installed tests for `DOWNSTREAM_PKGS` 
* build all `DOWNSTREAM_PKGS` (without dependencies)
* run tests for `DOWNSTREAM_PKGS`
* run all installed tests for `DOWNSTREAM_PKGS` 
* run installed tests for `DOWNSTREAM_PKGS` 

`DOWNSTREAM_PKGS` might be derived from `TARGET_PKGS`.
The list of upstream source depends can be easily derived from `TARGET_PKGS` and `DOWNSTREAM_PKGS`.

 We should discuss if we need to add a `DOWNSTREAM_WORKSPACE`.
I think this will include more problems than it solves as we have to deal with doubled packages.
Instead we might introduce a `ROSINSTALL_WORKSPACE` option as an alias for `UPSTREAM_WORKSPACE`.

In addition we have these issues:
- [ ] currently all dependencies get installed
- [ ] all packages from the rosinstall file get built per default
Both might take a lot of time.

ipa-mdl:
I have tried to summarize the types of packages:
![image](https://cloud.githubusercontent.com/assets/727227/23133186/88dfed52-f790-11e6-8095-043d10427aa1.png)

ipa-mdl:
Actually, the different groups may overlap:
![deps](https://cloud.githubusercontent.com/assets/727227/23676081/ca64c788-037b-11e7-80e6-ec21766ab5f9.png)

130s:
Sorry for taking forever to start looking into this. Just clarification for now,

> We want to:
> :
> - run installed tests for DOWNSTREAM_PKGS
> :
> - run tests for DOWNSTREAM_PKGS
> - run all installed tests for DOWNSTREAM_PKGS
[...]

PR #137, opened 2017-03-06
[WiP] Fix upstream downstream package logic

### LINKED ISSUE #118: test_results block never reaches upon failure in catkin_run_tests
During `catkin_run_tests` block, the [script terminates if tests return error](https://github.com/ros-industrial/industrial_ci/blob/c1594f46900f40a00697b5669acb5d062ffbdaf0/industrial_ci/src/tests/source_tests.sh#L137), so that `test_results` block, which comes later that runs `catkin_test_results --verbose` to show the error details, won't happen. Obviously this makes debugging hard.

Example project: [ar_track_alvar](https://github.com/sniekum/ar_track_alvar/blob/562fae16ec695a913d447e72951b6c65acd2b6b1/.travis.yml#L15), [result on Travis](https://travis-ci.org/sniekum/ar_track_alvar/jobs/199885756#L3336)

ipa-mdl:
addresses #106 

some unrelated minor fixes are bundled as well:
* ~~build to install space (#54)~~ (#150)
* based on improved #136 
* fold naming
* test result handling

The dependency helper covers most of the [designed logic](https://github.com/ros-industrial/industrial_ci/issues/106#issuecomment-281121013), but uses a consolidated rosdep install step for now.

It detects the target and downsteam packages automatically, if these are not provided.
The detection can be filtered by  `BUILD_PKGS_WHITELIST` and `BUILD_PKGS_BLACKLIST` (new).
The upstream packages cannot be black- or whitelisted.

Based on all upstream, target and downstream packages the list of rosdep src paths gets compiled.
In addtion the list of tests gets generated and can be filtered with `TEST_BLACKLIST` (new).

The resulting package list gets printed outside of the folds, including the ignores packages.
As unrelated packages get irgnored now, the travis jobs had to be extended to enforce the downstream packages.

All steps are run in separated folds, so failures can be attributed more easily.
I took the chance to fix #118 

(The list of commits is a little bit mixed up - sorted by date instead of relation - in the PR display because of the rebases)

TODO: 
- [ ] documentation
- [ ] setting DOWNSTREAM _PKGS should prevent the auto-detection
- [x] run rostest with env loader (removed source setup.bash by accident)
- [ ] make `deps.py`  a proper module
- [ ] add unit tests for `deps.py`


ipa-mdl:
For some reasons the actionlib tests fail.
This raised the question: Should downstream tests be exercised? Do we want an opt-out flag?

ipa-fxm:
Tested this together with @ipa-mdl on branch https://github.com/ipa320/cob_robots/pull/626
Showed that transitive dependencies are not handled correctly as they eventually end up in the `Ignored packages` set

ipa-mdl:
I have  rebased to current master and updated the logic.

`catkin`'s `topological_order` does not respect exec dependencies in the ordering.So some dependencies were skpipped. This okay for the sake of the builds and tests, but rosdep installs all dependencies and got into trouble.

ipa-mdl:
The latest update improved the handling of export and test depends.
exec_depends are now treated as transitive test_depends.
In addtion uses exec_depends  get skipped in rosdep install.

@ipa-fxm: please test again. On my laptop the full run of the kinetic build/test takes 20 minutes.

ipa-fxm:
See my https://github.com/ipa320/cob_robots/pull/626#issuecomment-284958637...

ipa-mdl:
The same technique could be used for prerelease tests (#145)

ipa-mdl:
rebased on top of #150

ipa-mdl:
rebased to current master, works with `ROSDEP_SKIP_KEYS`

ipa-mdl:
[relaxed](https://github.com/ros-industrial/industrial_ci/issues/106#issuecomment-313348034) mode is default, strict mode can be enabled with  `STRICT_TEST_DEPENDS=true`.

ipa-mdl:
as discussed with @ipa-fxm downstream packages should not get detected automatically

ipa-fxm:
having fixed our issue with the travis log length in #267, I would also like to see our travis build time reduced

what is the status of this feature/pr?
I'm willing to help getting this ready to merge

@ipa-mdl 
can you summarize what still needs to be done or what is currently preventing this from being merged (despite needing to rebase)?

130s:
I'm also interested in this feature.
Do you have any plan to come back to this @ipa-mdl?

ipa-mdl:
> Do you have any plan to come back to this 

Not in this form. I think we can migrate some of it to our current codebase. For the actual upstream/downstream detection I would rather use some interfaces of catkin-tools directly instead of my custom code.

ipa-mdl:
Proper handling of upstream and downstream workspaces is implemented in #361

PR #140, opened 2017-03-08
handle missing UPSTREAM_WORKSPACE file

### LINKED ISSUE #139: Missing rosinstall file is not handled
If `UPSTREAM_WORKSPACE=file` and  `$ROSINSTALL_FILENAME[.$ROS_DISTRO]` does not exist, the build should fail.
@ipa-fxm: FYI

ipa-mdl:
fixes #139

### DIFF
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -78,6 +78,8 @@ file) # When UPSTREAM_WORKSPACE is file, the dependended packages that need to b
     elif [ -e $TARGET_REPO_PATH/$ROSINSTALL_FILENAME ]; then
         # install (maybe unreleased version) dependencies from source
         $ROSWS merge file://$TARGET_REPO_PATH/$ROSINSTALL_FILENAME
+    else
+        error "UPSTREAM_WORKSPACE file '$TARGET_REPO_PATH/$ROSINSTALL_FILENAME[.$ROS_DISTRO]' does not exist"
     fi
     ;;
 http://* | https://*) # When UPSTREAM_WORKSPACE is an http url, use it directly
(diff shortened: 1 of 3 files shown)


industrial_ci/src/tests/source_tests.sh
```diff
@@ -78,6 +78,8 @@ file) # When UPSTREAM_WORKSPACE is file, the dependended packages that need to b
     elif [ -e $TARGET_REPO_PATH/$ROSINSTALL_FILENAME ]; then
         # install (maybe unreleased version) dependencies from source
         $ROSWS merge file://$TARGET_REPO_PATH/$ROSINSTALL_FILENAME
+    else
+        error "UPSTREAM_WORKSPACE file '$TARGET_REPO_PATH/$ROSINSTALL_FILENAME[.$ROS_DISTRO]' does not exist"
```
130s: Output of this line looks a bit confusing:

https://travis-ci.org/ros-industrial/industrial_ci/jobs/208837309#L163
```
\e[31mUPSTREAM_WORKSPACE file '/root//home/travis/build/ros-industrial/industrial_ci/.i.do.not.exist[.jade]' does not exist\e0m
```

- There may be a bug in `error` function. The extraneous prefix in the beginning (`\e[31mUPSTREAM_WORKSPACE`).
- What's the purpose for enclosing the file extension as `[.$ROS_DISTRO]`?

ipa-mdl: > What's the purpose for enclosing the file extension as [.$ROS_DISTRO]?

At this point we can't tell which file is missing, the generic or the distro-specific one.

ipa-mdl: > There may be a bug in error function.

make it two ;)

ipa-mdl:
The `TARGET_REPO_PATH` is fixed #136

industrial_ci/src/tests/source_tests.sh
130s: >>    What's the purpose for enclosing the file extension as [.$ROS_DISTRO]?
>
> At this point we can't tell which file is missing, the generic or the distro-specific one.

I think I got your point; either generic or distro-specific file is missing but there's no way for the script to know which one is intended by the user.

The `[]` notation may not look the best appropriate for this case, but then I don't know what's the best. Let's just leave this as is :)

PR #143, opened 2017-03-10
[doc] Clarify .rosinstall file usage.

130s:
I'm not entirely sure if stating the default value of `ROSINSTALL_FILENAME` as `.travis.rosinstall` -- this could give a wrong impression that this env var is always set (Fact is that it's only set/used when `UPSTREAM_WORKSPACE` is set as already explained.

But if one defines `UPSTREAM_WORKSPACE` but forgets `ROSINSTALL_FILENAME`, `.travis.rosinstall` will be sought for and if no file with that name found and the job fails. So I think it'd be still better to state default value.


130s:
I'm not entirely sure if stating the default value of `ROSINSTALL_FILENAME` as `.travis.rosinstall` -- this could give a wrong impression that this env var is always set (Fact is that it's only set/used when `UPSTREAM_WORKSPACE` is set as already explained.

But if one defines `UPSTREAM_WORKSPACE` but forgets `ROSINSTALL_FILENAME`, `.travis.rosinstall` will be sought for and if no file with that name found and the job fails. So I think it'd be still better to state default value.

doc/index.rst
```diff
@@ -332,10 +332,16 @@ Use .rosinstall file to specify the depended packages source repository
 
 WARNING: In all cases where you want to utilize `.rosinstall` (or similar name) files, be sure to set `USE_DEB` as `false`, or simply not define it.
 
-For using a rosinstall file located locally within the repository:
+For using a rosinstall file located locally within the repository, define both two variables as:
```
ipa-mdl: You don't have to define both if you stick to the default of `ROSINSTALL_FILENAME`

ipa-mdl:
Apart from the one comment, the update is more helpful.

In the midterm I'd like to implement the [moveit_ci solution](https://github.com/ros-planning/moveit_ci/blob/master/travis.sh#L92) with the additional file branch and the `.$ROS_DISTRO` feature.

130s:
Thanks for the review @ipa-mdl. Updated.

PR #144, opened 2017-03-16
Changed urandom output for date hashing as it may cause problems with docker output.

inigomartinez:
When using industrial_ci with a CI service as Jenkins the output may be lost.

### DIFF
--- a/industrial_ci/src/util.sh
+++ b/industrial_ci/src/util.sh
@@ -38,7 +38,7 @@
 function ici_time_start {
     if [ "$DEBUG_BASH" ] && [ "$DEBUG_BASH" == true ]; then set +x; fi
     ICI_START_TIME=$(date +%s%N)
-    ICI_TIME_ID=$(cat /dev/urandom | tr -dc 'a-z0-9' | fold -w 8 | head -n 1)
+    ICI_TIME_ID=$(date | md5sum | fold -w 8 | head -n 1)
     ICI_FOLD_NAME=$1
     echo -e "\e[0Kici_fold:start:$ICI_FOLD_NAME"
     echo -e "\e[0Kici_time:start:$ICI_TIME_ID\e[34m>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>\e[0m"
--- a/industrial_ci/src/util.sh
+++ b/industrial_ci/src/util.sh
@@ -38,7 +38,7 @@
 function ici_time_start {
     if [ "$DEBUG_BASH" ] && [ "$DEBUG_BASH" == true ]; then set +x; fi
     ICI_START_TIME=$(date +%s%N)
-    ICI_TIME_ID=$(date +%s%N | md5sum | fold -w 8 | head -n 1)
+    ICI_TIME_ID=$(printf "%x\n" $ICI_START_TIME)
     ICI_FOLD_NAME=$1
     echo -e "\e[0Kici_fold:start:$ICI_FOLD_NAME"
     echo -e "\e[0Kici_time:start:$ICI_TIME_ID\e[34m>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>\e[0m"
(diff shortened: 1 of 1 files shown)


ipa-mdl:
> When using industrial_ci with a CI service as Jenkins the output may be lost.

How could it get lost?

`date` just has a second resolution, so this might clash.

ipa-mdl:
We could use `date +%s%N` (same as `$ICI_START_TIME`).

ipa-mdl:
So, this seems PR to remove the color artifacts :+1:

inigomartinez:
I don't know why but as far as I know "cat /dev/urandom | tr -dc 'a-z0-9'" hides Jenkins output. You are right, date isn't the best option as it may clash.

If the problem is due to the use of the cat command, the following commands may solve the problem:

1. $(hexdump -n 4 -v -e '/1 "%02X"' /dev/urandom)
2. $(uuidgen | fold -w 8) command from libuuid, but is not always available.
3. $(openssl rand -hex 8) comman from openssl, but as uudigen, is not always available.

ipa-mdl:
`date +%s%N` has a resolution in the nanosecond range, which should be fine

industrial_ci/src/util.sh
```diff
@@ -38,7 +38,7 @@
 function ici_time_start {
     if [ "$DEBUG_BASH" ] && [ "$DEBUG_BASH" == true ]; then set +x; fi
     ICI_START_TIME=$(date +%s%N)
-    ICI_TIME_ID=$(cat /dev/urandom | tr -dc 'a-z0-9' | fold -w 8 | head -n 1)
+    ICI_TIME_ID=$(date +%s%N | md5sum | fold -w 8 | head -n 1)
```
ipa-mdl: `ICI_TIME_ID=$ICI_START_TIME`?

inigomartinez: I noticed it just after submitting the update, but I noticed that editing PR has generated another commit and I was trying to drop it.

ICI_TIME_ID=$(printf "%x\n" $ICI_START_TIME)

It does look fine for me.

ipa-mdl:
I have canceled the other build

inigomartinez:
Ok, thank you. Shouldn't be better if I cancel this PR and create a new fresh one with just one commit with the proper solution?

ipa-mdl:
>  Shouldn't be better if I cancel this PR and create a new fresh one with just one commit with the proper solution?

We can squash-merge the commits, so no need to reissue it.
But you can squash it yourself if you like to.

ipa-mdl:
@130s: Any objections?

130s:
+1

Just out of curiousity I measured the performance. The proposed command (command2 below) ***may*** run 'significantly' faster (not exactly sure. This is not a complete benchmark).

```
$ command1=$(cat /dev/urandom | tr -dc 'a-z0-9' | fold -w 8 | head -n 1)
$ response1="$(time ( "$command1" ) 2>&1 1>/dev/null )"
$ echo "$response1"
yothds18: command not found

real    0m0.405s
user    0m0.124s
sys     0m0.076s

$ ICI_START_TIME=$(date +%s%N)
$ command2=$(printf "%x" $ICI_START_TIME)
$ response2="$(time ( "$command2" ) 2>&1 1>/dev/null )"
$ echo "$response2"
14ac78f6025f7646: command not found

real    0m0.174s
user    0m0.144s
sys     0m0.040s
```

inigomartinez:
I may have an answer for that improvement. Instead of 4 piped commands, there is just only one, so there is no overhead regarding the data moving through commands. 

By the way, I was thinking about using an echo instead of printf, but the latter one seems to be [better](http://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo).

ipa-mdl:
@130s:
I don't think that your test measured the right thing.
You have basically measured how long the system takes to crawl all path for the command.

My results:
```
$ time (cat /dev/urandom | tr -dc 'a-z0-9' | fold -w 8 | head -n 1)
zm2835ar

real    0m0.007s
user    0m0.004s
sys     0m0.000s

$ time (printf "%x" $ICI_START_TIME)
14ac7c74c9204255
real    0m0.000s
user    0m0.000s
sys     0m0.000s

$ time (do_what_I_want)
do_what_I_want: command not found

real    0m0.105s
user    0m0.084s
sys     0m0.016s
```

> By the way, I was thinking about using an echo instead of printf

The printf just print the time as hex number.
This isn't necessary, but it shortens the ID.

PR #145, opened 2017-03-16
Use latest prerelease features

ipa-mdl:
* works for repositories that are not (yet) listed in rosdistro
* adds mockup feature
* slightly changes meaning of `PRERELEASE_REPONAME`
* add travis folding

The latter prevents testing packages other than the target repo as we did it before.


ipa-mdl:
New test case: https://travis-ci.org/ros-industrial/industrial_ci/jobs/211869139

ipa-mdl:
43657b2 can be removed as soon as https://github.com/ros-infrastructure/ros_buildfarm/issues/405 was merged.
09f95e43ba48cc969b63350af9f3eab258c19e32 should get removed or reverted after the release (https://github.com/ros-infrastructure/ros_buildfarm/issues/394)

130s:
Will try looking into this over the weekend!

doc/index.rst
```diff
@@ -166,9 +166,7 @@ Note that some of these currently tied only to a single option, but we still lea
...
-* `PRERELEASE_REPONAME` (default: not set): The target of Prerelease Test (that you select at `<http://prerelease.ros.org/indigo>`_, `<http://prerelease.ros.org/kinetic>`_ etc.).
-  * If not set then it tests the package of the repository's name. You can specify this by your ROS package name format (with underscore e.g. `industrial_core`), not Debian package name format. NOTE that this package name must be listed in the `rosdistro/distribution.yaml` (e.g. [for ROS Indigo](https://github.com/ros/rosdistro/blob/master/indigo/distribution.yaml)) (this requirement comes from ROS buildfarm's Prerelease Test).
-  * (As of Dec 2016) when this variable is set, development branch listed in `rosdistro/distribution.yaml` is tested. See [detail](https://github.com/ros-industrial/industrial_ci/pull/85#issue-196409011).
+* `PRERELEASE_REPONAME` (default: TARGET_REPO_NAME): The  name of the target of Prerelease Test in rosdistro (that you select at `<http://prerelease.ros.org>`_). You can specify this if your repository name differs from the corresponding rosdisto entry.
```
130s: > ```
> You can specify this if your repository name differs from the corresponding rosdisto entry.
> ```

This is a bit unclear. Readers may want to know, in which cases specifying different name here could be useful, if the name specified is different from the tested repo name, where is the specified package retrieved from? etc.

industrial_ci/src/tests/ros_prerelease.sh
```diff
@@ -72,41 +73,26 @@ function run_ros_prerelease() {
...
-    if [ "$downstream_depth" != "0" ]; then
-        ici_time_start run_prerelease_clone_overlay.sh
-        run_in_prerelease_docker ./prerelease_clone_overlay.sh
-        ici_time_end  # run_prerelease_clone_overlay
-        ici_time_start prerelease_build_overlay.sh
-        run_in_prerelease_docker ./prerelease_build_overlay.sh
-        run_in_prerelease_docker catkin_test_results $WORKSPACE/catkin_workspace_overlay/test_results
-        ici_time_end  # prerelease_build_overlay.sh
-    fi
+    run_in_prerelease_docker generate_prerelease_script.py https://raw.githubusercontent.com/ros-infrastructure/ros_buildfarm_config/production/index.yaml "$ROS_DISTRO" default ubuntu "$UBUNTU_OS_CODE_NAME" amd64 --level "$downstream_depth" --output-dir . --custom-repo "$reponame::::"
```
130s: > `"$reponame::::"` 

Just curious, are the trailing colons still needed?

.travis.yml
```diff
@@ -26,8 +26,9 @@ env:
     - ROS_DISTRO=indigo NOT_TEST_BUILD='true' NOT_TEST_INSTALL='true' POST_PROCESS='I_am_supposed_to_fail'
     - ROS_DISTRO=indigo CATKIN_PARALLEL_JOBS='-p1' ROS_PARALLEL_JOBS='-j1'  # Intend build on low-power platform
     # - env: ROS_DISTRO=indigo PRERELEASE=true  ## Comment out because this is meaningless for always failing without prerelease testable contents in industrial_ci.
-    - ROS_DISTRO=indigo PRERELEASE=true PRERELEASE_REPONAME=std_msgs PRERELEASE_DOWNSTREAM_DEPTH=0
-    - ROS_DISTRO=kinetic PRERELEASE=true PRERELEASE_REPONAME=std_msgs
+    - ROS_DISTRO=indigo PRERELEASE=true PRERELEASE_DOWNSTREAM_DEPTH=1
```
130s: Don't we (also) want to add a job for a package that is actually depended on by maybe just one or a few packages so that we can test `PRERELEASE_DOWNSTREAM_DEPTH`? We can even add another package to mockups and inject dependency.

doc/index.rst
ipa-mdl: It does not get retrieved. I is just the name in rosdistro.
E.g. if you key in rosdistro is `my_pkg` and for some reason for source is in `https://github.com/me/my_pkg_impl.git`

industrial_ci/src/tests/ros_prerelease.sh
ipa-mdl: Yes, all of them.

repo **:** type **:** url_scheme **:** url_path **:** branch
For us type, url_scheme, url_patch, branch are empty.
This way the prerelease will fail if the target source code is missing.

.travis.yml
ipa-mdl: This is not possible for now. With this patch, we cannot "fake" the identity anymore, and `industrial_ci` does not have any dowstream depencies.
In a later versions we might add a full rosinstall workspace as underlay, then we can test it again

doc/index.rst
130s: I see. That usecase is a good example. We should add it to the doc or at least link here. Please review https://github.com/ipa-mdl/industrial_ci/pull/8

.travis.yml
130s: So I assume this new job does not test anything distinct for now other than testing `PRERELEASE_DOWNSTREAM_DEPTH` option to not cause error.

PR #150, opened 2017-03-20
build directly to install space

ipa-mdl:
This PR picks some install space related commits from #137
The current master (still) builds all packages, cleans and rebuild to install space (unless `NOT_TEST_INSTALL=true).

These patches avoid rebuilding and fix the retry behavior.

### DIFF
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -154,7 +154,7 @@ if [ "$NOT_TEST_INSTALL" != "true" ]; then
 
     ici_time_start catkin_install_run_tests
 
-    export EXIT_STATUS=0
+    EXIT_STATUS=0
     # Test if the unit tests in the packages in the downstream repo pass.
     if [ "$BUILDER" == catkin ]; then
       for pkg in $PKGS_DOWNSTREAM; do
@@ -163,8 +163,8 @@ if [ "$NOT_TEST_INSTALL" != "true" ]; then
         echo "[$pkg] Found $(echo $rostest_files | wc -w) tests."
         for test_file in $rostest_files; do
           echo "[$pkg] Testing $test_file"
-          $CATKIN_WORKSPACE/install/env.sh rostest $test_file || export EXIT_STATUS=$?
-          if [ $? != 0 ]; then
+          $CATKIN_WORKSPACE/install/env.sh rostest $test_file || EXIT_STATUS=$?
+          if [ $EXIT_STATUS != 0 ]; then
             echo -e "[$pkg] Testing again the failed test: $test_file.\e[31m>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>\e[0m"
             $CATKIN_WORKSPACE/install/env.sh rostest --text $test_file
             echo -e "[$pkg] Testing again the failed test: $test_file.\e[31m<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<\e[0m"
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -141,7 +141,6 @@ if [ "$NOT_TEST_BUILD" != "true" ]; then
     ici_time_start catkin_run_tests
 
     if [ "$BUILDER" == catkin ]; then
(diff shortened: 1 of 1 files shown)


ipa-mdl:
Since the mockup tests are not considered (fixed in #137), I have prepared a test brach in `ros_canopen`: https://travis-ci.org/ipa-mdl/ros_canopen/builds/213134116

In addtion there is a `cob_robots` branch: https://travis-ci.org/ipa320/cob_robots/builds/213121299
This repository is the main reason for fast-forwarding the install-space patches: Without them the travis build gets stopped because it exceeds 50 minutes.

industrial_ci/src/tests/source_tests.sh
```diff
@@ -139,43 +141,32 @@ if [ "$NOT_TEST_BUILD" != "true" ]; then
...
-
-    ici_time_end  # catkin_install_build
     ici_time_start catkin_install_run_tests
 
-    export EXIT_STATUS=0
+    EXIT_STATUS=0
     # Test if the unit tests in the packages in the downstream repo pass.
     if [ "$BUILDER" == catkin ]; then
       for pkg in $PKGS_DOWNSTREAM; do
+        if [ ! -d "$CATKIN_WORKSPACE/install/share/$pkg" ]; then continue; fi # skip meta-packages
```
130s: Is this comment correct? Isn't this line skipping those specified in `$PKGS_DOWNSTREAM` (which are already tested in `catkin_run_tests` section)?

130s:
Seems like this works great!

I'm not sure if I understand correctly, but looking at https://travis-ci.org/ipa320/cob_robots/jobs/213121300, seems like in `catkin_run_tests` section all tests are run, and `catkin_install_run_tests` finished within a second (not sure why that short. Are all tests done are from downstream?). I assume with the workspace configured for `install`, `catkin run_tests` does all what we wanted?

If so, is keeping those 2 sections still worthy?
Maybe for other build tools that don't handle `install` space (e.g. catkin_make) we can still keep the logic in catkin_install_run_tests.

industrial_ci/src/tests/source_tests.sh
ipa-mdl: No, it just skips packages without a directory in `install/share`. AFAIK this is only the case for meta-packages.

ipa-mdl:
I don't know why, but the `catkin_install_run_tests` section runs all *.test files from target and downstream packages found in the install space.
 
We don't install *.test for the cob packages by default, so this block is a no-op.

ipa-mdl:
We might consider skipping `catkin_install_run_tests` if we don't find any *.test files.

130s:
> We might consider skipping `catkin_install_run_tests` if we don't find any *.test files.

I think that's the way we might want to do as long as tests get to run in `catkin_run_tests`.

ipa-mdl:
>  > We might consider skipping catkin_install_run_tests if we don't find any *.test files.
> I think that's the way we might want to do as long as tests get to run in catkin_run_tests.

Let's create another issue/PR for this and not mix it.

industrial_ci/src/tests/source_tests.sh
```diff
@@ -103,6 +109,8 @@ if [ "${USE_MOCKUP// }" != "" ]; then
     ln -s "$TARGET_REPO_PATH/$USE_MOCKUP" $CATKIN_WORKSPACE/src
 fi
 
+catkin config --install
```
130s: https://github.com/ros-industrial/industrial_ci/pull/150#pullrequestreview-31649321
>>    The current behaviour is really odd:
>>
>>    1.    build to build/devel space
>>    1.    run build test
>>    1.    clean
>>    1.    build to build/devel space and install to install space
>>    1.    run rostest on install space
>>
>>    So in step 4 we are rebuilding everything from step 1, which doubles the time the build takes.
>
> Ok, with this PR we're skipping the build in 4, is that correct?
> What I was afraid was to skipping step 1 and 2 (and 3 obvisously), which could cause errors on the repos where no install rules are not defined. But looks like I was wrong and worrying about nothing.

I overlooked adding this line. With this `NOT_TEST_INSTALL=true` that allows users to avoid testing against `install` space becomes not working (e.g. https://github.com/ros-industrial-consortium/descartes/pull/198#issuecomment-301516448).

130s: ([My comment above](https://github.com/ros-industrial/industrial_ci/pull/150/files#r116532276) sounds a bit misleading, so I rephrase) With this line, users have no way to avoid testing against `install` space. E.g. https://github.com/ros-industrial-consortium/descartes/pull/198#issuecomment-301516448 where the tests depend on the files that are intentionally not deployed.

PR #152, opened 2017-03-20
Silence rosdep

### LINKED ISSUE #141: Set locale in docker
Perl complains about locales..
I will fix this together with another little silence patch for rosdep.

ipa-mdl:
Filters most of the rosdep install output, just prints commands and errors.
closes #141

### DIFF
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -117,7 +117,10 @@ fi
 
 ici_time_start rosdep_install
 
-rosdep install -q --from-paths $CATKIN_WORKSPACE --ignore-src --rosdistro $ROS_DISTRO -y
+set -o pipefail # fail if rosdep install fails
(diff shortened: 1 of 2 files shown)


industrial_ci/src/tests/source_tests.sh
```diff
@@ -117,7 +117,12 @@ fi
 
 ici_time_start rosdep_install
 
-rosdep install -q --from-paths $CATKIN_WORKSPACE --ignore-src --rosdistro $ROS_DISTRO -y
+function rosdep_install {
```
130s: What for is this a function? Are we going to reuse this somewhere?

ipa-mdl: Actually this is a little hack to make the job fail if rosdep fails.
As an alternative I could have surrounded it with
```
set -o pipefail
set +o pipefail
```

Hwoever, I plan to factor all code into functions anyway.

130s: I see. Could you then leave a short inline comment for that hack?

ipa-mdl: I have switched to pipefail and added a comment

PR #154, opened 2017-03-22
pass CC and CXX to docker

ipa-mdl:


ipa-mdl:
easy fix, tests pass.

Issue #146, opened 2017-03-17
Using rosinstall, "Multiple packages found with the same name X"


130s:
E.g. [this result on Travis](https://travis-ci.org/ccny-ros-pkg/scan_tools/jobs/209873931#L232). Its [PR](https://github.com/ccny-ros-pkg/scan_tools/pull/52), and the [.rosinstall](https://github.com/ccny-ros-pkg/scan_tools/blob/20783d94fb9c2799f7edf7f702eef960c67ad3c9/.rosinstall) file, which does contain the repo where the PR is opened at.

```
ERROR: Rosdep experienced an error: Multiple packages found with the same name "laser_ortho_projector":
- src/ccny-ros-pkg/scan_tools/laser_ortho_projector
- src/scan_tools/laser_ortho_projector
Multiple packages found with the same name "laser_scan_matcher":
- src/ccny-ros-pkg/scan_tools/laser_scan_matcher
- src/scan_tools/laser_scan_matcher
Multiple packages found with the same name "laser_scan_sparsifier":
- src/ccny-ros-pkg/scan_tools/laser_scan_sparsifier
- src/scan_tools/laser_scan_sparsifier
Multiple packages found with the same name "laser_scan_splitter":
- src/ccny-ros-pkg/scan_tools/laser_scan_splitter
- src/scan_tools/laser_scan_splitter
Multiple packages found with the same name "ncd_parser":
- src/ccny-ros-pkg/scan_tools/ncd_parser
- src/scan_tools/ncd_parser
Multiple packages found with the same name "polar_scan_matcher":
- src/ccny-ros-pkg/scan_tools/polar_scan_matcher
- src/scan_tools/polar_scan_matcher
:
rosdep version: 0.11.5
```

ipa-mdl:
`local-name` must match the `$TARGET_REPO_NAME`  to get filtered out. This way it is independent from the fork.

130s:
See the discussion in https://github.com/ros-industrial/industrial_ci/pull/155
We as in industrial_ci team won't handle this issue as it's most likely? something that the users/developers need to take care of and CI can't help. Please add comment if there are still things that we can do.

PR #155, opened 2017-03-23
[WIP] Handle rosdep error "Multiple packages found with the same name X" (workaround to #146).

130s:
**DO NOT REVIEW & MERGE YET** -- only an early prototype.

### DIFF
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -117,6 +117,18 @@ fi
 
 ici_time_start rosdep_install
 
+# After ROSWS, there may be duplicate packages existent in the workspace.
+# In that case we only want to keep the one from the target repo.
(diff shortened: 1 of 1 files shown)


ipa-mdl:
I had the same idea, but then decided against it..

Duplicated packages might exist because of these reasons:
* configuration error (e.g. wrong file): needs to be detected
* acutal name clash (different packages): Not cool. To be avoided.
* target repo in common upstream workspace file: legit use case

However, there is no proper way to determine if it's an error or intended.
The rosinstall reuse can be handled by the current implementation (requires a naming convention).

So in order to safely delete the duplicated packages we would have to ask the user for a list of paths.
But this basically the same as specifying `BEFORE_SCRIPT="rm -rf $CATKIN_WORKSPACE/src/TO_BE_DELETED"`.
(It gets executed just between the invocation  of `wstool` and `rosdep install`.)

130s:
https://github.com/ros-industrial/industrial_ci/pull/155#issuecomment-288883211 makes sense. Closing.

PR #158, opened 2017-03-27
unset empty CC and CXX variables

ipa-mdl:
This fixes a regression bug from #154 
It turned out that docker passes variabales as empty if it they have not been set on the host

This works for catkin and cmake, but might not work for custom/3rd-party Makefiles etc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,13 +15,16 @@ env:
   global:
     - VERBOSE_OUTPUT='true'
     - POST_PROCESS='echo "post processing reached"'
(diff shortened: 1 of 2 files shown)


130s:
Would you add a test to catch regression for this?

ipa-mdl:
> Would you add a test to catch regression for this?

good point!

ipa-mdl:
Added a test and a test for the test :D

ipa-mdl:
test of test detected an error ;) 
looks good now..

130s:
Thanks!

PR #161, opened 2017-03-29
Install sudo in docker image

jonazpiazu:
the docker image used for kinetic and lunar (ubuntu:xenial) does not have sudo installed
this triggers a failure in prepare_docker_image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,11 +41,12 @@ env:
     - ROS_DISTRO=indigo USE_DEB=false # Checking backup compatibility with UPSTREAM_WORKSPACE
     - ROS_DISTRO=indigo ADDITIONAL_DEBS="ros-indigo-opencv3" VERBOSE_OUTPUT='false'
     - ROS_DISTRO=indigo ADDITIONAL_DEBS="ros-hydro-opencv3" DEBUG_BASH='true' EXPECT_EXIT_CODE=100 # This should fail (trying from a wrong distro).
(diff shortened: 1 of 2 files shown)


ipa-mdl:
`sudo` already gets installed [here](https://github.com/ros-industrial/industrial_ci/blob/master/industrial_ci/src/docker.sh#L181)

jonazpiazu:
But it gets called before being installed in [here](https://github.com/ros-industrial/industrial_ci/blob/master/industrial_ci/src/docker.sh#L169)

ipa-mdl:
> But it gets called before being installed in here

Good point.

.travis.yml
```diff
@@ -44,8 +44,10 @@ env:
     - ROS_DISTRO=jade   APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet  _DO_NOT_FOLD=true # Passing wrong SKS URL as a break test. Should still pass.
     - ROS_DISTRO=jade   UPSTREAM_WORKSPACE=file  ROSINSTALL_FILENAME=.ci.rosinstall  # Testing arbitrary file name without ROS_DISTRO suffix. As of 6/3/2016 this fails due to https://github.com/ros-industrial/industrial_core/pull/144#issuecomment-223186764
     - ROS_DISTRO=jade   UPSTREAM_WORKSPACE=file ROSINSTALL_FILENAME=.i.do.not.exist EXPECT_EXIT_CODE=1
+    - ROS_DISTRO=jade APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet  # Passing wrong SKS URL as a break test. Should still pass.
```
ipa-mdl: this does not test anything in addtion to the indigo test, so I would omit for the sake of faster tests

ipa-mdl:
I have canceled the outdated test job. I will merge if the new one passes.

PR #164, opened 2017-04-19
Handle problems with the xenial docker image

jonazpiazu:
Xenial docker image does not come with apt-utils installed, which leads to several warning messages:
`debconf: delaying package configuration, since apt-utils is not installed`
This also highly delays the creation of the docker container.

Second problem shows this error/warning message:
```
++ rosdep install -q --from-paths /root/catkin_ws --ignore-src --rosdistro kinetic -y
++ grep 'executing command'
debconf: unable to initialize frontend: Dialog
debconf: (No usable dialog-like program is installed, so the dialog based frontend cannot be used. at /usr/share/perl5/Debconf/FrontEnd/Dialog.pm line 76, <> line 132.)
debconf: falling back to frontend: Readline
```

### DIFF
--- a/industrial_ci/src/docker.sh
+++ b/industrial_ci/src/docker.sh
@@ -162,7 +162,9 @@ function ici_build_default_docker_image() {
   ici_docker_build - <<EOF > /dev/null
 FROM ubuntu:$UBUNTU_OS_CODE_NAME
 
-RUN apt-get update -qq && apt-get -qq install --no-install-recommends -y wget ca-certificates sudo
+RUN apt-get update -qq \
+    && apt-get -qq install --no-install-recommends -y apt-utils \
+    && apt-get -qq install --no-install-recommends -y wget ca-certificates sudo
 
 RUN echo "deb ${ROS_REPOSITORY_PATH} $UBUNTU_OS_CODE_NAME main" > /etc/apt/sources.list.d/ros-latest.list
 RUN apt-key adv --keyserver "${APTKEY_STORE_SKS}" --recv-key "${HASHKEY_SKS}" \
--- a/industrial_ci/src/docker.sh
+++ b/industrial_ci/src/docker.sh
@@ -165,8 +165,7 @@ FROM ubuntu:$UBUNTU_OS_CODE_NAME
 RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
 
 RUN apt-get update -qq \
-    && apt-get -qq install --no-install-recommends -y apt-utils \
-    && apt-get -qq install --no-install-recommends -y wget ca-certificates sudo
+    && apt-get -qq install --no-install-recommends -y apt-utils wget ca-certificates sudo
 
 RUN echo "deb ${ROS_REPOSITORY_PATH} $UBUNTU_OS_CODE_NAME main" > /etc/apt/sources.list.d/ros-latest.list
 RUN apt-key adv --keyserver "${APTKEY_STORE_SKS}" --recv-key "${HASHKEY_SKS}" \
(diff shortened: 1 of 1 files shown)


industrial_ci/src/docker.sh
```diff
@@ -162,7 +162,11 @@ function ici_build_default_docker_image() {
   ici_docker_build - <<EOF > /dev/null
 FROM ubuntu:$UBUNTU_OS_CODE_NAME
 
-RUN apt-get update -qq && apt-get -qq install --no-install-recommends -y wget ca-certificates sudo
+RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
```
ipa-mdl: Is this the same as `ENV DEBIAN_FRONTEND noninteractive`?

industrial_ci/src/docker.sh
```diff
@@ -162,7 +162,11 @@ function ici_build_default_docker_image() {
...
+
+RUN apt-get update -qq \
+    && apt-get -qq install --no-install-recommends -y apt-utils \
```
ipa-mdl: please combine into one line

industrial_ci/src/docker.sh
jonazpiazu: From the [debconf manpage](http://manpages.ubuntu.com/manpages/xenial/man7/debconf.7.html):

>   You  can  change  the  default  frontend  debconf uses by reconfiguring
       debconf. On the other hand, if you just want to change the frontend for
       a  minute,  you can set the DEBIAN_FRONTEND environment variable to the
       name of the frontend to use.

jonazpiazu:
Failed test in Travis in prerelase should be fixed by this commit:

https://github.com/ros-infrastructure/ros_buildfarm/commit/0e7c4023a0269f22a63fb52ca483f0c1a3d9eaca

industrial_ci/src/docker.sh
ipa-mdl: I have seen `DEBIAN_FRONTEND` with docker in many places and no `debconf`. I think it is easier to get what `DEBIAN_FRONTEND` does, in addtion it's faster at build time.

ipa-mdl:
> Failed test in Travis in prerelase

For now we have to wait for a new release of `ros_buildfarm` :-/

jonazpiazu:
> For now we have to wait for a new release of ros_buildfarm :-/

Maybe meanwhile the prerelease test in Travis for kinetic should be skipped? (or allowed to fail)

industrial_ci/src/docker.sh
jonazpiazu: Setting DEBIAN_FRONTEND does not fix all the messages.

I have set a branch with the DEBIAN_FRONTEND to noninteractive [here](https://github.com/tecnalia-advancedmanufacturing-robotics/industrial_ci/blob/debconf_fix_test/industrial_ci/src/docker.sh#L165) , and then a [new test for kinetic](https://travis-ci.org/tecnalia-advancedmanufacturing-robotics/industrial_ci/jobs/223988245).

You can see in the [raw logs](https://s3.amazonaws.com/archive.travis-ci.org/jobs/223988245/log.txt) still some debconf messages complaining:

```
debconf: unable to initialize frontend: Dialog

debconf: (No usable dialog-like program is installed, so the dialog based frontend cannot be used. at /usr/share/perl5/Debconf/FrontEnd/Dialog.pm line 76, <> line 125.)

debconf: falling back to frontend: Readline
```

ipa-mdl:
Thanks for your contribution  (and patience)!

> Maybe meanwhile the prerelease test in Travis for kinetic should be skipped? (or allowed to fail)

I want to improve it anyway, failing tests are a  good motivation ;)

PR #165, opened 2017-04-25
use zesty for lunar tests

ipa-mdl:
Some code can be built for xenial, but not for zesty, e.g. because of the c++11 compiler.
This fix uses zesty for lunar tests instead of xenial per default.

### DIFF
--- a/industrial_ci/src/env.sh
+++ b/industrial_ci/src/env.sh
@@ -65,8 +65,11 @@ if [ -z "$UBUNTU_OS_CODE_NAME" ]; then
     "indigo"|"jade")
         UBUNTU_OS_CODE_NAME="trusty"
         ;;
-    "kinetic"|"lunar")
+    "kinetic")
         UBUNTU_OS_CODE_NAME="xenial"
+        ;;
+    "lunar")
+        UBUNTU_OS_CODE_NAME="zesty"
         ;;
     *)
         error "ROS distro '$ROS_DISTRO' is not supported"


ipa-mdl:
`catkin-tools` cannot be installed on zesty (and yakkety) because of https://github.com/osrf/osrf_pycommon/issues/35

130s:
What's the motivation behind this? With this change ROS Lunar test will always run on `Zesty` unless specified by `UBUNTU_OS_CODE_NAME`. Zesty will go EOL in 9 months while `Xenial` is LTS, so I don't know if we want to do this.

Also btw, is `UBUNTU_OS_CODE_NAME` expected to be passed by users if they want to?

ipa-mdl:
> What's the motivation behind this? 

zesty introduces a lot of new frameworks, especially gcc6, which enables c++14 by default.
So most code will work on lunar/xenial out-of-the-box, but might not be compilable on lunar/zesty.

> Zesty will go EOL in 9 months while Xenial is LTS, so I don't know if we want to do this.

Hmm, good point! We could switch back later on.. 

> Also btw, is UBUNTU_OS_CODE_NAME expected to be passed by users if they want to?

Yes. I guess we should document it ;)

ipa-mdl:
> This change makes Zesty as a default for Lunar, which may not be intuitive

You just have convinced me to not merge it ^^
I think it is better to document it properly and so provide an example how to test `lunar` on 'zesty`

Issue #167, opened 2017-05-05
Docker: command not found


Jmeyer1292:
Greetings my friends in the industrial CI world. Once again I find myself with things that worked before that are now failing.

A number of packages, including `abb_experimental` and `keyence_experimental`, have started breaking with a command to the effect of:
```
/home/travis/build/ros-industrial/abb_experimental/.ci_config/industrial_ci/src/docker.sh: line 108: docker: command not found
```

See logs [here](https://travis-ci.org/ros-industrial/abb_experimental/jobs/228861494) and [here](https://travis-ci.org/ros-industrial/keyence_experimental/jobs/214380270).

At a quick gander, I see that the packages that are failing have `xenial` listed as their travis distribution, but that could be coincidence. Could you please shed some light on this?

130s:
So the quick fix I've opened to the mentioned repos seems to work.

Travis CI does not support Xenial as of today https://github.com/travis-ci/travis-ci/issues/5821#issuecomment-243617289.
- My guess is that Travis on those repo used to run on Ubuntu Precise, the default distro for Ubuntu, since `distro: xenial` was not resolved. 
- Travis CI [just updated their Precise image](https://blog.travis-ci.com/2017-05-04-precise-image-updates). Not clarified in the post though, I assume something changed regarding `Docker`.

130s:
Inspired by this issue, I opened #168 to update "documentation", which hopefully clarifies the current requirement of industrial_ci on Travis CI.

I'm closing this but please feel free to reopen if needed.
Thank you always for detailed reports and opinion @Jmeyer1292

Jmeyer1292:
@130s I appreciate your work and your assistance a lot.

For something like this issue, where I had a rough idea of the issue, I'll still always try to make posts so that others who might have made the same mistake have some reference in the issue sections!

Let me know if I can help you out on developments going forward.

ipa-mdl:
"Fun" fact: Travis passes the distro without checks 
```
Build system information
Build language: generic
Build group: stable
Build dist: xenial
Build id: 228861489
Job id: 228861494
travis-build version: 66bd59b85
Build image provisioning date and time
Tue Apr 11 21:31:54 UTC 2017
Operating System Details
Distributor ID:	Ubuntu
Description:	Ubuntu 12.04.5 LTS
Release:	12.04
Codename:	precise
```
(https://github.com/travis-ci/travis-ci/issues/7296)

PR #168, opened 2017-05-05
[enhance][doc?] Add a note for 'distro' attribute in .travis.yml

130s:
Inspired by #167

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@
 
 #  Greatly inspired by JSK travis: https://github.com/jsk-ros-pkg/jsk_travis
 
+#sudo: required  # Not required by industrial_ci that uses Docker.
+#dist: trusty    # Not required by industrial_ci that uses Docker.
 services:
   - docker
 # Force travis to use its minimal image with default Python settings
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@
 
 #  Greatly inspired by JSK travis: https://github.com/jsk-ros-pkg/jsk_travis
 
-sudo: required
-dist: trusty
+services:
+  - docker
 # Force travis to use its minimal image with default Python settings
 language: generic
 notifications:


130s:
This `.travis.yml` file is referenced from the [readme](https://github.com/ros-industrial/industrial_ci#concrete-examples-of-config-files) so hope this change without adding doc in any other location is enough.

ipa-mdl:
According to the [travis docs](https://docs.travis-ci.com/user/docker/) the failsafe way is to specify.
```
sudo: required

services:
  - docker
````
This enforces to run the build on trusty
> Travis CI automatically routes builds to run on Trusty sudo: required when services: docker is configured.

**Comment by 130s** [comment]
Oh, okay, I didn't know that. But then, either `dist` or `services` must be specified if I understand correctly.
Since none of the [example config files listed in the readme](https://github.com/ros-industrial/industrial_ci/blob/a3336ff0ec0daf361ce881dd1951bdc113b0ac91/README.rst#id6) use this `services` attribute with `docker` so far, wouldn't it still reasonable to recommend `dist` (caveat is that when Travis updates their spec changes may have to be applied on all repositories)?

**Comment by ipa-mdl** [comment]
The code does not depend on a specific distro, but on docker.
I would suggest to include `dist` and `services`, the latter emphasizes that docker is required.

**Comment by 130s** [comment]
> I would suggest to include `dist` and `services`, the latter emphasizes that docker is required.

Ok, emphasizing that `docker` is required sounds much more beneficial.
Commit updated accordingly.

File: .travis.yml [hunk 1]
```diff
@@ -3,7 +3,9 @@
 #  Greatly inspired by JSK travis: https://github.com/jsk-ros-pkg/jsk_travis
 
 sudo: required
-dist: trusty
+services:
+  - docker
+dist: trusty  # Optional, as long as "services" is set with "docker". This will have to be updated once Travis CI supports newer Ubuntu distros.
```
ipa-mdl: > This will have to be updated once Travis CI supports newer Ubuntu distros.

There is no need to.

.travis.yml
```diff
@@ -3,7 +3,9 @@
 #  Greatly inspired by JSK travis: https://github.com/jsk-ros-pkg/jsk_travis
 
 sudo: required
```
ipa-mdl: this is optional as well.

130s:
Actually I removed some optional attributes in https://github.com/ros-industrial/industrial_ci/pull/168/commits/28b05e78f9f6d2e65bf2a430564e492ef9a6c983. Repos that still use them should still function and there won't be any change required.

130s:
@ipa-mdl sorry I added another commit after getting your approval and take back my previous commit. As the commit message reads:

Since `.travis.yml` in this repo serves as reference for new users in addition to its normal purpose, it's worthwhile to leave them even commented out.

ipa-mdl:
>  sorry I added another commit after getting your approval and take back my previous commit. 

I am still oky with it, so merge it if you are ready.

> Since .travis.yml in this repo serves as reference for new users in addition to its normal purpose

This needs to be changed, our `.travis.yml`is way too complicated for the regular users.
I'd like to introduce a new sample config in `doc/` with a lot of explanations.

130s:
Thanks for thoughtful comments as always @ipa-mdl.

> This needs to be changed, our .travis.ymlis way too complicated for the regular users.

So true. Current `.travis.yml` may not even function as a reference at all.

Issue #176, opened 2017-05-15
There's no way to NOT use install space


130s:
Opening as a separate ticket of what I poseted in https://github.com/ros-industrial/industrial_ci/pull/150#pullrequestreview-38160656.

Because `catkin config --install` is always run:
- there seems to be no way to NOT build into `install` space.
- If I'm right I want to add a feature to allow disabling `install` space.

Am I wrong?

This can cause an issue like https://github.com/ros-industrial-consortium/descartes/pull/198#issuecomment-301516448 where users want to test with the resources that are not deployed.
(For his case, I thought `NOT_TEST_INSTALL=true` could have worked but I was wrong; `NOT_TEST_INSTALL` just avoids running tests during the test steps for the `install` space.)

ipa-mdl:
> there seems to be no way to NOT build into install space.

For now `BEFORE_SCRIPT='catkin config -w $CATKIN_WORKSPACE --no-install'` is the way to go.

There are a couple of ways to make this more user-friendly:

1. remove `catkin config --install`, this would break the behaviour again, so :-1:
2. suppress `catkin config --install` with an opt-out flag like `USE_DEVEL_SPACE_ONLY`
3. `catkin config ${CATKIN_CONFIG:---install}`
4. `catkin config --install` followed by `if [ -n "$CATKIN_CONFIG" ]; then catkin config $CATKIN_CONFIG; fi`
5. suppressable `catkin config --install` followed by `if [ -n "$CATKIN_CONFIG" ]; then catkin config $CATKIN_CONFIG; fi` 

I'd prefer option 4, as a `CATKIN_CONFIG` could be beneficial for other use cases.
Option 5 is okay, but I'd like to avoid yet another specific flag.

130s:
Addressed in #177

PR #177, opened 2017-05-15
Allow passing 'catkin config' parameters

130s:
Issue discussed in #176 

> 4.    catkin config --install followed by if [ -n "$CATKIN_CONFIG" ]; then catkin config $CATKIN_CONFIG; fi
> 5.    suppressable catkin config --install followed by if [ -n "$CATKIN_CONFIG" ]; then catkin config $CATKIN_CONFIG; fi
>
> I'd prefer option 4, as a CATKIN_CONFIG could be beneficial for other use cases.
> Option 5 is okay, but I'd like to avoid yet another specific flag.

Agreed (if I understand your "followed by" correctly) and this PR is based on this idea.

TODOs

- [x] Unit test; Is `industrial_ci/mockups/industrial_ci_testpkg` not ready (I see `CATKIN_IGNORE` is placed in it)?
- [x] Documentation.


industrial_ci/src/tests/source_tests.sh
```diff
@@ -109,7 +109,7 @@ if [ "${USE_MOCKUP// }" != "" ]; then
     ln -s "$TARGET_REPO_PATH/$USE_MOCKUP" $CATKIN_WORKSPACE/src
 fi
 
-catkin config --install
+if [ -n "$CATKIN_CONFIG" ]; then catkin config $CATKIN_CONFIG; fi
```
ipa-mdl: `CATKIN_CONFIG`  needs to be added to `docker.env`

ipa-mdl:
Actually, I meant
```
catkin config --install
if [ -n "$CATKIN_CONFIG" ]; then catkin config $CATKIN_CONFIG; fi
```
This way `--install`  keeps set, even if additional options were provided, e.g. `--cmake-args -DCMAKE_BUILD_TYPE=Release`.

And `CATKIN_CONFIG=--no-install` is more explicit that `CATKIN_CONFIG=`.

> Is industrial_ci/mockups/industrial_ci_testpkg not ready (I see CATKIN_IGNORE is placed in it)?

It's ready, it can be built with ` USE_MOCKUP='industrial_ci/mockups/industrial_ci_testpkg'`
`CATKIN_IGNORE` is placed in the parent directory.

130s:
> ```
> catkin config --install
> if [ -n "$CATKIN_CONFIG" ]; then catkin config $CATKIN_CONFIG; fi
> ```

Isn't that actually redundant? See my updated commit.

> It's ready, it can be built with ` USE_MOCKUP='industrial_ci/mockups/industrial_ci_testpkg'` `CATKIN_IGNORE` is placed in the parent directory.

I see. I added another mockup to test a package that installs limited resources.

ipa-mdl:
> Isn't that actually redundant? See my updated commit.

Yes, it meant to be.
This way `--no-install` has to be passed explicitly and `--install` is still present in other cases.

> Add a mockup package that installs limited resource

For the unit test it would be better to add an install tag that points to something missing.
This way we can test that it breaks for the default case and works for `--no-install`.

130s:
>> Isn't that actually redundant? See my updated commit.
>
> Yes, it meant to be.
> This way `--no-install` has to be passed explicitly and `--install` is still present in other cases.

What about https://github.com/ros-industrial/industrial_ci/pull/177/commits/6f6bb854b6f01f2ee1661d1518e2e6433e64f4eb that achieves what you described but does not duplicate `catkin config` lines.

>> Add a mockup package that installs limited resource
>
> For the unit test it would be better to add an install tag that points to something missing.
> This way we can test that it breaks for the default case and works for `--no-install`.

That's efficient only as a testcase in `industrial_ci` although it is not practical ROS package if I understand correctly. Ok, I edited the mockup.

ipa-mdl:
> What about 6f6bb85 that achieves what you described but does not duplicate catkin config lines.

It does not meet my original requirements, as I want to enforce `--install` as long as `--no-install` not present. For the users both ways are the same, because `CATKIN_CONFIG` cannot be empty.
However, both serve the main purpose to use install space by default, so I am basically okay with either of them :)

>  That's efficient only as a testcase in industrial_ci although it is not practical ROS package if I understand correctly. Ok, I edited the mockup.

Yes, indeed! The only way to confirm that our tools work for a specific set-up is create mockups that force the tools to fail for a certain reason.
I'd suggest to rename `industrial_ci_testpkg2` to `broken_install` (or something like that).
This way it is clear that it is meant to be broken and no proper example.

industrial_ci/src/env.sh
```diff
@@ -16,6 +16,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+if [ ! "$CATKIN_CONFIG" ]; then export CATKIN_CONFIG="--install"; fi
```
130s: https://github.com/ros-industrial/industrial_ci/pull/177#issuecomment-301912881
> I want to enforce --install as long as --no-install not present.

Default value is exported here, so yes `CATKIN_CONFIG=--install` is enforced IMO. Is this not sufficient / not the way it should be?

130s:
Ok, mock package renamed.

ipa-mdl:
> Default value is exported here, so yes CATKIN_CONFIG=--install is enforced IMO. Is this not sufficient / not the way it should be?

I meant to make `CATKIN_CONFIG` the additional config.
If additional config options are provided, they should not  overwrite `--install`. With the exception of `--no-install`.

This could be achieved with:
```
catkin config --install
if [ -n "$CATKIN_CONFIG" ]; then catkin config $CATKIN_CONFIG; fi
```
Example: `CATKIN_CONFIG='-DMyCustomBuildFlag=true ...'` (``--install` is implicit)

Your version requires to specify `CATKIN_CONFIG='--install -DMyCustomBuildFlag=true ...'`.
Both option are valid, I don't know which option the user expects.

Both options work the same for `--no-install`.

ipa-mdl:
The extra files in `testpkg_broken_install` are not needed for the package's purpose, so they could be deleted to not confuse us later on ;)

130s:
@ipa-mdl Your explanation https://github.com/ros-industrial/industrial_ci/pull/177#issuecomment-303050604 convinced me you're right! I've added a commit to follow your suggestion.
Interesting to see if the added test captures the usecase we're talking about.

130s:
And hope the new mock pkg is concise enough now.

130s:
Thanks for an additional commit @ipa-mdl. Now good for the final review.

- Without `CATKIN_CONFIG` defined, job failed at [installing a missing folder](https://travis-ci.org/ros-industrial/industrial_ci/jobs/235197837#L533) as expected.
- With `CATKIN_CONFIG='--no-install`, [job passed](https://travis-ci.org/ros-industrial/industrial_ci/jobs/235197838).

130s:
Please review once I push one more commit for a document update.

ipa-mdl:
I guess the failing tests are somehow related to https://github.com/ros-infrastructure/catkin_pkg/pull/174

doc/index.rst
```diff
@@ -154,6 +154,7 @@ Note that some of these currently tied only to a single option, but we still lea
 * `BEFORE_SCRIPT`: (default: not set): Used to specify shell commands that run before building packages.
 * `BUILD_PKGS_WHITELIST` (default: not set): Packages to be built can be explicitly specified with this, in ROS package name format (i.e. using underscore. No hyphen). This is useful when your repo contains some packages that you don't want to be used upon testing. Downstream packages, if necessary, should be also specified using this. Also these packages are to be built when `NOT_TEST_INSTALL` is set. Finally, packages specified with this will be built together with those speicified using unimplmented `USE_DEB`.
 * `BUILDER` (default: catkin): Currently only `catkin` is implemented (and with that `catkin_tools` is used instead of `catkin_make`. See `this discussion <https://github.com/ros-industrial/industrial_ci/issues/3>`_).
+* `CATKIN_CONFIG` (default: not set): `catkin config --install` is used by default and with this variable you can 2) pass additional config options, or 2) overwrite `--install` by `--no-install`. See more in `this section <https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst#optional-customize-catkin-config>`_.
```
ipa-mdl: First `2) ` should be `1)`?

industrial_ci/mockups/testpkg_broken_install/script/sample_talker.py
```diff
@@ -0,0 +1,10 @@
+# Taken from http://wiki.ros.org/rospy/Overview/Publishers%20and%20Subscribers
```
ipa-mdl: The sample talker is not really needed for this package, but I am getting picky ;)

130s: I see, but we need something for [the `install` rule](https://github.com/ros-industrial/industrial_ci/pull/177/files/c1b8bbfddd5f307b207c4bdc7a02c09d76bf5449#diff-2b29029d69488436374c1e649c2236b7R9) (although a directory without any content should work), so let us just leave it for now.

130s:
Commit squashed and comments addressed. Thanks @ipa-mdl!

PR #179, opened 2017-05-25
Pass subversion credentials from host to docker container.

jonazpiazu:
Installing the dependencies of the package using rosinstall might require to access a subversion repository. We copy the ~/.subversion folder to the docker container to copy the authorization when connecting directly to the subversion server (not through ssh).


gavanderhoorn:
A general question - because I see `docker cp ..` being used in more places in `industrial_ci`: what is the rationale for using `docker cp` instead of working with volumes / bind mounts?

ipa-mdl:
> what is the rationale for using docker cp instead of working with volumes / bind mounts?

This was my work-round for permission problems, especially `ssh` is very restrictive.
I haven't found a better fail-safe solution.

Issue #181, opened 2017-06-17
Cache compiled code with ccache


VictorLamoine:
The cache is saved/restored and allows faster builds.
Very useful for big projects!

https://gitlab.com/VictorLamoine/ros_gitlab_ci/blob/master/ccache.bash
https://gitlab.com/VictorLamoine/ros_gitlab_ci/blob/master/.gitlab-ci.yml#L10-12

I'm pretty sure you can do the same with Travis CI.

ipa-mdl:
Thanks for pointing this out!

> I'm pretty sure you can do the same with Travis CI.

Wasn't too complicated: https://github.com/ipa-mdl/ros_canopen/blob/ccache/.travis.yml#L14
It saved 2 out of 6 minutes for `ros_canopen`.
After 2nd run: https://travis-ci.org/ipa-mdl/ros_canopen/builds/244099816

Right now I am not sure if we should just document this use case or even implement a `USE_CCACHE` flag.

ipa-mdl:
> Right now I am not sure if we should just document this use case or even implement a USE_CCACHE flag.

I have implemented a [feature](https://github.com/ipa-mdl/industrial_ci/tree/ccache) based on `CCACHE_DIR`, which integrates nicely with travis.
PR will follow as soon as travis backlosg is back to normal (4k jobs queued at the moment).

VictorLamoine:
Looks good :)

130s:
Addressed in #182

PR #169, opened 2017-05-05
test external projects again

ipa-mdl:
This PR introduces a new feature for internal-use.
it can be used to run `industrial_ci` for external repositories, just as travis would do.
That's why I have introduced an internal script `_wrap_test.sh`.

@Jmeyer1292, @shaun-edwards, @gavanderhoorn, @ipa-fxm, @130s: Do you have some suggestions for other candidates?
In the end it should be a nice mix of ~5 external repos.

PR #182, opened 2017-06-18
simplify ccache integration

ipa-mdl:
This PR simplifies the use of `ccache` as reported in #181 
It is not enabled by default because it needs set-up in the CI config.

Example config for travis:

```YAML
language: generic
services:
  - docker

cache:
  directories:
    - $HOME/.ccache # can be any valid cache location

env:
  global:
    - CCACHE_DIR=$HOME/.ccache # should match the cached directory above
  matrix:
    - ROS_DISTRO="kinetic" ROS_REPO=ros

install:
  - git clone --depth=1 https://github.com/ros-industrial/industrial_ci.git .ci_config
script: 
- .ci_config/travis.sh
```

A test-case can be added on top of #169


ipa-mdl:
I have run some tests with #169 

Without cache: https://travis-ci.org/ros-industrial/industrial_ci/builds/244234945
With cache: https://travis-ci.org/ipa-mdl/industrial_ci/builds/244239898

industrial_ci/src/tests/source_tests.sh
```diff
@@ -40,17 +40,24 @@ function catkin {
...
 fi
 source /opt/ros/$ROS_DISTRO/setup.bash
 
-ici_time_end  # setup_ros
+ici_time_end  # setup_apt
+
+if [ "$CCACHE_DIR" ]; then
+    ici_time_start setup_ccache
+    sudo apt-get install -qq -y ccache || error "Could not install ccache. Exiting."
+    export PATH="/usr/lib/ccache:$PATH"
```
130s: So this line looks cumbersome, but we may just need it https://askubuntu.com/questions/470545/how-do-i-set-up-ccache (I found @VictorLamoine there :))?

ipa-mdl:
> Beginner question; the feature in this PR is different from what Travis CI offers and is not dependent on it, correct? Just making sure because I think we don't want to Travis-specific.

It depends on the CI system to provide caching between the jobs.
Travis supports it, Gitlab supports it as well.
But they differ in location restrictions for caching.

> For smaller number of jobs, e.g. ros_canopen, any idea why the total amount of time actually increased?:

The handling of the cache takes time.

> Do we want to use this feature for an existing job or add a new job?

Travis supports only global caching (?), so the  handling is done for all jobs, even if the cache is not used in the job.
However, this is just an annoyance for our tests, but it should not affect our users.

industrial_ci/src/tests/source_tests.sh
ipa-mdl: Using symlink is the [recommended way](https://ccache.samba.org/manual.html#_run_modes) to enable ccache.
On debian/ubuntu  these symlinks get set-up automatically in `/usr/lib/ccache` by [update-ccache-symlinks](http://manpages.ubuntu.com/manpages/trusty/man8/update-ccache-symlinks.8.html)

130s:
Any comment on the document I updated?

>>    Beginner question; the feature in this PR is different from [what Travis CI offers](https://docs.travis-ci.com/user/caching/) and is not dependent on it, correct? Just making sure because I think we don't want to Travis-specific.
> 
> It depends on the CI system to provide caching between the jobs.
> Travis supports it, Gitlab supports it as well.
> But they differ in location restrictions for caching.

I see. This means that my document update was based on my misunderstanding, so definitely needs review.

>>    For smaller number of jobs, e.g. ros_canopen, any idea why the total amount of time actually increased?:
>
> The handling of the cache takes time.

Dumb question but do you mean caching may not work for the purpose for builds with a) smaller number of jobs b) jobs that take shorter c) else?

>>    Do we want to use this feature for an existing job or add a new job?
>
> Travis supports only global caching (?), so the handling is done for all jobs, even if the cache is not used in the job.
> However, this is just an annoyance for our tests, but it should not affect our users.

So...do we want to test this feature in our own build (i.e. do we want to enable this feature `.travis.yml` on industrial_ci) or not?

ipa-mdl:
> Any comment on the document I updated?

Not yet, I will have a look as soon as I have some time.

> > The handling of the cache takes time.

> Dumb question but do you mean caching may not work for the purpose for builds with a) smaller number of jobs b) jobs that take shorter c) else?

The cache feature (of travis) downloads and unzips an archive before **every** build.
And after each build, the content gets compared and uploaded if something has changed.
In addtion compiling with `ccache` can be slower because of the overhead it introduces.
So, caching with `ccache` only makes sense if there are many files to compile.

In https://travis-ci.org/ipa-mdl/ros_canopen/builds/244209632 only job 4 uses `ccache` to speed-up the build.
The other jobs just have the Travis-cache handling overhead.

> So...do we want to test this feature in our own build (i.e. do we want to enable this feature .travis.yml on industrial_ci) or not?

I did not plan to add it in this PR, but perhaps in #169
But we could give it a try to see what happens

doc/index.rst
```diff
@@ -249,6 +250,24 @@ Reference:
  * `Discussion about install space <https://github.com/ros-industrial/industrial_ci/issues/54>`_
  * `Detail for catkin config <http://catkin-tools.readthedocs.io/en/latest/verbs/catkin_config.html>`_ for more info about `catkin-tools`.
 
+Cache build artifacts to speed up the subsequent builds (if any)
+----------------------------------------------------------------
+
```
ipa-mdl: Instructions for setting up the CI cache are missing

doc/index.rst
```diff
@@ -249,6 +250,24 @@ Reference:
...
+If `CCACHE_DIR` is set (by default it is not), `ccache <https://en.wikipedia.org/wiki/Ccache>`_ gets enabled for your build to speed up the subsequent builds in the same job if anything.
+Recommended value is `$HOME/.ccache`, but any non-used directory works.
+
+All you need to do is to define `CCACHE_DIR` variable. You can apply to all of your jobs by something like below::
+
+  env:
+    global:
+      - CCACHE_DIR="$HOME/.ccache"
```
ipa-mdl: I guess the quotes must be removed in order to allow variable substitution.

ipa-mdl:
> The cache feature (of travis) downloads and unzips an archive before every build.
And after each build, the content gets compared and uploaded if something has changed.

This is basically true, but all jobs get individual caches, so the cache should  be empty if it wasn't used. So the overhead is more or less neglectable.

>  But we could give it a try to see what happens

I have run some tests, and for the current tests we dont gain any speed-up with `ccache` because we don't build code in most jobs,
In addition I don't have an idea how we can test that the cache feature was actually run..

VictorLamoine:
`ccache -s`

```
cache directory                     /home/victor/.ccache
primary config                      /home/victor/.ccache/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
cache hit (direct)                   353
cache hit (preprocessed)              71
cache miss                          1213
called for link                     1136
called for preprocessing             889
compile failed                        50
preprocessor error                    36
no input file                        420
files in cache                       437
cache size                          64.0 MB
max cache size                     100.0 MB
```

- If the cache is empty; it has never been filled by Travis.
- If there is not hit, the cache was un-used, something's wrong!

130s:
Thank you for clarification. Doc updated.

130s:
@ipa-mdl please review the updated doc.

ipa-mdl:
Docs look good now :+1: 
I will add tests based on @VictorLamoine's hints.

ipa-mdl:
I have modified two existing tests to check that `ccache` is not available by default and the cache got filled if it was enabled explicitly.
Please note that I haven't enable the Travis caching, because the tests will not benefit  from it for now.

.travis.yml
```diff
@@ -37,8 +37,9 @@ env:
     - ROS_DISTRO=indigo PRERELEASE=true USE_MOCKUP='industrial_ci/mockups/failing_test' EXPECT_EXIT_CODE=1
     - ROS_DISTRO=kinetic PRERELEASE=true PRERELEASE_REPONAME=industrial_ci
     - ROS_DISTRO=indigo APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet  # Passing wrong SKS URL as a break test. Should still pass.
-    - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=debian
-    - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=file  # Using default file name for ROSINSTALL_FILENAME
+    - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=debian AFTER_SCRIPT='ccache 2> /dev/null && exit 1; [ "$?" = "127" ]'
+     # Using default file name for ROSINSTALL_FILENAME, test CCACHE, verify cache was filled
+    - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=file CCACHE_DIR=$HOME/.ccache AFTER_SCRIPT='num=($(ccache -s | grep "files in cache")) && (( num[-1] > 0 ))'
```
130s: I'm trying to understand what the commands in `AFTER_SCRIPT` are doing.
Locally I get this, which seems not good..:
```
$ num=($(ccache -s | grep "files in cache")) && (( num[-1] > 0 ))
$ echo $num
files
```

ipa-mdl: > Locally I get this, which seems not good..:

actually, that's correct behaviour ;)

`$(ccache -s | grep "files in cache")` -> select line from ccache stats
`num=($(ccache -s | grep "files in cache"))` -> define an array with the data (separated by whitespace)
`num[-1]` -> selects the last entry of this array
`(( num[-1] > 0 ))` evaluates an arithmetic expression, the exit code is used as the actual test result

`echo $num` just prints the first item of the array. To print all you have to use `echo "${num[@]}"`.
`echo $?' will print the exit code:
```
$ num=($(ccache -s | grep "files in cache")) && (( num[-1] > 0 ))
$ echo $?
0 # okay
$ num=($(CCACHE_DIR=/tmp ccache -s | grep "files in cache")) && (( num[-1] > 0 ))
$ echo $?
1 # error
```

130s: My very bad...man `(ba)sh` is/will never be my friend...
Sounds good.

130s:
>>    So...do we want to test this feature in our own build (i.e. do we want to enable this feature .travis.yml on industrial_ci) or not?
>
> I did not plan to add it in this PR, but perhaps in #169
> But we could give it a try to see what happens

I think having job(s) that uses `CCACHE_DIR` is good enough for now even without testcases for it; we may not be validating the new feature (i.e. we're not checking if the new feature does what's supposed to) but we still are making there's no regression.

ipa-mdl:
>  we may not be validating the new feature (i.e. we're not checking if the new feature does what's supposed to) but we still are making there's no regression.

If the cache got filled, ccache should have been run :)
We don't have to test `ccache` itself.

PR #187, opened 2017-07-05
[gitlab] use .ci_config again

### LINKED ISSUE #186: Directory to clone industrial_ci into already exists in gitlab-ci
I'm attempting to configure `industrial_ci` in a GitLab hosted project and I'm encountering some problems with the default `.gitlab-ci.yml` provided as an example.

I'm using the docker executor, and the problem I have is that if I clone `industrial_ci` as suggested in that config file, i.e. cloning into parent folder with

`git clone https://github.com/ros-industrial/industrial_ci ../.ci_config` 

then the build container complains that the directory already exists:

```
...
$ git clone https://github.com/ros-industrial/industrial_ci ../.ci_config
fatal: destination path '../.ci_config' already exists and is not an empty directory.
ERROR: Job failed: exit code 128
```

I guess the gitlab runner is mounting some volumes from the runner itself into the executor, and therefore the `../.ci_config` directory exists from previous jobs. In fact, by changing to some other path, the job succeeds on the first attempt but then fails in every subsequent one.

I can see that this path was introduced in #185 to fix the tests for `industrial_ci` itself, avoiding duplicate `industrial_ci` copies in the source directory, but I think that it is breaking the `for reguar users` use-case.

ipa-mdl:
fixes #186 

@miguelprada: please review

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -8,19 +8,19 @@ services:
 
 before_script:
   - apk add --update bash coreutils # install industrial_ci dependencies
-  # for regular users: - git clone https://github.com/ros-industrial/industrial_ci ../.ci_config
-  - cp -a $CI_PROJECT_DIR ../.ci_config # this is needed for branch testing of industrial_ci itself
+  # for regular users: - git clone https://github.com/ros-industrial/industrial_ci .ci_config
+  - mkdir .ci_config && cp -a * .ci_config # this is only needed for branch testing of industrial_ci itself
 
 indigo:
-  script: ../.ci_config/gitlab.sh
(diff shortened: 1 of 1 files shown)


ipa-mdl:
Pipeline: https://gitlab.com/ipa-mdl/industrial_ci/pipelines/9601658

miguelprada:
Looks good to me. Thanks for acting so fast on this!

ipa-mdl:
I have now removed the `touch CATKIN_IGNORE`.
We should merge this after a new release of `catkin_pkg`.

ipa-mdl:
I will merge now.
My gitlab jobs will fail until the new version was released, but updating the example is more important.

PR #189, opened 2017-07-06
fix changelog encoding

ipa-mdl:
fixes #188 
In addition I fixed the spelling of my name ;)

PR #191, opened 2017-07-08
[WIP][capability] Add text file encoding checker.

130s:
Text file encoding issues like https://github.com/ros-industrial/industrial_ci/pull/189 can happen in any ROS packages. The proposed script checks utf-8 decoding error.

But unlike a very custom code like this, I hope there's better, well-managed solution already out there. I'm happy to withdraw this if such thing exists.

### DIFF
--- a/industrial_ci/src/ci_main.sh
+++ b/industrial_ci/src/ci_main.sh
@@ -40,6 +40,11 @@ else
   source ${ICI_SRC_PATH}/tests/source_tests.sh
 fi
 
+# Check text files (e.g. CHANGELOG.rst, package.xml) encode fine.
+ici_time_start textfile_encoding_check
+./test/textfile_encoding_check.py
+ici_time_end  # textfile_encoding_check
+
 if [ "${AFTER_SCRIPT// }" != "" ]; then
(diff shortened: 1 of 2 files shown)


industrial_ci/src/ci_main.sh
```diff
@@ -40,6 +40,11 @@ else
   source ${ICI_SRC_PATH}/tests/source_tests.sh
 fi
 
+# Check text files (e.g. CHANGELOG.rst, package.xml) encode fine.
+ici_time_start textfile_encoding_check
+./test/textfile_encoding_check.py
```
ipa-mdl: ${ICI_SRC_PATH}/tests/textfile_encoding_check.py

ipa-mdl:
IMHO this feature could be very useful!
Instead of crawling in python, I would prefer a (reusable) bash implementation.
The encoding check could be simplified to `python -c "file('$f').read().decode('utf-8')"`.

ipa-mdl:
> Instead of crawling in python, I would prefer a (reusable) bash implementation.

Implemented in #468

PR #192, opened 2017-07-10
[doc] Add undocumented variables DEBUG_BASH, VERBOSE_OUTPUT. Update pre/post process script usage.

130s:
The env vars were added in https://github.com/ros-industrial/industrial_ci/pull/93, https://github.com/ros-industrial/industrial_ci/pull/94.


ipa-mdl:
Do we really want users to know about `DEBUG_BASH`?
It is barely usable.

130s:
> Do we really want users to know about `DEBUG_BASH`?

Yeah. At least I as a user like to know quite often:
- Which output lines are the result of what command.
- Commands are executed as expected or not, with intended variables/arguments.

ipa-mdl:
Just ping me if it's ready for review again :)

> Yeah. At least I as a user like to know quite often:
> * Which output lines are the result of what command.
> * Commands are executed as expected or not, with intended variables/arguments.
Do you really can read this out of https://travis-ci.org/ros-industrial/industrial_ci/jobs/251937970?

That beeing said, I am okay with documenting it :)
But we should improve the readability.
My vision is to regroup the commands into smaller function block, e.g. `rosdep install`.

doc/index.rst
```diff
@@ -158,8 +158,9 @@ Note that some of these currently tied only to a single option, but we still lea
...
 * `CATKIN_PARALLEL_TEST_JOBS` (default: -p4): Maximum number of packages which could be examined in parallel during the test run. If not set it's filled by `ROS_PARALLEL_JOBS`.
-* `CCACHE_DIR` (default: not set): If set, `ccache <https://en.wikipedia.org/wiki/Ccache>`_ gets enabled for your build to speed up the subsequent builds in the same job if anything. See `detail. <https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst#run-pre-install-custom-commands>`_
+* `CCACHE_DIR` (default: not set): If set, `ccache <https://en.wikipedia.org/wiki/Ccache>`_ gets enabled for your build to speed up the subsequent builds in the same job if anything. See `detail. <https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst#run-pre-post-process-custom-commands>`_
```
ipa-mdl: This should link to `doc/index.rst#cache-build-artifacts-to-speed-up-the-subsequent-builds-if-any` (?)

ipa-mdl: And link in line 69 needs to be updated instead.

130s: Wow, those look bad. Thanks for picking up, and I've updated commits.

PR #198, opened 2017-07-26
don't deduce OS names from DOCKER_BASE_IMAGE

### LINKED ISSUE #197: Bad OS_CODE_NAME deduction from DOCKER_BASE_IMAGE
I've been having a lot of issues troubleshooting some errors and finally I've figured out that my problems came from the [automatic deduction of `OS_CODE_NAME` from the `DOCKER_BASE_IMAGE` tag](https://github.com/ros-industrial/industrial_ci/blob/c371892602e0e0af3f4abd5cf4b0e11d83badbc9/industrial_ci/src/env.sh#L78) when this option is used.

The tag of the image I was using was `base` and therefore the [ROS repository configuration](https://github.com/ros-industrial/industrial_ci/blob/c371892602e0e0af3f4abd5cf4b0e11d83badbc9/industrial_ci/src/docker.sh#L179) was completely wrong.

This can be fixed by making sure to specify `OS_CODE_NAME` in addition to `DOCKER_BASE_IMAGE`, but maybe some warning should be added somewhere in the docs.

The prerelease tests may also suffer from this since `OS_NAME` will also be nonsense, but I'm not familiar whith these at all...

ipa-mdl:
Fixes #197 

This should only affect users that specify `DOCKER_BASE_IMAGE`.
For source tests the codename will be read with `lsb_release`.
For the prelease the codname has to be set explicitly, `OS_NAME` defaults to `ubuntu` for the sake of convenience.

@miguelprada: please test / review.

TODO;
- [x] add tests
- [x] ~~update docs~~

### DIFF
--- a/industrial_ci/src/docker.sh
+++ b/industrial_ci/src/docker.sh
@@ -15,6 +15,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# docker.sh script sets up Docker image.
+# It is dependent on environment variables that need to be exported in advance
+# (As of version 0.4.4 most of them are defined in ./env.sh).
+
 #######################################
 # rerun the CI script in docker container end exit the outer script
 #
--- a/industrial_ci/src/docker.sh
+++ b/industrial_ci/src/docker.sh
@@ -163,7 +163,8 @@ function ici_prepare_docker_image() {
 #   (None)
 
 function ici_build_default_docker_image() {
-  export DOCKER_IMAGE="industrial-ci/$OS_CODE_NAME"
+  # choose a unique image name
+  export DOCKER_IMAGE="industrial-ci/$ROS_DISTRO/$DOCKER_BASE_IMAGE"
   ici_generate_default_dockerfile | ici_docker_build - > /dev/null
 }
 
@@ -174,9 +175,9 @@ FROM $DOCKER_BASE_IMAGE
 RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
 
 RUN apt-get update -qq \
-    && apt-get -qq install --no-install-recommends -y apt-utils gnupg wget ca-certificates sudo
+    && apt-get -qq install --no-install-recommends -y apt-utils gnupg wget ca-certificates sudo lsb-release
 
-RUN echo "deb ${ROS_REPOSITORY_PATH} ${OS_CODE_NAME} main" > /etc/apt/sources.list.d/ros-latest.list
+RUN echo "deb ${ROS_REPOSITORY_PATH} \$(lsb_release -sc) main" > /etc/apt/sources.list.d/ros-latest.list
 RUN apt-key adv --keyserver "${APTKEY_STORE_SKS}" --recv-key "${HASHKEY_SKS}" \
     || { wget "${APTKEY_STORE_HTTPS}" -O - | sudo apt-key add -; }
 
--- a/industrial_ci/src/docker.sh
+++ b/industrial_ci/src/docker.sh
@@ -169,7 +169,10 @@ function ici_prepare_docker_image() {
 function ici_build_default_docker_image() {
   # choose a unique image name
   export DOCKER_IMAGE="industrial-ci/$ROS_DISTRO/$DOCKER_BASE_IMAGE"
-  ici_generate_default_dockerfile | ici_docker_build - > /dev/null
+  echo "Building image '$DOCKER_IMAGE':"
+  local dockerfile=$(ici_generate_default_dockerfile)
+  echo "$dockerfile"
+  ici_docker_build - <<< "$dockerfile" > /dev/null
 }
 
 function ici_generate_default_dockerfile() {
--- a/industrial_ci/src/env.sh
+++ b/industrial_ci/src/env.sh
@@ -63,38 +63,36 @@ export OS_CODE_NAME
 export OS_NAME
 export DOCKER_BASE_IMAGE
 
-# exit with error if OS_NAME is set, but OS_CODE_NAME is not
-if [ -n "$OS_NAME" ] && [ -z "$OS_CODE_NAME" ]; then
+# exit with error if OS_NAME is set, but OS_CODE_NAME is not.
+# assume ubuntu as default
+if [ -z "$OS_NAME" ]; then
+    OS_NAME=ubuntu
+elif [ -z "$OS_CODE_NAME" ]; then
     error "please specify OS_CODE_NAME"
 fi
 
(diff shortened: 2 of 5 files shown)


miguelprada:
You catched the issue with `lsb-release` before I even had a chance to report it!

Running again after `547fcfd` worked where I had issues before :+1:

ipa-mdl:
> You catched the issue with lsb-release before I even had a chance to report it!

That's what the tests are for :D
I will squash it and try to update the documentation.

ipa-mdl:
I came the conclusion that the documentation does not need an update, because the automatic deduction was not documented at all.

industrial_ci/src/docker.sh
```diff
@@ -163,7 +163,7 @@ function ici_prepare_docker_image() {
 #   (None)
 
 function ici_build_default_docker_image() {
-  export DOCKER_IMAGE="industrial-ci/$OS_CODE_NAME"
+  export DOCKER_IMAGE="industrial-ci/${OS_CODE_NAME:-${DOCKER_BASE_IMAGE##*:}}"
```
130s: Here we're assuming `DOCKER_BASE_IMAGE` is already set in a script previously run (I think `env.sh` in this case). In other words `docker.sh` is dependent on another script(s). Would be great if we can leave inline comment for that.

industrial_ci/src/env.sh
```diff
@@ -73,11 +73,7 @@ if [ -n "$UBUNTU_OS_CODE_NAME" ]; then # for backward-compatibility
     OS_NAME=ubuntu
 fi
 
-if [ -n "$DOCKER_BASE_IMAGE" ]; then
-    # try to guess OS from default image scheme
-    OS_CODE_NAME=${OS_CODE_NAME:-${DOCKER_BASE_IMAGE##*:}} # use tag
-    OS_NAME=${OS_NAME:-${DOCKER_BASE_IMAGE%%:*}} # use repo
-else
+if [ -z "$DOCKER_BASE_IMAGE" ]; then
```
130s: Now with this line and the following `if` block, we're assuming that the specified Docker image contains ROS in it. Can we always assume that? E.g. what if there's anyone who wants to use certain Docker image that doesn't contain ROS?

ipa-mdl:
> Has OS_CODE_NAME become mandatory? https://travis-ci.org/ros-industrial/industrial_ci/jobs/257777731#L523

No, only for the (rare?) case that `DOCKER_BASE_IMAGE` was set externally while requesting a prerelease.
Just because there is no reliable way to determine the codename from the image name.
This questing made me think: It should be possible adapt the logic, the  `DOCKER_BASE_IMAGE` won't be used at all!

industrial_ci/src/docker.sh
ipa-mdl: Perhaps this change is not necessary.
However, `DOCKER_BASE_IMAGE` has to be provided anyway!

industrial_ci/src/env.sh
ipa-mdl: No, it is assumed for `DOCKER_IMAGE`, but not for `DOCKER_BASE_IMAGE`.

ipa-mdl:
Let's see if this does the trick :)

industrial_ci/src/docker.sh
```diff
@@ -15,6 +15,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# docker.sh script sets up Docker image.
+# It is dependent on environment variables that need to be exported previously 
+# (As of version 0.4.4 most of them are defined in ./env.sh).
+
```
ipa-mdl: Technically, this is true for all of our scripts.

ipa-mdl:
@130s: my commits are lost, did you force-push?

130s:
Sorry, I don't know how these happened but looks like I erased 2 of your recent commits, https://github.com/ros-industrial/industrial_ci/commit/45510b1 and https://github.com/ros-industrial/industrial_ci/commit/d1d3bb6 when I pushed my commit https://github.com/ros-industrial/industrial_ci/pull/198/commits/93cbffae271a2bbb3f86d64dd99fb9e15cb87ef7 into your branch. Do you think you can push them again? Those commits should be available on your local machine unless you already rebased to the remote.

Pushing to someone's remote branch directly is sometimes convenient but now I know it can be dangerous...I think I should've rebased first before force pushed.

ipa-mdl:
> Pushing to someone's remote branch directly is sometimes convenient but now I know it can be dangerous...I think I should've rebased first before force pushed.

Only force-pushing (`-f`) is dangerous. The first push wins.
I have rebased my commits on top of your one.

ipa-mdl:
> Those commits should be available on your local machine unless you already rebased to the remote.

They will be kept in the cache for a reasonable time span. `git reflog` is your friend for recovering from those accidents :)

ipa-mdl:
This PR should get squash-merged.

industrial_ci/src/docker.sh
130s: Added similar comment lines to other script files.

ipa-mdl:
Somehow my commit got lost again.
@130s: how do you push your commits?

130s:
A job for ROS Hydro keeps failing (e.g. [this](https://travis-ci.org/ros-industrial/industrial_ci/jobs/258395001#L584)). I think @ipa-mdl already reported https://github.com/ros-infrastructure/rosdep/issues/532 
```
No packages found matching ros-hydro-rospy-tutorials.
No packages found matching ros-hydro-rostest.

ERROR: Rosdep experienced an internal error.
Please go to the rosdep page [1] and file a bug report with the message below.
[1] : http://www.ros.org/wiki/rosdep

rosdep version: 0.11.6

Bad installer [apt]: [Errno 2] No such file or directory
Function rosdep_install returned with code '1' after 0 min 0 sec 
```

ipa-mdl:
> $ git push ipa-mdl fix/197-auto-deduction  --> Error
$ git push ipa-mdl fix/197-auto-deduction -f

The error tells you something has changed and you should include the change in your push.
Don't use `-f`, it just overwrite the changes.
You can resolve the error with `git pull --rebase   ipa-mdl fix/197-auto-deduction`
This your commits will be put on top of the changes.
I tend to rebase/squash my commit a lot, in this case the rebase might fail and the code needs to be adapted/merged manually.

ipa-mdl:
ready for final review :)

130s:
Thanks for iterating, patience for my git usage.
I'll merge despite the `rosdep` issue https://github.com/ros-industrial/industrial_ci/pull/198#issuecomment-318697631, as this is a fix to an existing issue #197

PR #199, opened 2017-07-26
implemented ABI check based on abi-compliance-checker

ipa-mdl:
This patch adds an ABI compatibility check based on https://github.com/lvc/abi-compliance-checker
It does prevent running the source tests and won't use the upstream workspace file.


shaun-edwards:
Awesome feature!  I didn't even know this was possible!

130s:
Honest question: don't we want to add a feature like this somewhere upstream, like build system (eg. Catkin)?

130s:
> don't we want to add a feature like this somewhere upstream, like build system (eg. Catkin)?

Build system might be the right place since there shouldn't be any ways for it to know anything about previous binary interface. 
I may be just being a devil's advocate here. I still feel somewhere upstream than CI pkg would be nicer but if we decide to add one here I'd be fine.

ipa-mdl:
> don't we want to add a feature like this somewhere upstream, like build system (eg. Catkin)?

The ABI dump could be integrated in the build system, but it requires a special build with specific flags. In addition you might not want to run it for the full workspace..

I could imagine that the build farm could run special ABI jobs that compare the current source with the last release. @dirk-thomas: Are you interested?

gavanderhoorn:
@ipa-mdl: we might be interested in this in the context of our ROSIN project. We've been looking into extending the farm with some more functionality (mostly CI) and this could fit with that.

@wasowski: what do you think?

dirk-thomas:
> I could imagine that the build farm could run special ABI jobs that compare the current source with the last release. @dirk-thomas: Are you interested?

Absolutely! That is a feature we long wanted to have but never had time to spend on.

ipa-mdl:
So for now I will complete this PR (tests, docs) and I would suggest to merge here.
This way we gather some user experiences with it.

Integrating it into the build farm is bigger chunk of work:
* add scripts(docker&exec) to  `ros_buildfarm` (probably based on this PR)
* updating the distribution yaml for (opt-in/opt.out/config)
* add job generators to  `ros_buildfarm`

This could be done as part of ROSIN (TBD @gavanderhoorn).

Eventually `industrial_ci` would use the `ros_buildfarm` scripts as it is the case with the prerelease tests.

gavanderhoorn:
Yes, we've been having some discussions about opt-in/out, and how to handle that.

I agree with the rest.

130s:
> So for now I will complete this PR (tests, docs) and I would suggest to merge here.
> This way we gather some user experiences with it.

+1
Let us know once this gets out of WIP status.

ipa-mdl:
It still lacks proper documentation, but it is ready for testing.

ipa-mdl:
> Is this error negligible?

Perhaps ;)
The latest commit should get rid of it.

industrial_ci/src/tests/abi_check.sh
```diff
@@ -0,0 +1,145 @@
...
+        local l=$(basename "$n" ".dump")
+        local o="/abicheck/old/$ABICHECK_VERSION/abi_dumps/$l.dump"
+        if [ -f "$o" ]; then
+            ici_time_start "abi_check_$l"
+            local ret=0
+            abi-compliance-checker -report-path "$reports_dir/$l.html" -l "$l" -n "$n" -o "$o" || ret=$?
+            if [ "$ret" -eq "0" ]; then
+                ici_time_end # abi_check_*
+            elif [ "$ret" -eq "1" ]; then
+                links -dump "$reports_dir/$l.html"
```
130s: So in the case of [a failing CI job](https://travis-ci.org/ros-industrial/industrial_ci/jobs/282457731) (which is a very nice addition!),
https://travis-ci.org/ros-industrial/industrial_ci/jobs/282457731#L1548
```
Preparing, please wait ...
Comparing ABIs ...
Comparing APIs ...
Creating compatibility report ...
Binary compatibility: 100%
Source compatibility: 100%
Total binary compatibility problems: 0, warnings: 0
Total source compatibility problems: 0, warnings: 0
Report: /abicheck/reports/0.7.1/libtopic_to_socketcan.html
Function abi_check_libtopic_to_socketcan returned with code '0' after 0 min 0 sec

Broken libraries: libcanopen_motor libcanopen_ros_chain
:
```

It'd be really great if we can see in the CI output for why the ABI check failed. 
Assuming some info is available in this report html file, we might want to `cat` it?

doc/index.rst
```diff
@@ -150,6 +150,10 @@ Optional environment variables
 
 Note that some of these currently tied only to a single option, but we still leave them for the future when more options become available (e.g. ament with BUILDER).
 
+* `ABICHECK` (default: false): If `true`, run a binary compatibility check with `ABICC <https://github.com/lvc/abi-compliance-checker>`_ and exit (see `below <#abi-checks>`_ for details).
```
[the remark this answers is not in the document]
ipa-mdl: `ABICHECK` activates the test, just like `PRERELEASE`.
The other variables are used to alter the behavior as explained in the detail section.
I will review your [PR](https://github.com/ipa-mdl/industrial_ci/pull/10), lets discuss the docs over there :)

industrial_ci/src/docker.env
```diff
@@ -1,4 +1,7 @@
 _DO_NOT_FOLD
+ABICHECK
+ABICHECK_URL
+ABICHECK_VERSION
```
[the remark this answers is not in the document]
ipa-mdl: `ABICHECK_URL` will be set to `ABICHECK_FALLBACK` before entering docker, so it is not needed. I should add a test for this as well.

industrial_ci/src/tests/abi_check.sh
ipa-mdl: The last section is just a summary.
The details can be found in the previous folds, the broken ones are yellow instead of green :)

ipa-mdl:
The merge-mode (`ABICHECK_FALLBACK`) is a little bit confusing.
Perhaps it would be better to only use the parent-detection only for pull requests.

What do you think of this API:
`ABICHECK`: enables the ABI check
`ABICHECK_MERGE`: enables the check against the merge parent. This will get set to true for PRs in `travis.sh` if it was not set. The test will fail if the HEAD commit is not a merge.
`ABICHECK_URL`:  used if `ABICHECK_MERGE` is false, e.g. for push tests

ipa-mdl:
I have added the `ABICHECK_MERGE` implementation.

ipa-mdl:
The docs still needs to be improved.

ipa-mdl:
@130s: I have pushed more updates, please review.

`ABICHECK_URL` is required to use the same config for branch tests and pull requests.
That's why I have removed `ABICHECK=true` and just for non-empty URL.
In addition the auto-detection or PRs must be enabled explicitly ( `ABICHECK_MERGE=auto`).

doc/index.rst
```diff
@@ -150,6 +151,9 @@ Optional environment variables
 
 Note that some of these currently tied only to a single option, but we still leave them for the future when more options become available (e.g. ament with BUILDER).
 
+* `ABICHECK_MERGE` (default: not set): Used only when `ABICHECK_URL` is set. For travis it can be set to 'auto' to auto-detect pull requests. If set to 'true' the merge parent (see `Terminology section <#terminology>`_) will be checked against.
```
130s: > will be checked against.

Against what ;) ?

doc/index.rst
```diff
@@ -254,6 +258,35 @@ The jobs that run Prerelease Test may usually take longer than the tests defined
...
+The following is a few examples of those URL:
+
+  - https://github.com/ros-planning/moveit/archive/0.9.9.tar.gz
+  - https://gitlab.com/YOUR_ORGANIZATION/YOUR_REPO/repository/ver_0.0.8/archive.zip
+  - https://github.com/ros-industrial-release/ros_canopen-release/archive/upstream.zip
+  - github:ros-planning/moveit#0.9.9
+
+The last example shows the shortcut from `provider:organization/repository#version`, which is supported for bitbucket, github and gitlab.
+
+ABI check examples:
```
130s: Below there are only `ssh`-based URL examples available, while in the URL example above there are more `http` URLs. This could give an impression that http is more common IMO, and we should have usecase example that uses http in addition.

(Wish I could open another PR for a small fix like this but today my wifi is very bad.)

doc/index.rst
ipa-mdl: The packages `industrual_ci` was run for.. as all our tests..

doc/index.rst
ipa-mdl: Where do we have "ssh-based URL examples"? Only curl-compatible protocols (http(s), ftp?) can be used.
The shortcut syntax is based on `npn install` URLs.

ipa-mdl:
> I kind of liked the simplest usecase where you just need to set a single variable ABICHECK to true. With the latest update, is there an alternative?

What should this "simplest usecase" check?
IMHO the simplest usecase is to specify a fixed code baseline for the ABI check.
This will even work in forks and for pull request against arbitrary branches..

doc/index.rst
130s: Sorry I don't know why I commented this https://github.com/ros-industrial/industrial_ci/pull/199#discussion_r146259313. Please ignore.

doc/index.rst
130s: I thought the one that starts from `git`, i.e. `github:ros-planning/moveit#0.9.9`, was an ssh format, but it isn't. 

What is `npn install`?

.travis.yml
```diff
@@ -62,6 +62,9 @@ env:
     - ROS_DISTRO=kinetic _EXTERNAL_REPO='ros-industrial/motoman_experimental#kinetic-devel' UPSTREAM_WORKSPACE=file BEFORE_SCRIPT='touch $CATKIN_WORKSPACE/src/ros-industrial/industrial_experimental/IRC_v2/CATKIN_IGNORE'
     - ROS_DISTRO=kinetic _EXTERNAL_REPO='ros-controls/ros_control#kinetic-devel' UPSTREAM_WORKSPACE=file ROSINSTALL_FILENAME=ros_control.rosinstall ROS_PARALLEL_TEST_JOBS=-j1
     - ROS_DISTRO=kinetic _EXTERNAL_REPO='ipa320/cob_calibration_data#c0922cc6f87f31641add6ba4e0fb88c93f9a7c6c' ROS_REPO=ros UPSTREAM_WORKSPACE=file AFTER_SCRIPT='./.travis.xacro_test.sh'
+    - ROS_DISTRO=indigo _EXTERNAL_REPO='ros/actionlib#38ce66e2ae2ec9c19cf12ab22d57a8134a9285be' ROS_REPO=ros ABICHECK_URL=url ABICHECK_MERGE=true # actual URL will not be used in the case
```
130s: > ABICHECK_URL=url 

What does this do? If this is a standard usage then please add document.

doc/index.rst
130s: I'm not familiar with this format `github:ros-planning/moveit#0.9.9` at all. Don't know where to get that for packages I'm willing to run abi check for.
Could you also document about this format?

130s:
Left some comments. While this feature looks very valuable, how we can start using it is still unclear to me.

> What should this "simplest usecase" check?

"usecase" was a wrong choice of word of me. I wanted to say "Usage".

When `ABICHECK` was there, you didn't have to specify anything else when you didn't need to.
Now `ABICHECK_URL` is a minimum requirement, but I'm afraid figuring out what value needs to be given could be a challenge for some(/many) users.

Can we give an automated option? For example, resurrect `ABICHECK` and make it a trigger for abi check, then if `ABICHECK_URL` is not passed then industrial_ci looks for the last tagged version? Just an idea.

doc/index.rst
ipa-mdl: > What is npn install?

Sry, I misspelled it. `npm` is the packages manager for nodejs.

> I'm not familiar with this format github:ros-planning/moveit#0.9.9 at all. Don't know where to get that for packages I'm willing to run abi check for.
Could you also document about this format?

It is documented in line 274, you just have to fill in the parts.
`github:ros-planning/moveit#0.9.9` will be resolved to `https://github.com/ros-planning/moveit/archive/0.9.9.tar.gz`, it is just shorter.

(Note to self: fix typo in line 274 from -> form)

.travis.yml
ipa-mdl: I need to set `ABICHECK_URL` to enable the ABI check, but it will not be used in this test (see comment).  If for some reason the url is used, the test will fail.
This is just a config for this test, nothing that should be done by the user!

ipa-mdl:
> Now ABICHECK_URL is a minimum requirement

No, it was required before as well, but it wasn't obvious.

> , but I'm afraid figuring out what value needs to be given could be a challenge for some(/many) users.

In most cases you just point it to your stable / released branch.
If we want to automate this, we could somehow read it from the `rosdistro` repository, but this does not work for unreleased packages.

> if ABICHECK_URL is not passed then industrial_ci looks for the last tagged version? Just an idea.

This is rather tricky. The tags might not be part of the fork (only if pushed manually) and the latest tag is not necessarily the latest one.

gavanderhoorn:
I feel it's perfectly fine to ask a developer that wants to include ABI checks for a URL to a version that should be considered the *baseline*. It could be a tag, a commit or even just `HEAD`.

That should not be hard to explain, and should not be too difficult to understand.

doc/index.rst
```diff
@@ -150,6 +151,9 @@ Optional environment variables
 
 Note that some of these currently tied only to a single option, but we still leave them for the future when more options become available (e.g. ament with BUILDER).
 
+* `ABICHECK_MERGE` (default: not set): Used only when `ABICHECK_URL` is set. For travis it can be set to 'auto' to auto-detect pull requests. If set to 'true' the merge parent (see `Terminology section <#terminology>`_) will be checked against.
+* `ABICHECK_URL` (default: not set): Run binary compatibility check with `ABICC <https://github.com/lvc/abi-compliance-checker>`_. The URL should point to a baseline archive (\*.tar.\*,\*.zip, \*.tgz or \*.tbz2), see more in `the ABI checks section <#abi-checks>`_)
```
130s: `ABICHECK_URL` points to a branch in [this example](https://github.com/ros-industrial/industrial_ci/pull/199/files#diff-0ced3dfd44ba156272e6db34a5247462R282), not a tag. If I'm not misunderstood, please add explanation here (or somewhere appropriate).

doc/index.rst
```diff
@@ -150,6 +151,9 @@ Optional environment variables
...
+* `ABICHECK_VERSION` (default: not set): Used only when `ABICHECK_URL` is set. Vversion name (for display only) of the set of code, which the location is specified in `ABICHECK_URL` of. The version will be automatically read from the URL passed in `ABICHECK_URL` if possible, but for a URL that doesn't point to a version-based file name (e.g. the link for a tagged version on Gitlab doesn't).
```
130s: Vversion :)

doc/index.rst
ipa-mdl: IMHO setting it to a branch is the best option for maintenance, but this requires a stable branch.
The maintainer can choose any baseline: Some might choose the latest release, some might choose the first release for a certain distro.

ipa-mdl:
> Ok, for the content of ABICHECK_URL, I follow the democracy. Maybe it's just me who feels a bit awkward for that format then.

You can still use `http(s):`. We might even support `git:` later (I want to unify the code in util.sh).
The shortcut version is just meant to be better readable, it contains only the necessary infomation.

ipa-mdl:
I have updated the docs and squashed all commits.

doc/index.rst
ipa-mdl: For clarification:  IMHO setting it to a *branch-based URL* is the best option for maintenance.

130s:
@ipa-mdl Could you review https://github.com/ipa-mdl/industrial_ci/pull/11 again?

130s:
Friendly ping @ipa-mdl

ipa-mdl:
I'll try to catch this up at the weekend.

130s:
CI failing for a few that activate ABI check. E.g. https://travis-ci.org/ros-industrial/industrial_ci/jobs/310854132#L568:
```
/root/ici/tests/abi_check.sh: line 19: sudo: command not found
The command "industrial_ci/_wrap_test.sh" exited with 127.
0.01s$ if ! [ -z "$POST_PROCESS" ]; then $POST_PROCESS; fi
"post processing reached"
```

ipa-mdl:
This makes sense, someone removed sudo ;)

industrial_ci/src/util.sh
```diff
@@ -136,3 +136,9 @@ function error {
     fi
     ici_exit $exit_code
 }
+
+if ! which sudo > /dev/null; then
```
130s: This portion seems like another magic of `(ba)sh` to me, and apparently there's nothing wrong nor confusing. It's just...impressive.

ipa-mdl: And it turned out to be wrong [somehow](https://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script).. (but works for us)
We should migrate it to `command -v` on the next occasion

PR #202, opened 2017-07-31
fix test wrapper to support commits

ipa-mdl:


industrial_ci/_wrap_test.sh
```diff
@@ -36,7 +36,9 @@ if [ -n "$_EXTERNAL_REPO" ]; then
         repo_url="https://github.com/$repo_name.git" # treat as github repo
     fi
 
-    git clone "$repo_url" -b "${_EXTERNAL_REPO##*#}" "$TRAVIS_BUILD_DIR"
+    git clone "$repo_url" "$TRAVIS_BUILD_DIR"
+    (cd "$TRAVIS_BUILD_DIR" && git checkout "$repo_branch")
```
130s: Why subshell/parentheses?

ipa-mdl: I have rewriiten all scripts to not change the directory anymore -> no side effects.
An easy way to accomplish this is to run directory changes in subshell.
I just found out about `--git-dir`, I will give it a try.

PR #203, opened 2017-07-31
downgrade rosdep to 0.11.5 in hydro test

ipa-mdl:
work-around for https://github.com/ros-infrastructure/rosdep/issues/532

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ env:
     - BEFORE_SCRIPT='echo current dir: $(pwd)'
 
   matrix:
-    - ROS_DISTRO=hydro  USE_MOCKUP='industrial_ci/mockups/industrial_ci_testpkg'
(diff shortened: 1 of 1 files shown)


ipa-mdl:
v0.11.7 should get released soon, let's see if the issue is resolved before merging this work-around.

ipa-mdl:
> Does this make sense to you @ipa-mdl?

Yes. The image contains `0.11.6` from the repositories.
The before script is run after that line and overlays `rosdep`  version `0.11.5`.

130s:
> The before script is run after that line and overlays `rosdep`  version `0.11.5`.

I got it, [indeed `BEFORE_SCRIPT` is run after that](https://github.com/ros-industrial/industrial_ci/blob/14d1f65380354fe0725f09269f7946f06b626c05/industrial_ci/src/tests/source_tests.sh#L130).

Issue #204, opened 2017-07-31
Stalled travis jobs


ipa-mdl:
The number of jobs (here and in other repositories) that hit the travis stall detection timeout (default: 10 minutes) increased.

Example:
https://travis-ci.org/ros-industrial/industrial_ci/jobs/259056635 errored, but https://travis-ci.org/ros-industrial/industrial_ci/jobs/258793803 passed (in 3 minutes), both built code for the same commit.

Looks like something got stuck, but I haven't found the problem yet.
SInce our code is unchanged, there might be some change upstream that does work properly with our scripts.

ipa-mdl:
Today it went [fine](https://travis-ci.org/ros-industrial/industrial_ci/jobs/259705831)

130s:
Can this be closed? Might #205 be taking effect?

130s:
Wait, I don't know if it's the same issue or not, but in one of our private repos on gitlab.com CI gets stuck at (seemingly) random place and finish with hitting 1 hour limit. I haven't seen jobs on Gitlab CI halt because of detected inactivity so the observation-wise it sounds pretty similar with what's reported here.

In our case it has always happened during `catkin run_tests` that normally takes 10 to maybe 25 mins (depending on the test case to be used). Some jobs passed but rarely.

ipa-mdl:
I would guess that travis fixed something, I know they had some issues as well.
I haven't restarted any job in weeks, so this can be closed.

> Wait, I don't know if it's the same issue or not, but in one of our private repos on gitlab.com CI gets stuck at (seemingly) random place and finish with hitting 1 hour limit.

If you can share any details, please open a new issue for this.

130s:
Ok, closing and I'll open a new issue as the issue I'm seeing is on Gitlab CI.

PR #205, opened 2017-07-31
run python unbuffered on travis

ipa-mdl:
does not solve it, but is related to #204 

Should this be the default in env.sh?


ipa-mdl:
> Why not add PYTHONUNBUFFERED to a few of our jobs?

This would make either not difference at all or the tests would get as unstable as they are now  :-/
I am even thinking of enforcing it in `docker.env` with no way to opt out.

130s:
> This would make either not difference at all 

At least we might be able to catch possible issues that is based on the upstream. No?

> or the tests would get as unstable as they are now :-/

I'm confused.
- I understand there can be cases where enabling unbuffering could stabilize the test (#204).
- This PR only provides an option to enable unbuffering.

Am I wrong?

ipa-mdl:
> This PR only provides an option to enable unbuffering.

Ah, that's where the confusion comes from.
This PR disables the buffering for all travis jobs (not only ours).

ipa-mdl:
@130s: I have removed the whitespace, please review again.

PR #209, opened 2017-08-02
Revert "downgrade rosdep to 0.11.5 in hydro test"

ipa-mdl:
Reverts ros-industrial/industrial_ci#203


ipa-mdl:
The hydro job work now with 0.11.7, but some others will fail because of https://github.com/ros-infrastructure/rosdep/issues/534

ipa-mdl:
Green again!

PR #214, opened 2017-09-14
prerelease test: prefer linked docker instance over socket interface

ipa-mdl:
Gitlab runners might have both available and the link must be used.

### DIFF
--- a/industrial_ci/src/tests/ros_prerelease.sh
+++ b/industrial_ci/src/tests/ros_prerelease.sh
@@ -25,12 +25,14 @@ function setup_environment() {
     WORKSPACE=$(mktemp -d)
     echo "WORKSPACE: $WORKSPACE"
 
-    if [ -e /var/run/docker.sock ]; then
+    if [ -n "$DOCKER_PORT" ]; then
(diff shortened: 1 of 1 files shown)


ipa-mdl:
Gitlab job passed: https://gitlab.com/ipa-mdl/industrial_ci/-/jobs/32646484
Tested it with `run_ci` and on a local gitlab runner.

industrial_ci/src/tests/ros_prerelease.sh
```diff
@@ -25,12 +25,14 @@ function setup_environment() {
     WORKSPACE=$(mktemp -d)
     echo "WORKSPACE: $WORKSPACE"
 
-    if [ -e /var/run/docker.sock ]; then
+    if [ -n "$DOCKER_PORT" ]; then
```
130s: Where is `DOCKER_PORT` passed from? If this is an input from user then could you add this to the variable section?

ipa-mdl: I think this is very specific to Gitlab and their services.
It looks like a Docker link, because of the other environment variables with only parts of the URI.

The (regular) users shold better not pass it directly without a special reason.

PR #219, opened 2017-10-17
Add catkin_lint feature.

130s:
An enhancement mentioned in https://github.com/ros-industrial/industrial_ci/issues/98

Some considerations:
- By default `catkin_lint` in this PR is run with `--strict` option, which makes the job fail even when there's only warning returned by catkin_lint. I'm not sure if we want to allow disabling --strict option on CI (those who just want the result without failing the CI job can run catkin_lint locally IMO).
- `catkin_lint` runs between build-install and tests, which I'm not sure is the best location.


130s:
Some CI jobs for `industrial_ci` (e.g. [for Debian Jessie](https://travis-ci.org/ros-industrial/industrial_ci/jobs/288844009)) fail where `python-catkin-lint` isn't available (from apt).

For Debian it's tracked already https://github.com/fkie/catkin_lint/issues/36 but no progress seems to have been made for half an year.

I don't know if we want to allow_failure those failing jobs just for catkin_lint.

130s:
All CI jobs for the repos used for downstream tests (`industrial_core`, `motoman_experimental`, `ros_control`) failed due to catkin_lint emit errors/warnings. Allowing failures for those jobs is easy but that could blind us for other potential issues in the future.

gavanderhoorn:
@130s wrote:
> All CI jobs for the repos used for downstream tests (`industrial_core`, `motoman_experimental`, `ros_control`) failed due to catkin_lint emit errors/warnings.

You almost make me feel bad ;)

In any case: I don't think `motoman_experimental` is really a good package for testing `industrial_ci`. It's not required to be stable and may contain pkgs that don't follow conventions / are in a flux, etc.

But +1000 for `catkin_lint` btw. The failed tests shows me quite some things that need fixing in that repository.

🙇

ipa-mdl:
Great!

> Some CI jobs for industrial_ci (e.g. for Debian Jessie) fail where python-catkin-lint isn't available (from apt). 

I would suggest to install catkin_lint only if  `CATKIN_LINT=true`.
In addition ` sudo pip install catkin_lint` should work for all platforms.
(Please make this is separate commit, this enables us to revert it later on)

> I don't know if we want to allow_failure those failing jobs just for catkin_lint.

Nope ;)

> catkin_lint runs between build-install and tests, which I'm not sure is the best location.

I would place it between `rosdep install` and `catkin build`.

> All CI jobs for the repos used for downstream tests (industrial_core, motoman_experimental, ros_control) failed due to catkin_lint emit errors/warnings. 

We should run a test for the `industrial_ci` package (if it fails, we should fix it) and a failing test with a mock up package.
This way we do not tp depend on actual package,which evenually will fix their problems.

> By default catkin_lint is run with --strict option,

The `--strict` option is a no-op for the default settings.
IMHO it should be possible warn about some problems, but to fail only for errors.
However, the `--explain` option is a good candidate to be passed for all runs.

industrial_ci/src/tests/source_tests.sh
```diff
@@ -158,6 +158,12 @@ if [ "$BUILDER" == catkin ]; then catkin build $OPT_VI --summarize  --no-status
 
 ici_time_end  # catkin_build
 
+if [ "$CATKIN_LINT" == "true" ]; then
+    ici_time_start catkin_lint
+    catkin_lint --strict $CATKIN_LINT_ARGS $CATKIN_WORKSPACE/src && echo "catkin_lint passed." || error "`catkin_lint --strict` failed by either/both errors and/or warnings"
```
ipa-mdl: IMHO the echo ouput are superfluous, the fold will highlight the error anyway.

ipa-mdl: If you check `$CATKIN_WORKSPACE/src`, the rosinstall will get included.
It ight be better to check only `"$CATKIN_WORKSPACE/src/$(basename $TARGET_REPO_PATH)"`

ipa-mdl: Or simply `"$TARGET_REPO_PATH"`.

130s: > the echo ouputs are superfluous, the fold will highlight the error anyway.

`echo` is called only upon success, so no problem?

130s:
Thanks @ipa-mdl @gavanderhoorn for the feedback, all of which should be now addressed.

industrial_ci/src/tests/source_tests.sh
ipa-mdl: It's is harder to read / maintain and the ouput does not add any information
So I would remove it, but I leave the decision to you.

industrial_ci/src/tests/source_tests.sh
```diff
@@ -149,6 +149,13 @@ set +o pipefail
 
 ici_time_end  # rosdep_install
 
+if [ "$CATKIN_LINT" == "true" ]; then
+    ici_time_start catkin_lint
+    sudo apt-get install -qq -y python-catkin-lint || error
```
ipa-mdl: `|| error` is not needed, the script will fail on all errors.
Please use the pip version for compatibility with Debian.

industrial_ci/src/tests/source_tests.sh
```diff
@@ -149,6 +149,13 @@ set +o pipefail
...
+    catkin_lint --explain $CATKIN_LINT_ARGS $TARGET_REPO_PATH && echo "catkin_lint passed." || error "`catkin_lint --strict` failed by either/both errors and/or warnings"
```
ipa-mdl: error message still contains `--strict` ;)

.travis.yml
```diff
@@ -56,7 +56,7 @@ env:
     - ROS_DISTRO=kinetic OS_NAME=debian OS_CODE_NAME=xenial EXPECT_EXIT_CODE=1
     - ROS_DISTRO=kinetic DOCKER_BASE_IMAGE="ros:kinetic-ros-base" NOT_TEST_BUILD='true' DEBUG_BASH='true' VERBOSE_OUTPUT='false'
     - ROS_DISTRO=kinetic APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet  # Passing wrong SKS URL as a break test. Should still pass.
-    - ROS_DISTRO=lunar ROS_REPO=ros-shadow-fixed USE_MOCKUP='industrial_ci/mockups/industrial_ci_testpkg'
+    - ROS_DISTRO=lunar ROS_REPO=ros-shadow-fixed USE_MOCKUP='industrial_ci/mockups/industrial_ci_testpkg' CATKIN_LINT=true
```
ipa-mdl: `industrial_ci_testpkg` should get patched to issue a warning, so we can add a supposed-to-fail test using the `--strict' flag.

130s: Do you mean we should use something else, not `industrial_ci_testpkg` for checking catkin_lint feature? If so do you have a suggestion?

ipa-mdl: > Do you mean we should use something else, not industrial_ci_testpkg for checking catkin_lint feature? If so do you have a suggestion?

No, that's fine.
But I want to add another teststhat fails on purpose to actually that `catkin_lint` was run for the given situation.

industrial_ci/src/tests/source_tests.sh
130s: I found the output of `catkin_lint` is often not clearly telling me if it passed or not, e.g. with the following that includes a line from my commit (`catkin_lint passed.`), if you're not that used to the tool then it could be hard to tell without my commit. With folded in green-color, most of the cases you might not even to unfold the section, but I'd still prefer to being explicit.

https://travis-ci.org/ros-industrial/industrial_ci/jobs/289293486
```
catkin_lint: checked 1 packages and found 1 problems
catkin_lint: 1 notices have been ignored. Use -W2 to see them
catkin_lint passed.
Function catkin_lint returned with code '0' after 0 min 3 sec 
```

> It's harder to read / maintain 

How so?

ipa-mdl: > I found the output of catkin_lint is often not clearly telling me if it passed or not

The state if encoded in "Function catkin_lint returned with code '0' after 0 min 3 sec" and the color.

>> It's harder to read / maintain

>How so?

`catkin_lint --strict $CATKIN_LINT_ARGS $CATKIN_WORKSPACE/src && echo "catkin_lint passed." || error "catkin_lint --strict failed by either/both errors and/or warnings"` 
(Beware: do not use the backticks in the string, it is the same as `$()`!)

vs.

`catkin_lint --strict $CATKIN_LINT_ARGS $CATKIN_WORKSPACE/src`

industrial_ci/src/tests/source_tests.sh
```diff
@@ -149,6 +149,14 @@ set +o pipefail
 
 ici_time_end  # rosdep_install
 
+if [ "$CATKIN_LINT" == "true" ]; then
+    ici_time_start catkin_lint
+    sudo apt-get install -qq -y python-pip
```
ipa-mdl: `python-pip` should already be installed in our images.
I am undecided if we want to enforce it here again.

industrial_ci/src/tests/source_tests.sh
```diff
@@ -151,8 +151,9 @@ ici_time_end  # rosdep_install
 
 if [ "$CATKIN_LINT" == "true" ]; then
     ici_time_start catkin_lint
-    sudo apt-get install -qq -y python-catkin-lint || error
-    catkin_lint --explain $CATKIN_LINT_ARGS $TARGET_REPO_PATH && echo "catkin_lint passed." || error "`catkin_lint --strict` failed by either/both errors and/or warnings"
+    sudo apt-get install -qq -y python-pip
+    sudo pip install python-catkin-lint
```
ipa-mdl: `sudo pip install catkin_lint`

ipa-mdl:
What do you think of `CATKIN_LINT=pedantic` as a shortcut for `CATKIN_LINT=true CATKIN_LINT_ARGS="--strict -W2"`?

130s:
@ipa-mdl All your comments and suggestions should be addressed. Please review again.

130s:
> I have modified the pedantic test to expect to fail.
> However, I did not want to sacrify our only lunar test, so I have duplicated the other.

+1
Thanks for adding the nice finish.

PR #220, opened 2017-10-25
WIP: [gitlab] Inject ssh key variable.

130s:
Allowing to pass the ssh key that can be set as a "secret variable" on gitlab repo ([reference](https://docs.gitlab.com/ee/ci/ssh_keys/README.html)). Defining the convoluted process like this each time is cumbersome, and also the script sample in the reference does not work out of the box ([e.g.](https://gitlab.com/gitlab-org/gitlab-runner/issues/1491)) so it'd be nice to provide a working utility code.

Testing this on a private repo now.


industrial_ci/src/docker.env
```diff
@@ -34,6 +34,7 @@ ROS_DISTRO
 ROS_LOG_DIR
 ROS_PARALLEL_JOBS
 ROS_PARALLEL_TEST_JOBS
+_SSH_PRIVATE_KEY
```
ipa-mdl: There is no need for this variable, you have to set-up everthing outside docker

gitlab.sh
```diff
@@ -21,8 +21,57 @@
...
+    set -x
+  
+    if [ -z "$_SSH_PRIVATE_KEY" ]; then
+        echo "SSH_PRIVATE_KEY is not found.";
+        return 1;
+    fi      
+    
+    # Install ssh-agent if not already installed, it is required by Docker.
+    # (change apt-get to yum if you use a CentOS-based image)
+    which ssh-agent || ( apt-get update -y && apt-get install openssh-client -y )
```
ipa-mdl: This depends on the executor and the images used for the executor.
For our default example it should be something like this: `apk --no-cache add openssh-client`.

industrial_ci/src/docker.env
130s: Is it so when I want to use this variable in industrial_ci's process (e.g. during `wstool update` in `setup_rosws`)?

ipa-mdl: The key will be provided by the ssh agent.

130s: `SSH_PRIVATE_KEY` is defined on gitlab-ci host os. In order for `wstool` to access a repo via ssh on Docker, this needs to be passed.

With the updated commit, I confirmed that Docker can access the external repo that requires the ssh key.

ipa-mdl: >  In order for wstool to access a repo via ssh on Docker, this needs to be passed.


No, it gets passed via ssh-agent. (which gets setup in #92).
Your `inject_sshkey` gets called before entering the ROS docker container.

130s: >  Your `inject_sshkey` gets called before entering the ROS docker container.

Even though [it's moved in to source_tests.sh](https://github.com/ros-industrial/industrial_ci/pull/220/files#diff-5e81b8b00dc0b1bfa5dcb392b8b6ec45R91)?

ipa-mdl: Now the ssh-agent gets spawned inside docker.
But we all already have two ways to pass SSH keys.
There's no reason to add a 3rd one.
It is very easy to spawn the agent outside.

130s: Yeah I totally forgot #92. Then I'm not sure how to use that. Will figure it out and open a PR for documentation hopefully.

ipa-mdl:
Some comments:
* The `ssh-agent` will be already installed in almost all cases.
* ` bash -c ' ssh-add <(echo "$SSH_PRIVATE_KEY")`' works
* It is really a security risk to relax the ssh restrictions, especially if used without a docker executor.
*  There seems to be no way to do this properly for all Gitlab CI runner flavors

I would prefer to give uses some hints on setting this up in their `gitlab-ci.yml`.
It basically boils down to 4 lines of `before_script`.

ipa-mdl:
I have play around a littlie bit in https://gitlab.com/ipa-mdl/industrial_ci/blob/pass_sshkey_envvar/.gitlab-ci.yml

ipa-mdl:
I could imaging this to be added to gitlab.sh (but not in general!):

```bash
if [ -n "$SSH_PRIVATE_KEY" ]; then
  # start SSH agent
  eval $(ssh-agent -s)
  # add key to agent
  ssh-add <(echo "$SSH_PRIVATE_KEY")
  mkdir -p ~/.ssh
  # setup known hosts
  echo "$SSH_SERVER_HOSTKEYS" >> ~/.ssh/known_hosts 
fi
```

```

130s:
Thanks so much @ipa-mdl for taking time. I understand that embedding code for this into the general industrial_ci might not make sense for security reason.
Meanwhile this needs documented, and also I'm still having an issue so I'm continuing in https://github.com/ros-industrial/industrial_ci/issues/223

Issue #147, opened 2017-03-18
UPSTREAM_WORKSPACE syntax


ipa-mdl:
The current syntax of `UPSTREAM_WORKSPACE` along with `ROSINSTALL_FILENAME` is confusing.
I'd like to switch to an extended moveit_ci syntax:

UPSTREAM_WORKSPACE cases:
* `debian` (default): as is
* `file`: 
  * `ROSINSTALL_FILENAME `is set, try
     1.  `$ROSINSTALL_FILENAME.$ROS_DISTRO`
     1.  `$ROSINSTALL_FILENAME`
  *  otherwise try
     1. `.travis.rosinstall.$ROS_DISTRO`
     1. `.travis.rosinstall`
     1. `.ci.rosinstall.$ROS_DISTRO`
     1. `.ci.rosinstall`
     1. `.rosinstall.$ROS_DISTRO`
     1. `.rosinstall`
* `http*`: as is
*  otherwise, try
   1.  `$UPSTREAM_WORKSPACE.$ROS_DISTRO`
   2.  `$UPSTREAM_WORKSPACE.`
   3. fail

This effectively combines both variables.
In addition the new `file` paths can reuse existiing files and don't need a travis prefix

Open issues: 
* Do we want to support multiple workspaces?

ahoarau:
Multiple workspaces would be really great !

ipa-mdl:
A new systax was implemented in #361

PR #222, opened 2017-10-27
Run BEFORE_SCRIPT as early as possible in Docker.

130s:
I'm not sure why `BEFORE_SCRIPT` currently runs kind of late in the pipeline.

In one my usecases, I want to run it to fetch a `.rosinstall` file from another repo and pass it to `wstool`. Since current location is too late and wstool fails to find the .rosinstall file.

### DIFF
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -58,6 +58,15 @@ source /opt/ros/$ROS_DISTRO/setup.bash
 
 ici_time_end  # setup_apt
 
+# execute BEFORE_SCRIPT in repository, exit on errors
+if [ "${BEFORE_SCRIPT// }" != "" ]; then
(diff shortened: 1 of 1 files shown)


ipa-mdl:
This will break all before scripts that assume the workspace to be already set-up properly.
The current location is quite convenient because everything is prepared to run, but you can still modify all data before the heavy tasks start.

> Since current location is too late and wstool fails to find the .rosinstall file.

This should work with `BEFORE_SCRIPT='wstool merge -t $CATKIN_WORKSPACE/src YOUR_URL; wstool update -t $CATKIN_WORKSPACE/src'`.
 
It looks like you want https://github.com/ros-industrial/industrial_ci/issues/147 with multiple workspaces...

130s:
I see. I was rather just curious.

> It looks like you want #147 with multiple workspaces...

I'm not sure about that.

PR #225, opened 2017-11-04
[doc] More elaboration BEFORE/AFTER_SCRIPT variables.

130s:
- Passing multiple {script/command}(s) are missing.
- Link to detail missing.


doc/index.rst
```diff
@@ -381,7 +381,7 @@ You may want to add custom steps prior to the setup defined in `./travis.sh <./t
 
 * You want to run `ros_lint` (`thi discussion <https://github.com/ros-industrial/industrial_ci/issues/58#issuecomment-223601916>`_ may be of your interest).
 
-In such cases, you can specify the script(s) in `BEFORE_SCRIPT` and/or `AFTER_SCRIPT` variables. For example::
+In such cases, you can specify the script(s)/command(s) in `BEFORE_SCRIPT` and/or `AFTER_SCRIPT` variables. For example::
```
ipa-mdl: :+1:  for "command(s)"!
I would even remove "script(s)"

doc/index.rst
```diff
@@ -390,6 +390,10 @@ In such cases, you can specify the script(s) in `BEFORE_SCRIPT` and/or `AFTER_SC
   script:
     - .ci_config/travis.sh
 
+Multiple scripts/commands can be passed using semi-colon as a delimitter, e.g.::
+
+    - BEFORE_SCRIPT="echo 'Here is command 1.';echo 'Here is command 2.';"
```
ipa-mdl: There are two possible issues with this example:
1. This first command (in general) might fail without failing the test. To it is better to use `command1 && command2`.
2. If you use double quotes at the outer level, variables will get expanded right away. 

`BEFORE_SCRIPT='echo "Here is command 1." && echo "Here is command 2."'`

130s:
Thanks @ipa-mdl. Updated.

doc/index.rst
130s: Those are generic `bash` tips but very valuable in this context. I decided to avoid adding those and instead just mention that the values for the variables in question follow general bash grammar.

ipa-mdl: The 2nd hint does not cover bash subsitution, but travis substitution:
`BEFORE_SCRIPT="echo $(pwd)"` will print the (outer) travis directory.
`BEFORE_SCRIPT='echo $(pwd)'` will print the (inner) docker directory.

doc/index.rst
```diff
@@ -390,6 +390,14 @@ In such cases, you can specify the script(s) in `BEFORE_SCRIPT` and/or `AFTER_SC
...
     - .ci_config/travis.sh
 
+Multiple commands can be passed, as in a general `bash` manner. Using semi-colon as a delimitter,::
+
+    - BEFORE_SCRIPT="ls /tmp/1; ls /tmp/2"
+
+Or making a chain of commands::
+
+    - BEFORE_SCRIPT="ls /tmp/1 && ls /tmp/2 || ls /tmp/3"
+
```
ipa-mdl: For more complicated cases the commands should go into a dedicated script:
`BEFORE_SCRIPT='./my_before_script,sh`

doc/index.rst
```diff
@@ -390,6 +390,14 @@ In such cases, you can specify the script(s) in `BEFORE_SCRIPT` and/or `AFTER_SC
   script:
     - .ci_config/travis.sh
 
+Multiple commands can be passed, as in a general `bash` manner. Using semi-colon as a delimitter,::
+
+    - BEFORE_SCRIPT="ls /tmp/1; ls /tmp/2"
```
ipa-mdl: Please use single quotes in these examples. So users won't copy error-prone settings.

130s:
Updated. Please review again @ipa-mdl

130s:
Friendly ping @ipa-mdl

130s:
Closing in favor of #234, as it's getting confusing to maintain multiple PRs for updating the same doc file.

PR #226, opened 2017-11-04
remove overlay driver setting from gitlab example

### LINKED ISSUE #213: [gitlab-ci] "/bin/bash: /root/ici/ci_main.sh: No such file or directory" error
I try to use insdustrial_ci on gitlab instead of my homemade scripts.
But the job failed. It tries to execute a file named /root/ici/ci_main.sh but it does not exist.
This path comes from here https://github.com/ros-industrial/industrial_ci/blob/6dfcee82109157888fb93a6619be4022a73162ed/industrial_ci/src/docker.sh#L40 but I don't know what is this folder named "/root/ici".
Here is the log.

```
Running with gitlab-ci-multi-runner 9.5.0 (413da38)
  on CFP0196 (01fff31c)
Using Docker executor with image docker:git ...
Starting service docker:dind ...
Pulling docker image docker:dind ...
Using docker image docker:dind ID=sha256:7731e67e3def0af852ea8ab3788063fc640ce359d399b86fe5a009e6d814f625 for docker service...
Waiting for services to be up and running...
Using docker image sha256:0b3f8c0899237f2dd669ed3e3ef941c90fbd30fb5129653121890336bd72f650 for predefined container...
Pulling docker image docker:git ...
Using docker image docker:git ID=sha256:9fbf23232776ff80a3e098c2f54794e0b581e9dccf81a1bbc0349ab5baf0d220 for build container...
Running on runner-01fff31c-project-318-concurrent-0 via CFP0196...
Fetching changes...
Removing .ci_config/
HEAD is now at dc8403f Add gitlab ci from ros industrial
Checking out dc8403f9 as master...
Skipping Git submodules setup
$ apk add --update bash coreutils
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/community/x86_64/APKINDEX.tar.gz
(1/8) Installing ncurses-terminfo-base (6.0-r8)
(2/8) Installing ncurses-terminfo (6.0-r8)
(3/8) Installing ncurses-libs (6.0-r8)
(4/8) Installing readline (6.3.008-r5)
(5/8) Installing bash (4.3.48-r1)
Executing bash-4.3.48-r1.post-install
(6/8) Installing libattr (2.4.47-r6)
(7/8) Installing libacl (2.2.52-r3)
(8/8) Installing coreutils (8.27-r0)
Executing busybox-1.26.2-r5.trigger
OK: 41 MiB in 27 packages
$ git clone https://github.com/ros-industrial/industrial_ci.git .ci_config
Cloning into '.ci_config'...
$ .ci_config/gitlab.sh
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
Building image 'industrial-ci/kinetic/ubuntu:xenial':
FROM ubuntu:xenial

RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections

RUN apt-get update -qq     && apt-get -qq install --no-install-recommends -y apt-utils gnupg wget ca-certificates sudo lsb-release

RUN echo "deb http://packages.ros.org/ros-shadow-fixed/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros-latest.list
RUN apt-key adv --keyserver "hkp://ha.pool.sks-keyservers.net" --recv-key "0xB01FA116"     || { wget "https://raw.githubusercontent.com/ros/rosdistro/master/ros.key" -O - | sudo apt-key add -; }

RUN sed -i "/^# deb.*multiverse/ s/^# //" /etc/apt/sources.list     && apt-get update -qq     && apt-get -qq install --no-install-recommends -y         build-essential         python-catkin-tools         python-pip         python-rosdep         python-wstool         ros-kinetic-catkin         ssh-client     && apt-get clean     && rm -rf /var/lib/apt/lists/*
Function prepare_docker_image returned with code '0' after 0 min 0 sec 
/bin/bash: /root/ici/ci_main.sh: No such file or directory
ERROR: Job failed: exit code 127
```

ipa-mdl:
It might break local builds if the overlay driver is not enabled in the kernel.
Instead a link to the Gitlab documentation was added.

closes #213

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,7 +1,9 @@
-# the following block is needed for the shared Docker-based runner
+# The following block is needed for the shared Docker-based runner
+# For local runners you might want to enable the overlay driver:
+# https://docs.gitlab.com/ce/ci/docker/using_docker_build.html#using-the-overlayfs-driver
+
 image: docker:git # docker and git clients
 variables:
-  DOCKER_DRIVER: "overlay" # recommended for docker-in-docker, needs overlay kernel module loaded on host
   TMPDIR: "${CI_PROJECT_DIR}.tmp" # needed for prerelease tests with Gitlab's docker-in-docker
 services:
   - docker:dind # enable docker-in-docker


.gitlab-ci.yml
```diff
@@ -1,7 +1,9 @@
-# the following block is needed for the shared Docker-based runner
+# The following block is needed for the shared Docker-based runner
+# For local runners you might want to enable the overlay driver:
+# https://docs.gitlab.com/ce/ci/docker/using_docker_build.html#using-the-overlayfs-driver
+
 image: docker:git # docker and git clients
 variables:
-  DOCKER_DRIVER: "overlay" # recommended for docker-in-docker, needs overlay kernel module loaded on host
```
130s: I've taken a look at the gitlab.com website for a few minutes, but it's not yet clear to me (i.e. more than an average user) what I should do for the shared runner on my Gitlab CI (i.e. DOCKER_DRIVER: "overlay" if I'm not mistaken). Why don't we keep this line, comment it out, and mention that this line is necessary for the shared runner (or anything appropriate)?

ipa-mdl: Setting ` DOCKER_DRIVER: "overlay"` (or "overlay2"?) is not enough.
If you don't do the other step like configuring your kernel, you will end up with the problem described in #213.
However, If you set-up you runner you can even enable the overlay driver [for all jobs](https://docs.gitlab.com/ce/ci/docker/using_docker_build.html#use-driver-for-every-project), so you don't have to include this in each yaml file.

You do not need any of this and can just got with the default, but it might be slower..
According to my test you don't save anything if you set it for the shared runners explicitly.

130s: > You do not need any of this and can just got with the default, but it might be slower..

Ok, so `overlay` driver is just an optional setting, not required. I've been misunderstanding that. Then, +1.

PR #230, opened 2017-11-05
implemented run_travis

ipa-mdl:
```
Usage: industrial_ci/scripts/run_travis [PATH] [RANGE*] [-- [ENV*]]

Parses the travis config in the given path or in the current working directory and runs all specified tests sequentially
If no range is given, the list of jobs will get printed.

The number of tests can be reduced by specifying one or more ranges:
* single job: 1 (only first)
* range: 2-3 (only second and third)
* open start, only as first range: -4 (jobs 1 to 4)
* open end, only as last range: 7- (job 7 and all following jobs)
* open range: - (all jobs)

Complex examples for a matrix wih 12 jobs:

* -4 7 8: runs jobs 1 2 3 4 7 8
* 1 7-9: runs jobs 1 7 8 9
* 1 7-9 11-: runs jobs 1 7 8 9 11 12
* -: runs all jobs

The jobs will be run in clean environments.
Only DOCKER_PORT, SSH_AUTH_SOCK, and TERM will be kept.
Additional variable names can be passed at the end.
```

It should work for most `industrial_ci` users, but I haven't tested it yet with configs that include escaped variables.

### DIFF
--- a/industrial_ci/scripts/run_travis
+++ b/industrial_ci/scripts/run_travis
@@ -1,0 +1,153 @@
+#!/usr/bin/env python
+from __future__ import print_function
+
+import os
+import os.path
+import re
+import subprocess
+import sys
+
+import yaml
+
+# find config file either in first argument or curret working directory
+def find_config_file(args, names):
+    def test_file(dir): # test if one of the file names is present
+        for n in  names:
+            p = os.path.join(dir, n)
+            if os.path.isfile(p):
+                return os.path.abspath(p)
+        return None
+
+    if len(args) >  0: # test first argument if available
+        p = test_file(args[0])
+        if p is not None:
+            return p, args[1:]
+    return test_file(os.getcwd()), args
+
+# parse range tuples from arguments
+def parse_ranges(args, offset=0):
+    r = []
+    def apply_offset(i): # adjust for user offsets
+        ao = int(i) + offset
+        if ao < 0:
+            raise ValueError("%s is not in supported range" % str(i))
+        return ao
+
+    for a in args:
+        if '-' in a: # range string
+            p1, p2 = a.split('-', 2)
+            if p1 == '' and len(r) == 0: # -X
+                p1 = 0
+            else: # X-[Y]
+                p1 = apply_offset(p1)
+            if p2 == '': # [X]-
+                r.append((p1, -1))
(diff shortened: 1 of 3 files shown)


industrial_ci/scripts/run_travis
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python
+from __future__ import print_function
+
```
130s: Copyrights?

ipa-mdl:
> What is this for?

It run all your travis tests locally. Like `run_ci`, but it reads the environment variables out of `.travis.yml`.

> I read the usage output but I can't come up with a good usecase

You can run all tests before pushing or if you haven't enabled tests in your fork.
Or you can just run a single test in case you want to fix a specific problem.

> Who is this for?

All users, same as for `run_ci`.

> If this is for the end-users, can you start documenting?

I will add this, there's not much to write about.
However, it would helpful to get a feedback an the usage first.

130s:
> It run all your travis tests locally. Like `run_ci`, but it reads the environment variables out of `.travis.yml`.

Sounds like a tremendous advantage! +1 for this addition.

ipa-mdl:
> IIUC, the added code can be customized for the formats of other CIs.

Yes, that was the intention. However, e.g. Gitlab config is too flexible to read the variable that simple.

ipa-mdl:
@130s: copyright header added

130s:
Tagged https://github.com/ros-industrial/industrial_ci/releases/tag/0.6.0

industrial_ci/scripts/run_travis
```diff
@@ -0,0 +1,169 @@
...
+The jobs will be run in clean environments.
+Only DOCKER_PORT, SSH_AUTH_SOCK, and TERM will be kept.
+Additional variable names can be passed at the end.
+
+""" % cmd)
+
+def main(scripts_dir, argv):
+    if '--help' in argv:
+        print_help(argv[0])
```
130s: Any reason you didn't use [argparse](https://docs.python.org/3/library/argparse.html) @ipa-mdl?

ipa-mdl: It bloats the code and is overkill for just printing the usage.

PR #234, opened 2017-11-17
[doc] Cleanup, update, elaborate 

130s:
See message and diff in each commit.

PR #236, opened 2017-11-24
make sudo optional

ipa-mdl:


130s:
What for do we need this change? I see other issues or PRs linked but I'm not sure.

Btw, I just noticed that setting `sudo` is not enforced in industrial_ci's document (I think it's required since industrial_ci tries some admin commands). If I'm right, will this change effect the document too?

ipa-mdl:
> What for do we need this change? I see other issues or PRs linked but I'm not sure.

Not all base images include sudo.
And if we manage to install sudo without sudo, then we don't need sudo ;)

> Btw, I just noticed that setting sudo is not enforced in industrial_ci's document (I think it's required since industrial_ci tries some admin commands). If I'm right, will this change effect the document too?

In general our scripts are run as root in a container.
If someon creates a different baseimage, the uses could be changed to non-root.
But since we need to install things the base image should set-up sudo for this user.
This documentation would go in the section on how to provide custom base images ;)

130s:
I see. I'll add a doc in https://github.com/ros-industrial/industrial_ci/pull/234

PR #237, opened 2017-11-24
Inject QEMU for cross-platform build

### LINKED ISSUE #235: Add support for ARM-based containers
Now that Docker supports ARM : 
https://discourse.ros.org/t/announcing-ros-docker-images-for-arm-and-debian/2467

That'd be a very useful feature to have in i-ci !

ipa-mdl:
If `DOCKER_IMAGE` or `DOCKER_BASE_IMAGE` point to a different platform, then `INJECT_QEMU` can be set to  the appropriate version that should get injected automatically.

For example for ARM builds, `INJECT_QEMU=arm` has to be set.

Please note:
* For this to work the `qemu-user-static` package needs to be installed on the host (or the CI machine): `sudo apt-get install qemu-user-static`
* There is no need to inject QEMU if it is already present in the selected image
* The build is much slower, it is not suitable for regular CI checks. It might be better to test certain branches only or to provide an image with all dependencies (and QEMU).

includes #236, closes #235

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -173,6 +173,7 @@ Note that some of these currently tied only to a single option, but we still lea
 * `DOCKER_BUILD_OPTS` (default: not set): Used do specify additional build options for Docker.
 * `DOCKER_RUN_OPTS` (default: not set): Used do specify additional run options for Docker.
 * `EXPECT_EXIT_CODE` (default: 0): exit code must match this value for test to succeed
+* `INJECT_QEMU` (default: not set): Inject static qemu emulator for cross-platform builds, e.g. `INJECT_QEMU=arm`. This requires to install `qemu-user-static` on the host. The emulated build might take much longer!
 * `NOT_TEST_BUILD` (default: not set): If true, tests in `build` space won't be run.
 * `NOT_TEST_INSTALL` (default: not set): If true, tests in `install` space won't be run.
 * `OS_CODE_NAME` (default: derived from ROS_DISTRO): See `this section for the detail <https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst#optional-type-of-os-and-distribution>`_.
--- a/industrial_ci/src/docker.sh
+++ b/industrial_ci/src/docker.sh
@@ -81,6 +81,12 @@ function ici_run_cmd_in_docker() {
      run_opts+=(-v "$CCACHE_DIR:/root/.ccache" -e CCACHE_DIR=/root/.ccache)
   fi
 
+  if [ -n "$INJECT_QEMU" ]; then
+    local qemu_path
+    qemu_path=$(which "qemu-$INJECT_QEMU-static") || error "please install qemu-user-static"
+    run_opts+=(-v "$qemu_path:$qemu_path:ro")
+  fi
+
   local cid
   cid=$(docker create \
       --env-file "${ICI_SRC_PATH}"/docker.env \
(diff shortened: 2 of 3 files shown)


ipa-mdl:
ahoarau: Can you give this PR a try for your use case?

ahoarau:
`rosrun industrial_ci run_ci DOCKER_IMAGE=arm64v8/ros:kinetic
`

You're right it's pretty slow, but it seems to work thanks !

ahoarau:
``` 
Function setup_rosdep returned with code '0' after 2 min 55 sec 
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
Writing /root/catkin_ws/src/.rosinstall

update complete.
Obtain deb binary for upstream packages.
catkin not available. Make sure python-catkin-tools is installed. See also https://github.com/ros-industrial/industrial_ci/issues/216
Function setup_rosws returned with code '1' after 0 min 4 sec 
``` 
My image might be wrong ?

ipa-mdl:
Please try `rosrun industrial_ci run_ci DOCKER_BASE_IMAGE=arm64v8/ros:kinetic INJECT_QEMU=arm`

ahoarau:
Yup sorry ... I did't read the PR..

ipa-mdl:
If you set `DOCKER_BASE_IMAGE`, the scripts will setup ROS all dependencies.
If you set `DOCKER_IMAGE` the dependencies needs to be installed already.

Our setup might conflict with `arm64v8/ros:kinetic` because by the shadow fixed repo is used by default.
So `DOCKER_BASE_IMAGE=arm64v8/ubuntu:xenial` might better.

ahoarau:
``` 
CMake Error at /usr/share/cmake-3.5/Modules/CMakeDetermineCCompiler.cmake:57 (message):
  Could not find compiler set in environment variable CC:

  ccache gcc-5.
Call Stack (most recent call first):
  CMakeLists.txt:2 (project)


CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
``` 
I'm getting this with `rosrun industrial_ci run_ci DOCKER_BASE_IMAGE=arm64v8/ros:kinetic INJECT_QEMU=arm`

ipa-mdl:
Have you set-up ccache locally?

ahoarau:
Again you are right .. it's building now ! But it's extremely slow. Is there a max_cpu option or max_allowed_ressource enabled somewhere ?

ipa-mdl:
> Again you are right .. it's building now ! But it's extremely slow. Is there a max_cpu option or max_allowed_ressource enabled somewhere ?

The docker container is not throttled. However, emulation is quite slow.
I don't know any other option to cross-compile for ARM.

ahoarau:
```
[build] Found '1' packages in 0.0 seconds.                                                                         
Starting  >>> catkin_tools_prebuild               
^C3744e08096b32d72ba781bf205ffba947dd0d1f996982862acd1f081ef6c80c4
```
Well it's basically stuck there. `CC=gcc-5` and `CXX=g++-5`

ahoarau:
@ipa-mdl Do you have any idea why it would be stuck ?

130s:
Before I start reviewing, I feel like waiting for the discussion to settle and the conflicts resolved.

ipa-mdl:
> @ipa-mdl Do you have any idea why it would be stuck ?

It get's stuck in my local tests as well.
Qemu is notorious for getting stuck as it does not support all syscalls :-/

ipa-mdl:
It worked with ubuntu 17.10.
Perhaps we can bundle a newer qemu version somehow..

> I feel like waiting for the discussion to settle and the conflicts resolved.

At least the conflicts are resolved :D

doc/index.rst
```diff
@@ -169,6 +169,7 @@ Note that some of these currently tied only to a single option, but we still lea
 * `DOCKER_BUILD_OPTS` (default: not set): Used do specify additional build options for Docker.
 * `DOCKER_RUN_OPTS` (default: not set): Used do specify additional run options for Docker.
 * `EXPECT_EXIT_CODE` (default: 0): exit code must match this value for test to succeed
+* `INJECT_QEMU` (default: not set): Inject static qemu emulator for cross-platform builds, e.g. `INJECT_QEMU=arm`. This requires to install `qemu-user-static` on the host. The emulated build might take much longer!
```
130s: I just assume the term `INJECT_QEMU` is intuitive for ARM users.

industrial_ci/src/docker.sh
```diff
@@ -177,6 +199,7 @@ function ici_build_default_docker_image() {
 
 function ici_generate_default_dockerfile() {
   cat <<EOF
+$inject_qemu
```
130s: Just wondering what this does. Also why is this in lower-case?
Since this line is not conditional to QEMU so all build is affected, it'd be nice to clarify.

ipa-mdl: > Just wondering what this does. 

Nothing (anymore), it is just a left-over from my tests with build stages, I will remove it.

> Also why is this in lower-case?

At first it was an internal variable (=lower case..)

130s:
Travis CI started failing. 
https://travis-ci.org/ros-industrial/industrial_ci/jobs/322946251#L502
```
standard_init_linux.go:185: exec user process caused "no such file or directory"
The command "industrial_ci/_wrap_test.sh" exited with 1.
```

ipa-mdl:
Should be fixed with last commit

ipa-mdl:
This is a little bit tricky, but it should work now.

PR #238, opened 2017-11-25
install latest docker in prerelease container

ipa-mdl:
Prerelease tests started to [fail](https://gitlab.com/ipa-mdl/industrial_ci/pipelines/14361205) on Gitlab because of TLS problems.
Updating to a recent docker version [resolves this](https://gitlab.com/ipa-mdl/industrial_ci/pipelines/14361018).

### DIFF
--- a/industrial_ci/src/tests/ros_prerelease.sh
+++ b/industrial_ci/src/tests/ros_prerelease.sh
@@ -38,15 +38,18 @@ function setup_environment() {
     docker build -t "industrial-ci/prerelease" - <<EOF > /dev/null
 FROM ubuntu:xenial
 
-RUN apt-get update -qq && apt-get -qq install --no-install-recommends -y wget ca-certificates sudo
+RUN apt-get update -qq && apt-get -qq install --no-install-recommends -y wget apt-transport-https ca-certificates sudo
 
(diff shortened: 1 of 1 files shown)


ipa-mdl:
> So core of this PR is to replace an obsolete pkg docker.io with docker-ce?

It is not obsolete, but outdated..

> Not sure why the issue only occurs on Gitlab CI.

I guess there is some compatibility issue with new servers (17.09.0-ce), old clients (1.13.1) and docker-in-docker-in-docker with TLS.

> Is this change unique to industrial_ci? Chance to escalate to somewhere upstream (e.g. ROS' prerelease test)?

Pretty unique. Nothing to escalate.

130s:
Thanks for clarification!

>>    So core of this PR is to replace an obsolete pkg docker.io with docker-ce?
>
> It is not obsolete, but outdated..

My thesaurus says they are synonyms. Maybe `obsolete` emphasizes "no longer in use" 🔤 ?

ipa-mdl:
> Maybe obsolete emphasizes "no longer in use" :abc: ?

At least this is a connation of "obsolet" in German ;)
I think I meant "dated".

Issue #223, opened 2017-10-29
How to access to private repo via ssh is unclear


130s:
Assigning myself. Likely this will end up adding documentation at most.

I'm still struggling after getting some advices in https://github.com/ros-industrial/industrial_ci/pull/220 from @ipa-mdl to `git clone` from a repo to a private repo on gitlab.

I suspect if the way the target repo and/or auth keys are setup on my side is wrong.

Using sample projects, from [sample_pub_repo](https://gitlab.com/130s/sample_pub_repo/blob/master/.gitlab-ci.yml) I'm trying to `git clone` [sample_private_repo](https://gitlab.com/130s/sample_private_repo) where a pub key is deployed. 

https://gitlab.com/130s/sample_pub_repo/-/jobs/38093210
```
Cloning into 'sample_private_repo'...
Warning: Permanently added 'gitlab.com,52.167.219.168' (ECDSA) to the list of known hosts.

Permission denied (publickey).

fatal: Could not read from remote repository.
```

Using a temporary branch of industrial_ci to print debug messages in Docker, `~/.ssh/known_hosts` being empty may seem critical to me, but I'm not sure.
```
DEBUG1 /tmp/ssh-6jyYxQbE2C9G/agent.27
DEBUG2 .ssh copied
total 32
drwxr-xr-x 2 root root 4096 Oct 29 01:59 .
drwx------ 1 root root 4096 Oct 29 01:59 ..
-rw-r--r-- 1 root root   35 Oct 29 01:59 config
-rw-r--r-- 1 root root    1 Oct 29 01:59 known_hosts

```

miguelprada:
I got it working in a less-than-stellar fashion. In the before script of my `.gitlab-ci.ym` I usually add:

```
before_script:
  ...
  - mkdir -p ~/.ssh
  - echo "$SSH_KEY_SECRET" > ~/.ssh/id_rsa
  - chmod og-r ~/.ssh/id_rsa
  - echo -e "Host my.gitlab.host.com\n\tStrictHostKeyChecking no\n\n" > ~/.ssh/config
```

This is being run by a Docker executor based on `docker:git` and using `docker:dind` as a service.

I do feel that there must be a better way, though.

ipa-mdl:
Has anyone of you actually tried [this](https://github.com/ros-industrial/industrial_ci/pull/220#issuecomment-340076453)?

It has two advantanges:
1. The private key will not get revealed
[...]

PR #242, opened 2017-12-04
Replace ~ with $HOME for correct expansion.

miguelprada:
Some jobs relying on my SSH credentials being copied into the docker container have started to fail. I believe this might be because the `~` character is not correctly expanded inside double quotes [here](https://github.com/ros-industrial/industrial_ci/blob/a0356c3a546d43a1bea819d7ffdb0f61ea849e7c/industrial_ci/src/docker.sh#L89), which were introduced in #240. Replacing with `$HOME` does the job for me.

### DIFF
--- a/industrial_ci/src/docker.sh
+++ b/industrial_ci/src/docker.sh
@@ -86,8 +86,8 @@ function ici_run_cmd_in_docker() {
 
   # pass common credentials to container
   for d in .docker .ssh .subversion; do
-    if [ -d "~/$d" ]; then
-      docker cp "~/$d" "$cid:/root/"
+    if [ -d "$HOME/$d" ]; then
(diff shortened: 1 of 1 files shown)


130s:
Thanks for catching this. If you don't mind educating me...in what kind of environment does the code without this change fail?

ipa-mdl:
Probably https://github.com/ros-industrial/industrial_ci/pull/179 ^^

miguelprada:
@ipa-mdl @130s I also rely on this to copy credentials in `.ssh` to my GitLab CI executor, since I'm unsure about how to properly use `ssh-agent`. :sweat_smile: 

Hopefully #223 ends up leading somewhere :wink:

Issue #247, opened 2017-12-15
[Travis] Invalid SSH keys/config for wstool update


ipa-jba:
To fix the error ocurring since the latest update of travis I had to change the permissions of the .ssh folder.

See: https://github.com/ipa-jba/industrial_ci/commit/3cb5be401afac22d76388bd14f71719002f7d58e

However if this is a good solution is up for disussion.

@ipa-fxm @ipa-mdl

ipa-mdl:
I think this is unrelated to the new travis image, but might be an issue with the new docker version.
`docker cp` used to create files as root, but this seems to have changed.

This new version broke some other things (not related to `industrial_ci`) for me as well.

ipa-mdl:
Looks like a [bug](https://github.com/moby/moby/issues/34096) in docker.

ipa-mdl:
A fix for docker was [proposed](https://github.com/moby/moby/pull/35874).

ipa-mdl:
I guess this issue was resolved.

PR #248, opened 2017-12-16
set proper uid/gid for files copied to docker

ipa-mdl:
addresses #247 

Since docker 17.06 ownership of files copied into docker containers is kept instead of beeing rewritten to the target user. This PR emulates the old behavior until this bug was fixed in docker (WIP).

In addition this PR enables passing  credential files even for non-root images.

This patch might break again as soon as the docker bug is fixed.

@ipa-jba: please test

### DIFF
--- a/industrial_ci/src/docker.sh
+++ b/industrial_ci/src/docker.sh
@@ -103,6 +103,8 @@ function ici_run_cmd_in_docker() {
   return $ret
 }
 
+# work-around for https://github.com/moby/moby/issues/34096
+# ensures that the permissions of copied files are owned by the target user
 function docker_cp {
   set -o pipefail
   tar --owner=${docker_uid:-root} --group=${docker_gid:-root} -c -f - -C "$(dirname $1)" "$(basename $1)" | docker cp - $2
--- a/industrial_ci/src/docker.sh
+++ b/industrial_ci/src/docker.sh
@@ -36,6 +36,9 @@ function ici_require_run_in_docker() {
   if ! [ "$IN_DOCKER" ]; then
     ici_prepare_docker_image
 
+    docker_uid=$(docker run --rm $DOCKER_IMAGE id -u)
+    docker_gid=$(docker run --rm $DOCKER_IMAGE id -g)
+
     local docker_target_repo_path=/root/src/$TARGET_REPO_NAME
     local docker_ici_src_path=/root/ici
     ici_run_cmd_in_docker -e "TARGET_REPO_PATH=$docker_target_repo_path" \
@@ -87,7 +90,7 @@ function ici_run_cmd_in_docker() {
   # pass common credentials to container
   for d in .docker .ssh .subversion; do
     if [ -d "$HOME/$d" ]; then
-      docker cp "$HOME/$d" "$cid:/root/"
+      docker_cp "$HOME/$d" "$cid:/root/"
     fi
   done
 
@@ -100,6 +103,11 @@ function ici_run_cmd_in_docker() {
   return $ret
 }
 
+function docker_cp {
+  set -o pipefail
+  tar --owner=${docker_uid:-root} --group=${docker_gid:-root} -c -f - -C "$(dirname $1)" "$(basename $1)" | docker cp - $2
+  set +o pipefail
+}
 #######################################
 # wrapper for docker build
 #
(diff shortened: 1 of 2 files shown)


ipa-jba:
I checked it with one of our repositories and it seems fine

ipa-mdl:
Gitlab test: https://gitlab.com/ipa-mdl/test-ssh/-/jobs/45013697

ipa-mdl:
This does not work for Gitlab because alpine's `tar` does not support the owner parameters..

ipa-mdl:
So installing tar in .gitlab-ci.yml [did the trick](https://gitlab.com/ipa-mdl/test-ssh/-/jobs/45016363).
However, this PR will break all exisiting installations of private SSH-based Gitlab set-up (with an easy fix!).
Set-ups without SSH are [not affected](https://gitlab.com/ipa-mdl/industrial_ci/pipelines/15251773).

@130s, @miguelprada: What do you think?

miguelprada:
Since the fix would only entail to add `tar` to the list of installed packages in the CI config, that doesn't seem like a lot of trouble to me.

However, I fail to see what is being fixed by this. Do you have an example of what #247 refers to by _the error ocurring since the latest update of travis_?

ipa-jba:
I attach a log with the relevant lines
```
Function setup_rosdep returned with code '0' after 0 min 11 sec 
setup_rosws
0.71s>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
Writing /root/catkin_ws/src/.rosinstall
update complete.
     Performing actions: 
     Add new elements:
  <list of private repositories that are and were accessible with the ssh key>
Config changed, maybe you need run wstool update to update SCM entries.
Overwriting /root/catkin_ws/src/.rosinstall
update complete.
[repository_1] Fetching git@github.com:ipa320/repository_1.git (version indigo_dev) to /root/catkin_ws/src/repository_1
Cloning into '/root/catkin_ws/src/repostiory_1'...
Bad owner or permissions on /root/.ssh/config
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.
…
Exception caught during install: Error processing 'repository_1' : repository_1] Checkout of git@github.com:ipa320repository_1.git version indigo_dev into /root/catkin_ws/src/repository_1 failed.
…
Function setup_rosws returned with code '1' after 0 min 0 sec 
The command ".ci_config/travis.sh" exited with 1.
Done. Your build exited with 1.
```

but as @ipa-mdl mentioned 
> SSH checks the permissions of the .ssh files to match the current user and fails if this is not the case. 

> For Gitlab this does not have any effect because it uses root in both places, but Travis is using a dedicated user to run the builds.
Travis uses "travis" as user and group whereas industrial-ci uses "root" (the same as gitlab)

> So at the moment all private Travis builds are broken.
Strange we seem to be the only ones noticing/submitting an issue

miguelprada:
> For Gitlab this does not have any effect because it uses root in both places, but Travis is using a dedicated user to run the builds.

This was the bit that I was missing.

As I said before, adding tar to my setups is not a big nuisance for me, specially since I'm going to update them anyways to use #243.

ipa-mdl:
>> So at the moment all private Travis builds are broken.

> Strange we seem to be the only ones noticing/submitting an issue

This only affects private repositories which are using docker for testing AND pulling code from other private repositories inside the container..

industrial_ci/src/docker.sh
```diff
@@ -100,6 +103,11 @@ function ici_run_cmd_in_docker() {
   return $ret
 }
 
+function docker_cp {
```
130s: Am I correct that once the relevant issue in `docker` is resolved this function won't be necessary? If so I'd add a comment here just as a reminder (since I easily forget..), but I leave that up.

ipa-mdl:
No change is needed for Travis users.
But gitlab users need to install [tar](https://github.com/ros-industrial/industrial_ci/pull/248/commits/b262f307cbf6298204b86e2c5172f980102802d4) **if they use SSH forwarding etc.**.

industrial_ci/src/docker.sh
ipa-mdl: Not sure if we want to remove later..
But I will add a small comment anyway.

130s:
Thanks for the confirmation @ipa-mdl

130s:
Made [a quick announcement](https://discourse.ros.org/t/industrial-ci-0-5-0-available-abi-compatibility-check-catkin-lint-arm-support-etc/3403/2).

PR #255, opened 2018-01-04
non-verbose catkin run_tests

ipa-fxm:
fixes #253 

see https://github.com/ros-industrial/industrial_ci/issues/253#issuecomment-355256633 for explanations
I'll keep the discussion for this PR in #253

Issue #172, opened 2017-05-07
Add BUILD_PKGS_BLACKLIST?


gavanderhoorn:
This would be the opposite of `BUILD_PKGS_WHITELIST`.

Example use-case: `motoman_experimental` depends on `industrial_experimental`, for a single package only: `industrial_robot_pkg_gen`. As `industrial_experimental` has not been released, I'm using a `.travis.rosinstall` file. Everything works, but: `industrial_experimental` also contains `IRC_v2` packages, which run a lot of tests.

Testing is good, but `motoman_experimental` does not use anything from `IRCv2`, and I'd rather ignore the `IRCv2` packages completely in the Travis runs for `motoman_experimental`.

Instead of having to list each and every pkg I do want tested, it would be nice if we could invert the setting and use a blacklist.

ipa-mdl:
I have started to implemented this (other features) in https://github.com/ros-industrial/industrial_ci/pull/137

gavanderhoorn:
Ok, nice.

Would you want to keep this one open then, and close after implementation / merge of #137, or close right now as a sort-of duplicate?

ipa-mdl:
> Would you want to keep this one open 

Let's keep it open as a reminder

> and I'd rather ignore the IRCv2 packages completely in the Travis runs for motoman_experimental.

As a temporary fix you could create a `CATKIN_IGNORE` file in `BEFORE_SCRIPT` or run `catkin config ..` to blacklist it

gavanderhoorn:
Yeah, I tried to do that, but forgot that the `.rosinstall` file only gets processed by `travis.sh` itself. So my [line added before `travis.sh`](https://github.com/ros-industrial/motoman_experimental/blob/fa7237c2913f13b60697e53af2cacebd92ffc9f1/.travis.yml#L28) under `script` didn't work as I intended.

ipa-mdl:
> Does BEFORE_SCRIPT only work with actual scripts, or would a single script line (ie: touch ..) also work?

The string gets executed by `bash -c`, so `touch ...` works

130s:
>>    Does BEFORE_SCRIPT only work with actual scripts, or would a single script line (ie: touch ..) also work?
>
> The string gets executed by `bash -c`, so `touch ...` works

If I'm not mistaken, target repo is read-only so you can't `touch` into it (there fore something like https://github.com/ros-industrial/industrial_ci/pull/123 is useful)?

ipa-mdl:
[...]

PR #256, opened 2018-01-05
Fix run_tests verbosity

### LINKED ISSUE #253: `catkin_test_results --verbose` explodes travis log beyond max
with the changes introduced in #251 (related to #118), our Travis jobs began to fail due to exceeding log length maximum (we haven't seen this earlier due to other Travis/Docker issues - see #247 #248)

```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).

The job has been terminated
```

I can ensure that "the test suite is raising the same exception over and over" is **not** true in our case!

To give some numbers:
 - all CI steps from start to build-step produce about 2000 lines of output
 - the test-step (--verbose) produces more than 8000 lines of output....

Is there a way to fix #118 without using `--verbose` during the test-step?

Our `travis.yml` looks pretty standard (and has worked before):
```
sudo: required
dist: trusty
language: generic
notifications:
  email:
    on_success: change
    on_failure: always
env:
  global:
    - ROS_REPO=ros
    - NOT_TEST_INSTALL=true
  matrix:
    - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=file
    - ROS_DISTRO=indigo UPSTREAM_WORKSPACE=file $ROSINSTALL_FILENAME="custom.rosinstall"
    - ROS_DISTRO=kinetic UPSTREAM_WORKSPACE=file
matrix:
  allow_failures:
    - env: ROS_DISTRO=kinetic UPSTREAM_WORKSPACE=file
install:
  - git clone https://github.com/ros-industrial/industrial_ci.git .ci_config
script:
  - .ci_config/travis.sh
#  - ./travis.sh  # Enable this when you have a package-local script 
```

ipa-mdl:
This is my attempt to fix #253, as a replacement for #255 

It reverts #244, but sets `TARGET_PKGS` to all packages to address #232.
To make sure that all downstream packages and their dependencies get built, a dedicated step was added.

In addition `tests` will be built before `run_tests` to build the tests with the build options. 

@miguelprada: please test your bare rosinstall usecase
@ipa-fxm: please test with your setup

### DIFF
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -172,7 +172,7 @@ if [ "$NOT_TEST_BUILD" != "true" ]; then
     ici_time_start catkin_run_tests
 
     if [ "$BUILDER" == catkin ]; then
-        catkin run_tests $OPT_VI --no-status $PKGS_DOWNSTREAM $CATKIN_PARALLEL_TEST_JOBS --make-args $ROS_PARALLEL_TEST_JOBS --
+        catkin run_tests $OPT_VI --no-deps --no-status $PKGS_DOWNSTREAM $CATKIN_PARALLEL_TEST_JOBS --make-args $ROS_PARALLEL_TEST_JOBS --
         if [ "${ROS_DISTRO}" == "hydro" ]; then
             PATH=/usr/local/bin:$PATH  # for installed catkin_test_results
             PYTHONPATH=/usr/local/lib/python2.7/dist-packages:$PYTHONPATH
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -177,9 +177,15 @@ if [ "$NOT_TEST_BUILD" != "true" ]; then
     fi
     ici_time_end  # catkin_build_downstream_pkgs
 
+    ici_time_start catkin_build_tests
+    if [ "$BUILDER" == catkin ]; then
+        catkin build --no-deps --catkin-make-args tests -- $OPT_VI --summarize  --no-status $PKGS_DOWNSTREAM $CATKIN_PARALLEL_JOBS --make-args $ROS_PARALLEL_JOBS --
+    fi
+    ici_time_end  # catkin_build_tests
+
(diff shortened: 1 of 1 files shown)


miguelprada:
The bare rosinstall usecase still works. Thanks for keeping it in mind, @ipa-mdl!

ipa-mdl:
@130s: do you have any objections?

ipa-mdl:
`TARGET_PKGS` won't get touched if it was provided by the user.

Old behavior:
Fill it with the the names of the packages in target repo.
For the special case of #232, it was empty.
For `catkin build` this was okay, because it just build all packages.
But `catin run_tests` because `--no-deps` requires listing the package names.

New behavior:
1. Fill it with the the names of the packages in target repo.
2. If this empty, fill it with all package names.

So `catkin build` will still build all, but in addition `catkin run_tests`  will run all tests.

The only side effect I can think of is that users with empty target repos cannot blacklist packages in `CATKIN_CONFIG` anymore (@miguelprada: so, beware! ^^).
But this is true for regular users as well (with or without this PR).
A proper solution would be #172

industrial_ci/src/tests/source_tests.sh
```diff
@@ -163,16 +163,29 @@ ici_time_start catkin_build
 
 # for catkin
 if [ "${TARGET_PKGS// }" == "" ]; then export TARGET_PKGS=`catkin_topological_order ${TARGET_REPO_PATH} --only-names`; fi
+# fall-back to all packages
+if [ "${TARGET_PKGS// }" == "" ]; then export TARGET_PKGS=`catkin_topological_order $CATKIN_WORKSPACE/src --only-names`; fi
```
130s: So this line and previous line is the "new behavior (regarding `TARGET_PKGS`)" in https://github.com/ros-industrial/industrial_ci/pull/256#issuecomment-356976994 . I'd never figure out by myself why these very similar 2 lines are needed. For future maintenance, would you add a comment? Just add a link is fine IMO.

ipa-mdl: Addressed.

PR #266, opened 2018-01-15
Fix issue where catkin_lint checks industrial_ci itself.

130s:
I'm getting an error. All the packages in the target repo seem to pass `catkin_lit`, but CI job still fails with the error below. I think this happens when `catkin_lint` checks `industrial_ci` itself, as it's lying under the same directory on CI:
```
image: docker:git
variables:
  ADDITIONAL_DEBS: "build-essential curl python-pip ssh"
  CATKIN_CONFIG: "-DCMAKE_BUILD_TYPE=Release"
    :
before_script:
  - apk add --update bash coreutils tar  # install industrial_ci depedencies
  - git clone https://github.com/ros-industrial/industrial_ci.git .ci_config
:
```

```
:
You should consider upgrading via the 'pip install --upgrade pip' command.
++ '[' pedantic == pedantic ']'
++ CATKIN_LINT_ARGS=' --strict -W2'
++ catkin_lint --explain --strict -W2 /root/src/pkga_msgs
catkin_lint: downloading kinetic package index from https://raw.githubusercontent.com/ros/rosdistro/master/index.yaml
catkin_lint: downloading package manifest for 'geometry_msgs'
catkin_lint: downloading package manifest for 'message_generation'
catkin_lint: downloading package manifest for 'roscpp'
catkin_lint: downloading package manifest for 'sensor_msgs'
catkin_lint: downloading package manifest for 'message_runtime'
industrial_ci: error: file '_wrap_test.sh' is executable but not installed
     *         Your package contains a file that is marked as
     * executable but not         installed. If it is a script
       * intended to be run (e.g. with rosrun), it         will not work
       * outside the devel tree. If it is not an executable
         * script, you should fix the file permissions.
      * You can ignore this problem with --ignore uninstalled_script
      industrial_ci: error: package description starts with boilerplate 'This package contains'
      *         Your package description starts with a number of
   * typical filler words which         do not actually describe the
     * contents of your package. Typically, you can         simply
     * delete these words from the description, and it will still make
       * sense         and be much more concise.
       * You can ignore this problem with --ignore description_boilerplate
    catkin_lint: checked 2 packages and found 2 problems
    ++ error 'catkin_lint failed by either/both errors and/or warnings'
    ++ local exit_code=1
    ++ '[' -n 'catkin_lint failed by either/both errors and/or warnings' ']'
    ++ echo -e '\e[31mcatkin_lint failed by either/both errors and/or warnings\e[0m'
    catkin_lint failed by either/both errors and/or warnings
    :
    ```


ipa-mdl:
This works on travis: https://travis-ci.org/ros-industrial/ros_canopen/jobs/326344040.
Gitlab example: https://gitlab.com/ipa-mdl/ci-example/-/jobs/48066347 fails, but `industrial_ci` is not listed.
Not sure what's going on with your build.

ipa-mdl:
Have you tried `run_ci`?

ipa-mdl:
I think this is some bug in the `catkin_lint` package crawler.

130s:
The issue this PR tries to tackle makes lint option unusable in some situation. Hope I can continue on this on ROS-I day 2018.

ipa-mdl:
I guess, I found the problem: https://github.com/fkie/catkin_lint/pull/57
This issue is related to https://github.com/ros-industrial/industrial_ci/issues/186 ;)

130s:
Indeed the issue raised in this PR has not been happening to the same repo for 10 months. Thanks @ipa-mdl for fixing this in the upstream!

PR #278, opened 2018-05-15
[doc] Seasonal update metric May, 2018

130s:
> - `Searching Github repos that contain string industrial_ci <https://github.com/search?p=1&q=industrial_ci+-repo%3Aros-industrial%2Findustrial_ci&ref=searchresults&type=Code&utf8=%E2%9C%93>`_) (with some duplicates. Excluding industrial_ci repo):
> 
> +  - 675 (May 15, 2018)
>     - 457 (Dec 12, 2017) 
>     - 142 (Jan 20, 2017) 

Not sure what this search results actually means, but this exact same search command has been used for an year and half, and the resulted number has grown steadily.


miguelprada:
No need for much review, right? I would suggest to merge it already.

Nice numbers, by the way! :smile:

ipa-mdl:
What do you think of moving the screenshots to a dedicated traffic.rst and showing only the most recent in the main Readme?

ipa-mdl:
Another aspects: In the long-run would we embed a lot of binary files that will slow down the git clone eventually. Can we store the images somewhere else?

130s:
> What do you think of moving the screenshots to a dedicated traffic.rst and showing only the most recent in the main Readme?

Well, screenshots can be an eye candy :) On serious note, screenshots are the facts and proven result for our effort. So I hope particularly new comers find it helpful, thus I suggest keep them at the top page.

> Another aspects: In the long-run would we embed a lot of binary files that will slow down the git clone eventually. Can we store the images somewhere else?

Never thought about that, but I'm in. Where can we store them with no cost and have multiple persons access to? Does GitHub offers something like that?
- Looks like [`gist` allows uploading image files](https://gist.github.com/remarkablemark/feff40b0a522f0c41c4eff0b77ea1d47), but I'm not sure if gist is enabled for industrial_ci repo (sorry I've never used gist).
- [GitHub Wiki for the repo](https://github.com/ros-industrial/industrial_ci/wiki) doesn't seem to accept uploading files ([readme](https://help.github.com/articles/adding-images-to-wikis/)).
- In the worst case I can create a branch for storing img files, and have readme reference there...

ipa-mdl:
> Well, screenshots can be an eye candy :) On serio

That's why I would keep the latest one. We could even plot the data in combined image.

> In the worst case I can create a branch for storing img files, and have readme reference there...

As long as users clone with `--depth 1`  we could store images in another branch.
That's why I proposed `gh-pages`. We coulld even generate a noce metrics page.

130s:
Ok, I moved metric image files to `gh-pages` branch.

This does keep `master` from non-CI functional stuff, but it does NOT necessarily keep the repo from getting bigger as more images are added. Metric images shouldn't grow that big hopefully, and I do prefer keeping every resource at one place as much.

> As long as users clone with `--depth 1` we could store images in another branch.

Won't be easy to change the way existing repos clone `industrial_ci` I assume, but nice to know.

PR #282, opened 2018-05-30
add upgrade pip to default dockerfile

fmessmer:
This PR fixes the following situation:

I recently added a new rosdep key for a pip python package: https://github.com/ros/rosdistro/pull/17952

This package has first been added to pip on March 4, 2017 (https://pypi.org/project/face_recognition/#history)
`industrial_ci` uses pip with the version 8.1.1 which was released on March 17, 2016 (https://pypi.org/project/pip/#history)
Thus, for pip 8.1.1 the package is not known.

During a CI run, I see the following output
```
You are using pip version 8.1.1, however version 10.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
ERROR: the following rosdeps failed to install
  pip: Failed to detect successful installation of [face_recognition]
```

Would upgrading pip be acceptable?
Still, I don't understand, why `sudo apt-get install python-pip` installs such an old version rather than the latest version in the first place...

### DIFF
--- a/industrial_ci/src/docker.sh
+++ b/industrial_ci/src/docker.sh
@@ -234,6 +234,7 @@ RUN sed -i "/^# deb.*multiverse/ s/^# //" /etc/apt/sources.list \
         python-wstool \
         ros-$ROS_DISTRO-catkin \
         ssh-client \
+    && pip install --upgrade pip \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 EOF


ipa-mdl:
> Still, I don't understand, why sudo apt-get install python-pip installs such an old version rather than the latest version in the first place...

Standard (conservative) debian packaging rules.

fmessmer:
> Maybe your dependency uses some features that is not present in the old pip version.

Yes, as described above `face_recognition` wasn't released by the time `pip 8.1.1` came out

ipa-mdl:
> Yes, as described above face_recognition wasn't released by the time pip 8.1.1 came out

This is not what I meant. pip is like apt.

fmessmer:
well, my package is python and simply imports (the pip) face_recognition...beyond installing the dependency I don't do much unittesting or anything...
also, the error message says:
```
ERROR: the following rosdeps failed to install
  pip: Failed to detect successful installation of [face_recognition]
```
so the problem clearly is `rosdep install` of the pip package `face_recognition`

ipa-mdl:
The docker images do not contain the recommended depedencies for pip.

ipa-mdl:
> well, my package is python and simply imports (the pip) face_recognition...beyond installing the dependency I don't do much unittesting or anything...

A dependency of `face_recognition` uses C extension and must be compiled. This might fail.
Please try to find out which dependency is missing to fix the build by setting `ADDITIONAL_DEBS`.

ipa-mdl:
In my test `python-dev` seems to do the trick, so please try to include it in your PR.

ipa-mdl:
> is this a branch of yours?
cant find the test you mentioned

Just a local test with docker.

fmessmer:
sorry, i dont get what you did/mean

ipa-mdl:
> sorry, i dont get what you did/mean

Remove the upgrade line, but add 'python-dev` to the list of installed packages.
Finally, try to test your repository again.

fmessmer:
haha :man_facepalming: 
I clearly havent recovered from my fever yet
will test your hint, thanks!

fmessmer:
@ipa-mdl how did you test?
for me, the current version of this PR, i.e. no upgrade pip but added python-dev, brings back the error:
```
You are using pip version 8.1.1, however version 10.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
ERROR: the following rosdeps failed to install
  pip: Failed to detect successful installation of [face_recognition]
```

ipa-mdl:
`python-wheel` seems to be missing as well.
Might be an issue with rosdep as well, installation succeed in my tests, but it cannot be detected.

ipa-mdl:
> Might be an issue with rosdep as well, installation succeed in my tests, but it cannot be detected.

Actually, the `face_recognition`  is the wrong package name, PyPi uses (and reports) dashes.

ipa-mdl:
My commit should address the build errors.
`pip install -U pip`  breaks pip for ubuntu 18.04.

ipa-mdl:
I have opened an issue in rosdistro..

fmessmer:
@ipa-mdl thank you for your help in getting this fixed!

PR #279, opened 2018-05-20
[WIP] rerun_ci script

ipa-mdl:
`rerun_ci` runs local tests like `run_ci`, but speeds up recurring builds by reusing the  image from the last run.

Each run will commit the container to `industrial-ci/run_ci/$REPO_NAME:$CLI_HASH`. The latter hashes the command line arguments. Thee will added to the images as a comment as well.
So each combination of (absolute) target path and enviromnent settings gets a dedicated image.

There are some differences between `run_ci` and `rerun_ci`:
* the first argument must be the target path, "." is okay as well.
* the images will persist, the users must delete them manually
* the environement gets filled only with the command line arguments

Please give it a try.

PR #283, opened 2018-06-02
Add DOCKER_PULL option

ipa-mdl:
This PR adds a flag to disable automatic pull of images.
Setting it to false assumes that the (base) image was already pulled or provided otherwise.

Cherry-picked from #279 and extended to include `DOCKER_BASE_IMAGE`.
I will update #279 and #280 afterwards.

The test jobs are run on gitlab only: https://gitlab.com/ipa-mdl/industrial_ci/pipelines/23079479

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -187,9 +187,10 @@ Note that some of these currently tied only to a single option, but we still lea
 * **CCACHE_DIR** (default: not set): If set, `ccache <https://en.wikipedia.org/wiki/Ccache>`_ gets enabled for your build to speed up the subsequent builds in the same job if anything. See `detail. <https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst#cache-build-artifacts-to-speed-up-the-subsequent-builds-if-any>`_
 * **DEBUG_BASH** (default: not set): If set with any value (e.g. `true`), all executed commands that are not printed by default to reduce print space will be printed.
 * **DOCKER_BASE_IMAGE** (default: $OS_NAME:$OS_CODE_NAME): Base image used for building the CI image. Could be used to pre-bundle dependecies or to run tests for different architectures. See `this PR <https://github.com/ros-industrial/industrial_ci/pull/174>`_ for more info.
-* **DOCKER_IMAGE** (default: not set): Selects a Docker images different from default one. Please note, this disables the handling of `ROS_REPOSITORY_PATH` and `ROS_DISTRO` as ROS needs already to be installed in the image.
-* **DOCKER_FILE** (default: not set): Instead of pulling an images from the Docker hub, build it from the given path or URL. Please note, this disables the handling of `ROS_REPOSITORY_PATH` and `ROS_DISTRO`, they have to be set in the build file instead.
 * **DOCKER_BUILD_OPTS** (default: not set): Used do specify additional build options for Docker.
+* **DOCKER_FILE** (default: not set): Instead of pulling an images from the Docker hub, build it from the given path or URL. Please note, this disables the handling of `ROS_REPOSITORY_PATH` and `ROS_DISTRO`, they have to be set in the build file instead.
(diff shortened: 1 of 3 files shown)


ipa-mdl:
@miguelprada: Do you have time for a review?

130s:
I can get to review this in a few hours hopefully

doc/index.rst
```diff
@@ -187,9 +187,10 @@ Note that some of these currently tied only to a single option, but we still lea
...
-* **DOCKER_IMAGE** (default: not set): Selects a Docker images different from default one. Please note, this disables the handling of `ROS_REPOSITORY_PATH` and `ROS_DISTRO` as ROS needs already to be installed in the image.
-* **DOCKER_FILE** (default: not set): Instead of pulling an images from the Docker hub, build it from the given path or URL. Please note, this disables the handling of `ROS_REPOSITORY_PATH` and `ROS_DISTRO`, they have to be set in the build file instead.
 * **DOCKER_BUILD_OPTS** (default: not set): Used do specify additional build options for Docker.
+* **DOCKER_FILE** (default: not set): Instead of pulling an images from the Docker hub, build it from the given path or URL. Please note, this disables the handling of `ROS_REPOSITORY_PATH` and `ROS_DISTRO`, they have to be set in the build file instead.
+* **DOCKER_IMAGE** (default: not set): Selects a Docker images different from default one. Please note, this disables the handling of `ROS_REPOSITORY_PATH` and `ROS_DISTRO` as ROS needs already to be installed in the image.
+* **DOCKER_PULL** (default: true): set to false if custom docker image should not be pulled, e.g. if it was created locally
```
130s: If `false`, shouldn't some other `DOCKER*` variable need to be passed? If that is correct, let's add that as a note (if I'm wrong, sorry in advance).

industrial_ci/src/docker.sh
```diff
@@ -135,6 +135,9 @@ function docker_cp {
 #######################################
 function ici_docker_build() {
   local opts=($DOCKER_BUILD_OPTS)
+  if [ "$DOCKER_PULL" != false ]; then
```
130s: By mistake users can pass `DOCKER_PULL` with invalid values (similar to #296). So it'd be nice for this `if` logic to be strict.

industrial_ci/src/docker.sh
```diff
@@ -166,7 +169,7 @@ function ici_prepare_docker_image() {
     fi
   elif [ -z "$DOCKER_IMAGE" ]; then # image was not provided, use default
      ici_build_default_docker_image
-  else
+  elif [ "$DOCKER_PULL" != false ]; then
```
130s: Related to https://github.com/ros-industrial/industrial_ci/pull/283/files#r201791665, it'd be friendlier if error prints what could be missing when necessary `DOCKER*` variables aren't passed. If `docker` already prints something obvious, that may be good enough though.

doc/index.rst
ipa-mdl: These have to be passed anyway. It just adds the option to not pull the image before the run.
This can be used to create the image in another Gitlab stage, which then can  be reused in other jobs.

industrial_ci/src/docker.sh
ipa-mdl: For now I don't want  to check these cases, because we don't have the proper infrastructure (something like #285)

doc/index.rst
130s: Yeah, but what's the necessary variable set might not be obvious for esp. users new to this feature. Maybe adding a small sample section is a way to go (I think the tests you're adding in this PR can be used there).

ipa-mdl: Actually, I have added this just for #279. Most users don' t have to deal with this.

130s: I can open another PR for doc then.

industrial_ci/src/docker.sh
130s: Agreed. That is better than repeating to embed same/similar if-else logic.

industrial_ci/src/docker.sh
130s: I guess this follows the decision made in https://github.com/ros-industrial/industrial_ci/pull/283#discussion_r201791665

PR #286, opened 2018-06-06
Feature: cache images

ipa-pgt:


ipa-mdl:
I have implemented a similar feature in #279 
I stumbled upon the same docker pull issue, which gets addressed properly in #283 

What do you think of consolidating both?

industrial_ci/src/docker.sh
```diff
@@ -166,7 +166,7 @@ function ici_prepare_docker_image() {
   ici_time_start prepare_docker_image
 
   if [ -n "$DOCKER_FILE" ]; then # docker file was provided
-    DOCKER_IMAGE=${DOCKER_IMAGE:"industrial-ci/custom"}
+    DOCKER_IMAGE=${DOCKER_IMAGE:-"industrial-ci/custom"}
```
ipa-mdl: :+1:

ipa-pgt:
Sorry, I didn't look through the open pull requests before opening mine.
I think both commits get addressed by the two pull requests. I also prefer your variant with the extra DOCKER_PULL variable.
However, the _COMMIT_IMAGE variable should be and stay accessible from GitLab CI via gitlab.sh, because we have 2 builds per ROS_DISTRO in the same pipeline. I want to use the cached docker image so that the dependencies don't have to be installed again for the 2nd build.

ipa-mdl:
I think `DOCKER_PULL` needs an "auto" option, to pull images if they do not exists (like in your code).
I can add this in #283. As an alternatove, you could just add the pull instruction in your pipeline (like I in the new gitlab tests).

Your PR should be rebased on top.
However, I would rename `KEEP_IMAGE_AFTER_RUN` to something else because it implies boolean options. `DOCKER_COMMIT`?

> However, the _COMMIT_IMAGE variable should be and stay accessible from GitLab CI via gitlab.sh

It is unset for the running bash instance only. I must admit, that I don't remember anymore why I added this.

ipa-pgt:
> I think DOCKER_PULL needs an "auto" option, to pull images if they do not exists (like in your code).
I can add this in #283

This could be a good default value instead of _true_, even though _true_ and _false_ should already be sufficient.

> Your PR should be rebased on top.
However, I would rename KEEP_IMAGE_AFTER_RUN to something else because it implies boolean options. DOCKER_COMMIT?
> > However, the _COMMIT_IMAGE variable should be and stay accessible from GitLab CI via gitlab.sh
>
> It is unset for the running bash instance only. I must admit, that I don't remember anymore why I added this.

_COMMIT_IMAGE could be included as a "public" variable (COMMIT_IMAGE) of industrial_ci, if unsetting it is not necessary. But I could also assign _COMMIT_IMAGE from another variable, so my commit is not necessary.

ipa-mdl:
Superseded by #341

PR #287, opened 2018-06-08
remove rosdep install output filter

ipa-mdl:
Bug in rosdep has been fixed (https://github.com/ros-infrastructure/rosdep/pull/612).
Should get merged after the next release.

### DIFF
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -152,9 +152,7 @@ rosdep_opts=(-q --from-paths $CATKIN_WORKSPACE/src --ignore-src --rosdistro $ROS
 if [ -n "$ROSDEP_SKIP_KEYS" ]; then
   rosdep_opts+=(--skip-keys "$ROSDEP_SKIP_KEYS")
 fi
-set -o pipefail # fail if rosdep install fails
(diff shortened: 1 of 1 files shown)


130s:
The same jobs seems to have kept failing with args `ROS_DISTRO=kinetic OS_NAME=debian OS_CODE_NAME=jessie AFTER_SCRIPT='grep -q ID=debian /etc/os-release && grep -q VERSION_ID=\"8\" /etc/os-release'`.

https://travis-ci.org/ros-industrial/industrial_ci/jobs/389602235#L559
```
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
industrial_ci: No definition of [catkin] for OS [debian]
Function rosdep_install returned with code '1' after 0 min 0 sec
```

ipa-mdl:
rebased to current master to include https://github.com/ros-industrial/industrial_ci/pull/292, this should resolve the error.

130s:
>  Many bigger repositories might fail because of the 10k limit of travis.

Thanks for catching that.

> This was meant as a warning to not merge this.

I clearly didn't get that. In the future, would you mind adding WIP when a PR isn't ready for review/merge (or do something clear enough so others can tell the status)?

PR #290, opened 2018-06-16
WIP: Modularize gitlab codes.

130s:
The way Gitlab CI requires on ssh key handling can be the same on every Gitlab repo, so it'd be nice if industrial_ci allows repos to just call that function, instead of running the entire CI process.

For example, I have a CI job that only builds Docker image where build and test steps (in ROS) are not needed. It needs to access other private Gitlab repos. industrial_ci is already pulled in in the same .gitlab-ci.yml for other jobs that use industrial_ci in a "normal" way, so it'd be ideal for the building docker job to be able to use the ssh key function.

There's problem I haven't got this working yet, hence WIP. Gitlab CI (`image: docker:git`) returns error when sourcing the split .sh file (that looked like the one `sh` tried to read `bash`, but I don't have an output now).

```
$ . industrial_ci/src/tests/gitlab_module.bash
sh: 25: industrial_ci/src/tests/gitlab_module.bash: function: not found
sh: 36: industrial_ci/src/tests/gitlab_module.bash: Syntax error: "(" unexpected
```
The following works apparently but bash terminates so can't inherit anything I'm afraid.
```
$ bash industrial_ci/src/tests/gitlab_module.bash
$ 
```

### DIFF
--- a/gitlab.sh
+++ b/gitlab.sh
@@ -21,26 +21,11 @@
 # 2016/05/18 http://stackoverflow.com/questions/59895/can-a-bash-script-tell-what-directory-its-stored-in
 DIR_THIS="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
+source industrial_ci/src/tests/gitlab_module.bash
 export TARGET_REPO_PATH=$CI_PROJECT_DIR
 export TARGET_REPO_NAME=$CI_PROJECT_NAME
 export _DO_NOT_FOLD=true
 
-if [ -n "$SSH_PRIVATE_KEY" ]; then
-  if [ "$CI_DISPOSABLE_ENVIRONMENT" != true ] && ! [ -f /.dockerenv ] ; then
-    echo "SSH auto set-up cannot be used in non-disposable environments"
-    exit 1
-  fi
-
-  # start SSH agent
-  eval $(ssh-agent -s)
-  # add key to agent
-  ssh-add <(echo "$SSH_PRIVATE_KEY") || { res=$?; echo "could not add ssh key"; exit $res; }
-
-  if [ -n "$SSH_SERVER_HOSTKEYS" ]; then
-    mkdir -p ~/.ssh
-    # setup known hosts
-    echo "$SSH_SERVER_HOSTKEYS" > ~/.ssh/known_hosts
-  fi
-fi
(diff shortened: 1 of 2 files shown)


130s:
Thanks for the advice! I've been trying to improve this PR as this is getting more in need, although I haven't got it working yet.

I can think of 2 approaches:

- (A) Get `bash` working on `docker:git` -> Having an issue ([asked on SO](https://stackoverflow.com/questions/51378433/bash-starts-bin-sh-on-dockergit)).
- (B) Port the sh file in question to `sh`.

Feel like (A) is better as (B) basically throws out a well-tested module, and there should be a way to get (A) work.

ipa-mdl:
(B) is really hard, because we are using quite a lot of `bash`  features.
(A) is [easy](https://github.com/ros-industrial/industrial_ci/blob/master/.gitlab-ci.yml#L17).
(your issue is about `source`, which is `bash` specific. For `sh` you should use `.`)

(C) prepare image `rosindustrial/ci:gitlab` (docker:git + bash coreutils tar).

130s:
> (B) is really hard, because we are using quite a lot of bash features.
> For some files, it should not make a difference at all :)

Well for now I'm only thinking of `gitlab.sh`. It seems to me the first line failing is the one below and nothing complicated after that line, so I'll see if I can port that to `sh`. 

```
 -  ssh-add <(echo "$SSH_PRIVATE_KEY") || { res=$?; echo "could not add ssh key"; exit $res; }
```

> (A) is [easy](https://github.com/ros-industrial/industrial_ci/blob/master/.gitlab-ci.yml#L17).
> (your issue is about `source`, which is `bash` specific. For `sh` you should use `.`)

Are you able to keep using `bash` on Gitlab CI? For some reason I can't, hence asked [the SO question](https://stackoverflow.com/questions/51378433/bash-starts-bin-sh-on-dockergit).

ipa-mdl:
> Are you able to keep using bash on Gitlab CI? For some reason I can't, hence asked the SO question.

Running a bash script (with shebang) works fine.
However, you try to source a bash script in sh. This fails for two reasons:
1. `source`  is not available, so use `.`
2. You cannot source a script with bash features into a running `sh`  instance

ipa-mdl:
@130s: Are you still working on this?

Issue #293, opened 2018-07-01
[run_travis] "TypeError: sequence item 1: expected string, dict found"


130s:
```
$ cd industrial_ci
$ git log -1
commit 1eaffcfb291c1d3d5ebb631c1c07100e0fb4302f
Merge: cba6908 e373080
Author: Isaac I.Y. Saito <130s@users.noreply.github.com>
Date:   Thu Jun 28 19:12:58 2018 -0700

    Merge pull request #292 from ipa-mdl/debian-stretch
    
    test for debian stretch instead of jessie
$ ./industrial_ci/scripts/run_travis -h                                                                            
Traceback (most recent call last):
  File "./industrial_ci/scripts/run_travis", line 174, in <module>
    main(os.path.abspath(os.path.dirname(__file__)), sys.argv)
  File "./industrial_ci/scripts/run_travis", line 143, in main
    global_env, job_envs = read_env(read_yaml(path)['env'])
  File "./industrial_ci/scripts/run_travis", line 91, in read_env
    g = ' '.join(env['global'])
TypeError: sequence item 1: expected string, dict found

$ ll
total 128K
drwxrwxr-x  8 n130s n130s 4.0K Jul  1 03:22 .git/
drwxrwxr-x  5 n130s n130s 4.0K Jul  1 02:48 ./
-rw-rw-r--  1 n130s n130s 6.9K Jul  1 02:48 .travis.yml
drwxrwxr-x  2 n130s n130s 4.0K Jul  1 02:48 doc/
drwxrwxr-x  5 n130s n130s 4.0K Jul  1 02:48 industrial_ci/
-rw-rw-r--  1 n130s n130s 1.2K Jul  1 02:48 .gitlab-ci.yml
-rw-rw-r--  1 n130s n130s 5.3K Jul  1 02:48 README.rst
-rwxrwxr-x  1 n130s n130s 1.6K Jun 22 22:00 gitlab.sh*
-rwxrwxr-x  1 n130s n130s 1.2K Jun 12 19:45 travis.sh*
-rw-rw-r--  1 n130s n130s   83 Jun 12 19:45 .ci.rosinstall.jade
-rwxrwxr-x  1 n130s n130s 1.1K Jun 12 19:45 ci.sh*
-rw-rw-r--  1 n130s n130s  12K Jun 12 19:45 LICENSE
-rw-rw-r--  1 n130s n130s  179 Jun 12 19:45 .travis.rosinstall
[...]

PR #294, opened 2018-07-01
[run_travis] Add '-h' option.

130s:
This fixes https://github.com/ros-industrial/industrial_ci/issues/293 in an adhoc way.

If we want to address #293 in more standard way, we might want to use `argparse` as noted in https://github.com/ros-industrial/industrial_ci/pull/230#pullrequestreview-133449630, but IMO for now this is fine.

### DIFF
--- a/industrial_ci/scripts/run_travis
+++ b/industrial_ci/scripts/run_travis
@@ -133,7 +133,7 @@ Additional variable names can be passed at the end.
 """ % cmd)
 
 def main(scripts_dir, argv):
-    if '--help' in argv:
+    if ('--help' in argv or '-h' in argv):
         print_help(argv[0])
         sys.exit(0)
 
--- a/industrial_ci/scripts/run_travis
+++ b/industrial_ci/scripts/run_travis
@@ -140,6 +140,11 @@ def main(scripts_dir, argv):
     args, extra_env = parse_extra_args(argv[1:])
 
     path, args = find_config_file(args, ['.travis.yml', '.travis.yaml'])
+    if path is None:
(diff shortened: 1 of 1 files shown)


ipa-mdl:
This does not fix #293 properly. Same would happen with `-v` etc.
We could switch to `argparse` or just check path. I will add a commit for the latter.

130s:
Hmm.

```
$ ./industrial_ci/scripts/run_travis -v
Traceback (most recent call last):
  File "./industrial_ci/scripts/run_travis", line 179, in <module>
    main(os.path.abspath(os.path.dirname(__file__)), sys.argv)
  File "./industrial_ci/scripts/run_travis", line 148, in main
    global_env, job_envs = read_env(read_yaml(path)['env'])
  File "./industrial_ci/scripts/run_travis", line 91, in read_env
    g = ' '.join(env['global'])
TypeError: sequence item 1: expected string, dict found
$ git log -1
commit 9f1e3dd57413ce72ec91bb7aa61dd716b41fd555
Author: Mathias Lüdtke <mathias.luedtke@ipa.fraunhofer.de>
Date:   Sun Jul 1 13:53:40 2018 +0200
```
If we should just switch to `argpase` I can do that.

machinekoder:
@ipa-mdl argparse gives the option of forwarding unhandled arguments, so I think it would be possible without breaking the CLI.

machinekoder:
`arguments, unknown = parser.parse_known_args()` does the trick.

130s:
Closing in favor of #300

Issue #296, opened 2018-07-09
Defined but empty  'DOCKER_IMAGE' does not cause error.


130s:
When passing empty `DOCKER_IMAGE` variable, CI continues although the intended docker image wasn't pulled. This makes debugging harder.

ipa-mdl:
Can you give more context, please?
`DOCKER_IMAGE`  has multiple purposes, it might be used differently with regard to other settings.
In addition it would be good to know what the error looks like.

ipa-mdl:
I see.
This case should be fixed [here](https://github.com/ros-industrial/industrial_ci/blob/d66bf0f1578008c23f0673ff8894e20308f067c0/industrial_ci/src/docker.sh#L167), because it just tests the length..

machinekoder:
Working on this for #wrid18

130s:
Maybe related to #212 btw.

ipa-mdl:
#212 is not related.

130s:
> #212 is not related.

You're right (I'm blaming my bad view on github from cellphone :/ ).

ipa-mdl:
Reviewing this issue, I would now opt for "wont fix".
`DOCKER_IMAGE=''` was not handled as error, because it means "use the default".
If we check this, then we have to a lot more variables.

PR #297, opened 2018-07-09
WIP: [fix] defined but empty 'DOCKER_IMAGE' does not cause error (#296)

130s:
**Issue**: See #296.

**Approach**: Let CI fail when `DOCKER_IMAGE` is defined but empty. Ref. https://stackoverflow.com/a/16753536/577001

I'm testing on my private repo but not working so far.

### DIFF
--- a/industrial_ci/src/docker.sh
+++ b/industrial_ci/src/docker.sh
@@ -156,7 +156,7 @@ function ici_prepare_docker_image() {
   ici_time_start prepare_docker_image
 
   if [ -n "$DOCKER_FILE" ]; then # docker file was provided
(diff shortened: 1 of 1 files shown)


ipa-mdl:
> I'm testing on my private repo but not working so far.

I think you have patched the wrong line.

ipa-mdl:
> Approach: Let CI fail when DOCKER_IMAGE is defined but empty

Why do you want it to fail?
In this particular mode the image name does not really matter, we just need one for technical reasons.
For now I'd prefer [this](https://github.com/ros-industrial/industrial_ci/pull/286/commits/4ef6ac5dfbb780d17c9d4373e348b3313dd6a06e).

ipa-mdl:
superseded  by #299

PR #299, opened 2018-07-11
Error on defined but empty DOCKER_IMAGE

machinekoder:
Check if DOCKER_IMAGE environment variable was set.
If it is set and empty, throw an error.

See https://github.com/ros-industrial/industrial_ci/issues/296
Supersedes #297

### DIFF
--- a/industrial_ci/src/docker.sh
+++ b/industrial_ci/src/docker.sh
@@ -176,9 +176,7 @@ function ici_prepare_docker_image() {
   elif [ -z "${DOCKER_IMAGE+x}" ]; then # image was not provided, use default
      ici_build_default_docker_image
   elif [ -z "$DOCKER_IMAGE" ]; then
-     error "Empty string passed to DOCKER_IMAGE." \
-     "Specify a valid docker image or unset the environment variable to use the default image." 1>&2
-     exit 1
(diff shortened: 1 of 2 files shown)


machinekoder:
@ipa-mdl What's the best way to test this? I don't understand how `source_tests.sh` works.

ipa-mdl:
I have just added the test :)

ipa-mdl:
The tests passed, but only the first line is colored. I would keep it this way.

industrial_ci/src/docker.sh
```diff
@@ -164,8 +164,10 @@ function ici_prepare_docker_image() {
     else # url, run directly
         ici_docker_build "$DOCKER_FILE" > /dev/null
     fi
-  elif [ -z "$DOCKER_IMAGE" ]; then # image was not provided, use default
+  elif [ -z ${DOCKER_IMAGE+x} ]; then # image was not provided, use default
```
ipa-mdl: Why did you remove the quotes?

130s:
I took advantage of maintainership and pushed a few commits to @machinekoder's branch to address some comments.

ipa-mdl:
I think it is ready for a squash(!)-merge, but I am concerned that this might break some test set-ups.
There is no way to unset a global variable in Travis :-/
Until now it was sufficient to just empty the variable for specific jobs.

ipa-mdl:
supersed by #617

PR #300, opened 2018-07-11
Fix for #293 and rework of #294

machinekoder:
See https://github.com/ros-industrial/industrial_ci/pull/294 and https://github.com/ros-industrial/industrial_ci/issues/293#issuecomment-404210938

This one uses argparse to filter out the -h argument. Argparse can also be used to add new keyword arguments in the future.

### DIFF
--- a/industrial_ci/scripts/run_travis
+++ b/industrial_ci/scripts/run_travis
@@ -133,7 +133,7 @@ Additional variable names can be passed at the end.
 """ % cmd)
 
 def main(scripts_dir, argv):
-    if '--help' in argv:
+    if ('--help' in argv or '-h' in argv):
         print_help(argv[0])
         sys.exit(0)
 
--- a/industrial_ci/scripts/run_travis
+++ b/industrial_ci/scripts/run_travis
@@ -18,6 +18,7 @@
 
 from __future__ import print_function
 
+import argparse
 import os
 import os.path
 import re
@@ -105,9 +106,8 @@ def gen_env(e):
     return '%s=%s' % (e, os.getenv(e, ''))
 
 
-def print_help(cmd):
-    print("""
-Usage: %s [PATH] [RANGE*] [-- [ENV[=VALUE]*]]
+_help_message = """
+%(prog)s [PATH] [RANGE*] [-- [ENV[=VALUE]*]]
 
 Parses the travis config in the given path or in the current working directory and runs all specified tests sequentially
 If no range is given, the list of jobs will get printed.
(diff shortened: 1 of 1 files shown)


ipa-mdl:
For everything except `-h`  and `--help` I get
```
Traceback (most recent call last):
  File "industrial_ci/scripts/run_travis", line 177, in <module>
    main(os.path.abspath(os.path.dirname(__file__)), sys.argv)
  File "industrial_ci/scripts/run_travis", line 146, in main
    global_env, job_envs = read_env(read_yaml(path)['env'])
  File "industrial_ci/scripts/run_travis", line 83, in read_yaml
    return yaml.safe_load(f)
  File "/usr/lib/python2.7/dist-packages/yaml/__init__.py", line 93, in safe_load
    return load(stream, SafeLoader)
  File "/usr/lib/python2.7/dist-packages/yaml/__init__.py", line 71, in load
    return loader.get_single_data()
  File "/usr/lib/python2.7/dist-packages/yaml/constructor.py", line 37, in get_single_data
    node = self.get_single_node()
  File "/usr/lib/python2.7/dist-packages/yaml/composer.py", line 39, in get_single_node
    if not self.check_event(StreamEndEvent):
  File "/usr/lib/python2.7/dist-packages/yaml/parser.py", line 98, in check_event
    self.current_event = self.state()
  File "/usr/lib/python2.7/dist-packages/yaml/parser.py", line 174, in parse_document_start
    self.peek_token().start_mark)
yaml.parser.ParserError: expected '<document start>', but found '<block mapping start>'
  in "industrial_ci/industrial_ci/scripts/run_travis", line 31, column 1
```

industrial_ci/scripts/run_travis
```diff
@@ -130,16 +130,19 @@ The jobs will be run in clean environments.
...
 
-""" % cmd)
+"""
 
 def main(scripts_dir, argv):
-    if '--help' in argv:
-        print_help(argv[0])
-        sys.exit(0)
-
-    args, extra_env = parse_extra_args(argv[1:])
```
130s: Since we're dropping this `argv[1:]`, the 0th argument by default is the script's path itself, isn't it?

At least changing [this line](https://github.com/ros-industrial/industrial_ci/pull/300/files#diff-a4313c4772034419fabd7360eba4e6b3R40) to take 1st arg yields different result for me.

ipa-mdl: Yes, this is needed.

ipa-mdl:
Switching to argparse does not help that much..
We could change the CLI, but I am not sure if this is really worth the effort.

Issue #298, opened 2018-07-11
Tests in "test_depend"-ed pkgs shouldn't run?


130s:
I haven't found a way to specify the packages that are `test_depend`-ed and need to be built, but don't want to run the tests of.

What I tried (with a risk that this may be wrong) is to clone test_depended pkgs via .rosinstall, add them in `BUILD_PKGS_WHITELIST` so that they get built.
With this, I see tests in `test_depend`ed packages are run in `catkin_run_tests`, which I do not want.

This might be the scope in https://github.com/ros-industrial/industrial_ci/pull/137, but that looks to be aiming larger than just this, beside it is unmerged yet.

ipa-mdl:
Can you give an example?
`test_depend`-ed packages will to be built if needed, but, tests won't be run.

130s:
I don't have a public example for now, so only verbally tracks the logic.

1. Pkgs in `BUILD_PKGS_WHITELIST` get added into `PKGS_DOWNSTREAM` https://github.com/ros-industrial/industrial_ci/blob/d375a781f6f7ffdfb71e0abb4e4a47e0be309bd1/industrial_ci/src/tests/source_tests.sh#L177
2. Pkgs in `PKGS_DOWNSTREAM` are:
  -  `catkin run_tests`-ed https://github.com/ros-industrial/industrial_ci/blob/d375a781f6f7ffdfb71e0abb4e4a47e0be309bd1/industrial_ci/src/tests/source_tests.sh#L197
  - `rostest`-ed https://github.com/ros-industrial/industrial_ci/blob/d375a781f6f7ffdfb71e0abb4e4a47e0be309bd1/industrial_ci/src/tests/source_tests.sh#L220

I might be misunderstanding something, but due to the lines above, currently the tests in pkgs in `BUILD_PKGS_WHITELIST` are supposed to be run (which I wonder if there's an option to disable).

130s:
Closing as I think the usecase is the same as discussed in https://github.com/ros-industrial/industrial_ci/issues/172 (and by starting to ask about test_depend I might have been creating XY problem).

PR #302, opened 2018-07-12
[doc] Wrong description of 'BUILD_PKGS_WHITELIST'.

130s:
I (finally) found that my confusion in https://github.com/ros-industrial/industrial_ci/issues/298 was based on `BUILD_PKGS_WHITELIST`'s wrong description.


ipa-mdl:
> This can be used to reciprocally specify the packages that you don't want to be built nor tested (those not specified here won't be built nor tested).

This sounds wrong.

1. `catkin build $BUILD_PKGS_WHITELIST` (will build all if `BUILD_PKGS_WHITELIST` is empty)
2. `catkin build $PKGS_DOWNSTREAM ` (external config, `BUILD_PKGS_WHITELIST` or `TARGET_PKGS`)
3. `catkin build --no-deps --catkin-make-args tests $PKGS_DOWNSTREAM` (like 2.)
4. `catkin build --no-deps --catkin-make-args run_tests $PKGS_DOWNSTREAM` (like 2.)
5. ` for pkg in $PKGS_DOWNSTREAM: rostest ...` (like 2.)

So if you specify `BUILD_PKGS_WHITELIST` (and not `PKGS_DOWNSTREAM`), these packages will be built an tested.

I think this logic is very complicated, but not very helpful.
That's why I opened https://github.com/ros-industrial/industrial_ci/issues/106

ipa-mdl:
`BUILD_PKGS_WHITELIST` got removed

PR #303, opened 2018-07-12
Create CONTRIBUTING.md

tfoote:
re: https://github.com/ros-industrial/ros_industrial_issues/issues/26


ipa-mdl:
While the general process is the same, not all points make sense for a CI script system.
@130s, @miguelprada: What do you think?

tfoote:
Feel free to update as you see fit. I was just trying to push the process along re: https://github.com/ros-industrial/ros_industrial_issues/issues/26

PR #305, opened 2018-07-12
add support for bitbucket-pipelines

machinekoder:
This PR adds support for [BitBucket Pipelines](https://confluence.atlassian.com/bitbucket/get-started-with-bitbucket-pipelines-792298921.html).

### DIFF
--- a/bitbucket.sh
+++ b/bitbucket.sh
@@ -2,6 +2,7 @@
 
 # Copyright (c) 2016, Isaac I. Y. Saito
 # Copyright (c) 2017, Mathias Lüdtke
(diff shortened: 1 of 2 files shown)


bitbucket.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+# Copyright (c) 2016, Isaac I. Y. Saito
+# Copyright (c) 2017, Mathias Lüdtke
```
ipa-mdl: Please add your name here as well :)

machinekoder: done

Issue #306, opened 2018-07-13
Cannot skip rosdep step


machinekoder:
In some cases, e.g. when using an already prepared Docker image, we can skip the `rosdep install` step. This is especially useful when our dependencies aren't released yet.

machinekoder:
Alternatively, overwriting the `rosdep` command would also help. Moreover, it may be useful to execute the `wstool init` step before the `rosdep` step, since this might add additional dependencies.

machinekoder:
Ignore my last statement, `wstool` is executed before the `rosdep install` step.

machinekoder:
I think `rosdep` is missing the `--os=debian:stretch` option.

ipa-mdl:
> In some cases, e.g. when using an already prepared Docker image, we can skip the rosdep install step. This is especially useful when our dependencies aren't released yet.

Why do you want to skip rosdep in this case?
In addition to installing them, `rosdep` checks that the dependencies exist and can be resolved.
However, I could imagine to skip `rosdep update`.

> I think rosdep is missing the --os=debian:stretch option.

rosdep auto-detects the OS. You could override the detection by setting `ROS_OS_OVERRIDE`.

machinekoder:
Then the OS autodetection does not work correctly. It looks for `stretch` rather than `debian:stretch`. The rosdep works perfectly fine when passing `--os` manually.

ipa-mdl:
If rosdep fails to detect Debian stretch, this should better be fix in `rosdep`/`rospkg`.
Which error do you get?

machinekoder:
I'm doing something wrong, I did not specify `UPSTREAM_WORKSPACE` to select a file for the `.rosinstall`.

PR #307, opened 2018-07-13
add os argument to rosdep call

machinekoder:
The rosdep call was missing the `--os` argument: https://github.com/ros-industrial/industrial_ci/issues/306


ipa-mdl:
Example: `ROS_DISTRO=kinetic DOCKER_IMAGE=rosindustrial/ci:kinetic-jessie`  would not work anymore. (With #280 you could even leave out `ROS_DISTRO`.)

PR #309, opened 2018-07-17
Remove quiet option from rosdep install

agutenkunst:
Hi! I would like to open the discussion if removing the quiet option from the rosdep install step could reduce the need for `travis_wait`? 

I noticed that I would have timeouts so one recommendation here is using travis_wait but this in return does not produce output (See https://github.com/travis-ci/travis-ci/issues/7323).

Maybe for some of us would this could be a solution for now.

### DIFF
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -148,7 +148,7 @@ fi
 
 ici_time_start rosdep_install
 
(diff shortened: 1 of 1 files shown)


ipa-mdl:
The quiet flag is not used at all (bug in rosdep).
For now the output is silenced  by `| { grep "executing command" || true; }`, which buffers the other output.
As soon as the new rosdep version was released, we can merge #287.

BTW, `travis_wait` works fine with `industrial_ci` .

agutenkunst:
Ok then I will check again what the issue behind my `travis_wait` is. Closing.

PR #310, opened 2018-07-31
Added environment variable for proxy

agutenkunst:
If you are behind a proxy setting this variable will make `rosdep install` work.
Usually on a host system behind the same proxy this variable is already set
and therefore can be substituted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,6 @@ services:
   - docker
 # Force travis to use its minimal image with default Python settings
 language: generic
-notifications:
-  #  Whoever receives success and/or failure via email
-  email:
-    recipients:
-      - gm130s@gmail.com  # Change this to yours should you copy this file.
 
 git:
   quiet: true


agutenkunst:
Ok I notice getting a test for this may be an issue...

ipa-mdl:
I think these variables should be added to docker.env as well.

ipa-mdl:
Please try `DOCKER_BUIILD_OPTS="--build-arg http_proxy=$http_proxy"`  and `DOCKER_RUN_OPTS="-e http_proxy=$http_proxy"`.

If this works, it could be set automatically via in this PR:

ipa-mdl:
> Ok I notice getting a test for this may be an issue...

We could run a local proxy in one test: `http_proxy=http://localhost:3128`

And handle it in `_wrap_test.sh`
```
[ -z "$http_proxy" ] || sudo apt-get install squid
...
[ -z "$http_proxy" ] || [ -s /var/log/squid/access.log ]
```

PR #317, opened 2018-09-06
WIP: Add yaml_lint feature.

130s:
- Motivation: Adding a feature to check the file format for all the yaml files in the repo. It's simple enough to do locally but still involves a few steps on CI, so hope this is a nice addition.
- Verification: Not done yet. I'll run testing.
- Concern:
  - File format checking may not be a core of CI functionality. So I'm not sure if that goes into `source_tests.sh`, which is kind of core of industrial_ci IMO. But then I'm not sure either what should be the 'core' steps of CI.


130s:
@ipa-mdl please review.

I only tested with a private repo. Obviously the design of this feature covers more usecases than that. So far for the test case jobs in this repo I've enabled this feature for only a few jobs (incl. added a couple of new jobs), by which I hope to expand the coverage more.

industrial_ci/src/tests/source_tests.sh
```diff
@@ -168,6 +168,12 @@ if [ "$CATKIN_LINT" == "true" ] || [ "$CATKIN_LINT" == "pedantic" ]; then
     ici_time_end  # catkin_lint
 fi
 
+if [ "$YAML_LINT" == "true" ]; then
```
130s: > - Concern:
>  - File format checking may not be a core of CI functionality. So I'm not sure if that goes into `source_tests.sh`, which is kind of core of industrial_ci IMO. But then I'm not sure either what should be the 'core' steps of CI.

"Core" steps for CI can be those that happen in a chain? E.g. `filling dependency -> build -> install (-> test)` (which is what `rosdep` and `Catkin` chain do).

I'm not sure, but things like file formatting check (which is what this PR adds) can be done independently from that chain, and thus could be run separately from the "main" sequence (source_tests.sh etc.) As there can be more convenient features like this that can run independently, I don't feel comfortable to keep inserting into the main sequence.

Moved back to WIP.

industrial_ci/src/util.sh
```diff
@@ -170,6 +170,50 @@ function ici_enforce_deprecated {
...
+#######################################
+function run_yamllint {
+    target_path_default="$TARGET_REPO_PATH"
+    # If there are '*rosinstall*' files in the repo, add to the search criteria.
+    files_rosinstall=$(find "$TARGET_REPO_PATH" -path ./.git -prune -o -iname "*rosinstall*" -print)
+    if [ $? == 0 ]; then target_path_default="$files_rosinstall $target_path_default"; fi
+
+    yamllint_conf=$1
+    target_paths=${2:-"$target_path_default"}  # multiple elements delimitted by space.
+    sudo apt-get install -qq -y python3-pkg-resources yamllint || (echo "WARN: Required package 'yaml_lint' isn't available. Skipping to check yaml files." && return);
```
ipa-mdl: I this fails, the test should fail.

ipa-mdl:
It would be nice to port this over to the `master` branch, but without bundling *opinionated*  default configs. Instead, the user should pass the config, perhaps with support for URLs.

Issue #318, opened 2018-09-14
Option to access to Docker image built inside industrial_ci


130s:
@ipa-mdl Is there any reason why there is no way to access the docker image built during industrial_ci process?

I see in a recent change there's an option to `docker commit` the built image https://github.com/ros-industrial/industrial_ci/blob/d8ac9340ab52a2fc625a55d558759831a3a79b4b/industrial_ci/src/docker.sh#L114, but the option is not publicly exposed (only available when running `rerun_ci` script AFAIUC). And regardless of which option to start industrial_ci, the built image is deleted https://github.com/ros-industrial/industrial_ci/blob/d8ac9340ab52a2fc625a55d558759831a3a79b4b/industrial_ci/src/docker.sh#L116.

If there's no problem, I'm interested in adding a generic option to get an access to the built image.

In my usecase, I want to store Docker images with the latest code. To achieve that currently I have CI jobs that run `docker build`, separately from industrial_ci jobs. While that works ok, since in industrial_ci an image is already built I'm interested in re-using what's already built by industrial_ci. That allows to use the same image for testing on CI and for the future purpose, and also save some CI time (it's non-trivial for a private repo on a hosting service). It could also allow a usecase where the update gets deployed by container (common in web app world, but it's possible in robotics as well).

CC @AustinDeric

ipa-mdl:
> Is there any reason why there is no way to access the docker image built during industrial_ci process?
I see in a recent change there's an option to docker commit the built image

Yes, #279 was merged while it still was WIP.

> but the option is not publicly exposed (only available when running rerun_ci script AFAIUC).

https://github.com/ros-industrial/industrial_ci/pull/279#issuecomment-396168104

Just rename `_COMMIT_IMAGE` (and perhaps `_COMMIT_IMAGE_MSG`) and add documentation.

AustinDeric:
@ipa-mdl 
> The container gets deleted, but not the image.
I cannot find the image that gets built by industrial CI. i run `docker images` after the industrial ci  script and it doesn't return the recently built image:
```
  after_script:
    - docker images
```
Furthermore, looking at the build logs, the last command is `docker rm` and i do not see a commit of the image.

ipa-mdl:
The commit will only happen if `_COMMIT_IMAGE` is set.

130s:
Addressed by #319

PR #319, opened 2018-09-18
Expose COMMIT_IMAGE to user

AustinDeric:
This PR is nessesary for [CI/CD built docker images](https://docs.gitlab.com/ee/ci/docker/using_docker_build.html) Allows the user to commit the image built by the CI script.  This allows the user to `docker push` from CI to a docker registry.  

addresses #318.


AustinDeric:
> Please do not mix in the YAML lint feature.

fixed.

>Why did you add that boolean flag? I try to avoid these if possible.

Agree in general but i was trying to explicitly maintain the previous behavior if the flag COMMIT_IMAGE flag is false.

Ready to merge.

doc/index.rst
```diff
@@ -185,6 +185,9 @@ Note that some of these currently tied only to a single option, but we still lea
...
 * **CATKIN_PARALLEL_TEST_JOBS** (default: -p4): Maximum number of packages which could be examined in parallel during the test run. If not set it's filled by `ROS_PARALLEL_JOBS`.
 * **CCACHE_DIR** (default: not set): If set, `ccache <https://en.wikipedia.org/wiki/Ccache>`_ gets enabled for your build to speed up the subsequent builds in the same job if anything. See `detail. <https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst#cache-build-artifacts-to-speed-up-the-subsequent-builds-if-any>`_
+* **COMMIT_IMAGE** (default: false): If set to true, the docker images which contains the build and test will be saved in the outer-layer docker which runs the script.  The commited image is then be available to be pushed to a docker registry.  This behavior enables a building Docker images with CI/CD approach. If set to false, or unset the container will not be commited and is removed.
```
130s: Naming `COMMIT_IMAGE_*` might not be intuitive enough. So far, all Docker-related variables are prefixed with `DOCKER_`. Should we follow that convention? That will make some of the names long but I think verbose name is still better.

130s:
Friendly ping @ipa-mdl for another review please.

doc/index.rst
```diff
@@ -185,6 +185,9 @@ Note that some of these currently tied only to a single option, but we still lea
...
+* **COMMIT_IMAGE_NAME** (default: not set): used to specify an image name during the docker commit command which is triggered by setting COMMIT_IMAGE=true. If unset and if COMMIT_IMAGE=true then container id will be used as the image name.
+* **COMMIT_IMAGE_MSG** (default: not set): used to specify a commit during the docker commit command which is triggered by setting COMMIT_IMAGE=true. If unset and if COMMIT_IMAGE=true then the commit message will be empty.
```
130s: > If unset and if COMMIT_IMAGE=true then the commit message will be empty.

Not sure if this is true. In [this line](https://github.com/ros-industrial/industrial_ci/pull/319/files#diff-67ae1351f65f3d5321aa8cc3782c7f2eR55) the variable seems to be filled with certain value..

industrial_ci/src/docker.sh
```diff
@@ -65,8 +65,8 @@ function ici_require_run_in_docker() {
 #######################################
 function ici_run_cmd_in_docker() {
   local run_opts=($DOCKER_RUN_OPTS)
-  local commit_image=$_COMMIT_IMAGE
-  unset _COMMIT_IMAGE
+  local commit_image=$COMMIT_IMAGE
```
130s: If the role of `COMMIT_IMAGE` is nothing more than boolean, then we can check by other `COMMIT_IMAGE`-related variables whether to enable this feature or not. This way we can omit COMMIT_IMAGE variable.

ipa-mdl:
@130s| @AustinDeric: this push went wrong

130s:
@ipa-mdl 

> @130s| @AustinDeric: this push went wrong

Could you elaborate? I looked at the latest changes and can't figure out where you mean is wrong.

130s:
Note that currently there's [1 CI job failing](https://travis-ci.org/ros-industrial/industrial_ci/jobs/439163927) due to downstream issue https://github.com/ros-controls/ros_control/issues/361.

ipa-mdl:
At least 4b021a1911088602735551d36be13d3c75db1d7a does not belong in here.

130s:
Rebased to origin. Please review @ipa-mdl

.travis.yml
```diff
@@ -77,3 +78,4 @@ install:
 script:
   - industrial_ci/_wrap_test.sh # this script is only for internal use
   - if ! [ -z "$POST_PROCESS" ]; then $POST_PROCESS; fi
+  - if ! [ -z "$DOCKER_COMMIT" ]; then  echo "Test if the Docker image is committed or not. See the size of the Docker image is non-zero or not."; if docker image inspect $DOCKER_COMMIT --format='{{.Size}}'; then exit 0; else exit 1; fi fi
```
ipa-mdl: No need for the inner if branch

industrial_ci/src/env.sh
```diff
@@ -64,6 +64,7 @@ fi
 export OS_CODE_NAME
 export OS_NAME
 export DOCKER_BASE_IMAGE
+export DOCKER_COMMIT
```
ipa-mdl: not needed, the other ones might be superfluous as well..

industrial_ci/src/docker.sh
```diff
@@ -111,7 +111,10 @@ function ici_run_cmd_in_docker() {
   wait %% || ret=$?
   trap - INT
   if [ -n "$commit_image" ]; then
-    docker commit -m "$_COMMIT_IMAGE_MSG" "$cid" $commit_image > /dev/null
+    if [ "use_containerid" != "$commit_image" ]; then
```
ipa-mdl: The effect could be achieved by passing an empty string as `DOCKER_COMMIT`.
(but the length check in line 113 needs to be updated)
In addtion, `use_containerid`  is really misleading, because the container is not used, but the image will be untagged.

IMHO it is a bad idea to commit without a tag, it just adds lot anonymous images.

doc/index.rst
```diff
@@ -188,6 +188,8 @@ Note that some of these currently tied only to a single option, but we still lea
 * **DEBUG_BASH** (default: not set): If set with any value (e.g. `true`), all executed commands that are not printed by default to reduce print space will be printed.
 * **DOCKER_BASE_IMAGE** (default: $OS_NAME:$OS_CODE_NAME): Base image used for building the CI image. Could be used to pre-bundle dependecies or to run tests for different architectures. See `this PR <https://github.com/ros-industrial/industrial_ci/pull/174>`_ for more info.
 * **DOCKER_BUILD_OPTS** (default: not set): Used do specify additional build options for Docker.
+* **DOCKER_COMMIT** (default: not set): If set, the docker images which contains the build and test will be saved in the outer-layer docker which runs the script.  The commited image is then be available to be pushed to a docker registry.  This behavior enables a building Docker images with CI/CD approach. If unset the container will not be commited and is removed. The value set is used to specify an image name during the docker commit command which is triggered by setting ``DOCKER_COMMIT_IMAGE=true``. If special value "use_containerid" is set, then container id will be used as the image name.
```
ipa-mdl: needs to be updated, `DOCKER_COMMIT_IMAGE` is not used anymore

.travis.yml
130s: I'm not sure what you exactly mean. Does https://github.com/AustinDeric/industrial_ci/pull/3 look better @ipa-mdl?

industrial_ci/src/docker.sh
130s: Talked to @AustinDeric offline. He thinks we can remove this logic. I'll add the change in  https://github.com/AustinDeric/industrial_ci/pull/3

.travis.yml
```diff
@@ -77,3 +78,4 @@ install:
 script:
   - industrial_ci/_wrap_test.sh # this script is only for internal use
   - if ! [ -z "$POST_PROCESS" ]; then $POST_PROCESS; fi
+  - if [ -n "$DOCKER_COMMIT" ]; then echo "Test if the Docker image is committed or not. See the size of the Docker image is non-zero or not."; docker image inspect "$DOCKER_COMMIT" --format="$DOCKER_COMMIT: {{.Size}} bytes"; fi
```
ipa-mdl: The echo'd text does not make sense anymore.
It might be better to just set
```
POST_PROCESS=docker image inspect "$DOCKER_COMMIT" --format="$DOCKER_COMMIT: {{.Size}} bytes"
```
And line ending is missing

doc/index.rst
```diff
@@ -188,6 +188,8 @@ Note that some of these currently tied only to a single option, but we still lea
 * **DEBUG_BASH** (default: not set): If set with any value (e.g. `true`), all executed commands that are not printed by default to reduce print space will be printed.
 * **DOCKER_BASE_IMAGE** (default: $OS_NAME:$OS_CODE_NAME): Base image used for building the CI image. Could be used to pre-bundle dependecies or to run tests for different architectures. See `this PR <https://github.com/ros-industrial/industrial_ci/pull/174>`_ for more info.
 * **DOCKER_BUILD_OPTS** (default: not set): Used do specify additional build options for Docker.
+* **DOCKER_COMMIT** (default: not set): If set, the docker image, which contains the build and test artifacts, will be saved in the outer-layer docker which runs the ``industrial_ci`` script and thus will become accessible for later usage (e.g. you can then push to your docker registry). If unset, the container will not be commited and is removed. The value is used to specify an image name during the ``docker commit`` command.
```
ipa-mdl: "If set, the docker **container**"
" will be saved **as an image**"

doc/index.rst
```diff
@@ -188,6 +188,8 @@ Note that some of these currently tied only to a single option, but we still lea
...
+* **DOCKER_COMMIT_MSG** (default: not set): used to specify a commit during the docker commit command which is triggered by setting ``DOCKER_COMMIT``. If unset and if ``DOCKER_COMMIT`` is set then the commit message will be empty. See more ``DOCKER_COMMIT``.
```
ipa-mdl: I think the second sentence can get removed, it correct, but superfluous..

doc/index.rst
```diff
@@ -245,6 +247,24 @@ On CI platform usually some variables are available for the convenience. Since a
 Still, you may want to pass some other vars. `DOCKER_RUN_OPTS='-e MY_VARIABLE_VALUE'` should do the trick.
 You can even set it to a specific value: `DOCKER_RUN_OPTS='-e MY_VARIABLE_VALUE=42'` (format varies per CI platform. These are Gitlab CI example).
 
+Re-use the container image
+--------------------------
+
+``industrial_ci`` builds a ``Docker`` image using the associated repository on the specified operating system per every job. While the built Docker image is thrown away once the job finishes by default, there's a way to access the built image post job so that you can re-use it.
```
ipa-mdl: The built image is kept.
Only the container gets removed.

doc/index.rst
```diff
@@ -245,6 +247,24 @@ On CI platform usually some variables are available for the convenience. Since a
...
+
+  variables:
+      DOCKER_COMMIT=registry.gitlab.com/your-org/your-repo:your_img
+  :
+  script:
+      - docker push $DOCKER_COMMIT
+
+Note for the built Docker image:
+
+- ``apt`` repositories (i.e. Result of ``apt-get update``) are removed.
```
ipa-mdl: This depends on `DOCKER_BASE_IMAGE`, but is true for almost all Debian/Ubuntu-based images.

doc/index.rst
```diff
@@ -245,6 +247,24 @@ On CI platform usually some variables are available for the convenience. Since a
...
+- Some utility packages are pre-installed. As of Sept 2018, those are: ``build-essential, python-catkin-tools, python-pip, python-rosdep, python-wstool, ros-$ROS_DISTRO-catkin, ssh-client``.
```
ipa-mdl: This is only true if `DOCKER_IMAGE` is not set.

doc/index.rst
```diff
@@ -245,6 +247,24 @@ On CI platform usually some variables are available for the convenience. Since a
...
+
+To do so, simply set ``DOCKER_COMMIT`` the name of the image of your choice. Then you'll be able to access that image. For example in your CI config (e.g. ``.travis.yml``), add something like ::
+
+  variables:
+      DOCKER_COMMIT=registry.gitlab.com/your-org/your-repo:your_img
+  :
+  script:
+      - docker push $DOCKER_COMMIT
+
+Note for the built Docker image:
```
ipa-mdl: I would remove this note section. It is way more complicated and highl depends on the use case.

industrial_ci/src/docker.sh
```diff
@@ -111,7 +111,7 @@ function ici_run_cmd_in_docker() {
   wait %% || ret=$?
   trap - INT
   if [ -n "$commit_image" ]; then
-    docker commit -m "$_COMMIT_IMAGE_MSG" "$cid" $commit_image > /dev/null
```
ipa-mdl: I'd like to keep `> /dev/null`, otherwise the SHA gets printed without context.
Instead a line could be printed: "Committing container to $DOCKER_COMMIT".

.travis.yml
```diff
@@ -58,6 +58,7 @@ env:
     - ROS_DISTRO=melodic OS_NAME=debian OS_CODE_NAME=stretch AFTER_SCRIPT='grep -q ID=debian /etc/os-release && grep -q VERSION_ID=\"9\" /etc/os-release'
     - ROS_DISTRO=melodic OS_NAME=debian EXPECT_EXIT_CODE=1
     - ROS_DISTRO=melodic OS_NAME=debian OS_CODE_NAME=bionic EXPECT_EXIT_CODE=1
+    - ROS_DISTRO=melodic DOCKER_COMMIT="img_temp"
```
ipa-mdl: Do we really need to test this in a separate job?
We could add this easily to any other entry.

130s:
I'm testing this change (on Gitlab CI) and noticed a job gets stuck at `docker commit`.

Looks like docker is trying to pull `gitlab/gitlab-runner-helper:x86_64-e900028d` image, but not yet sure why.
```
:
Function catkin_build returned with code '0' after 7 min 44 sec 
:
++ '[' -n registry.gitlab.com/130s/projS:tagT ']'
++ '[' use_containerid '!=' registry.gitlab.com/130s/projS:tagT ']'
++ commit_image_name=registry.gitlab.com/130s/projS:tagT
++ docker commit -m '' cc44e77b76648ed18cb5521061a1bf55188f19eb4c36db2ef7f216cb4eecc046 registry.gitlab.com/130s/projS:tagT
Pulling docker image gitlab/gitlab-runner-helper:x86_64-e900028d ...
(no progress after this)
```
```
$ more .gitlab-ci.yml 
image: docker:git
variables:
  ADDITIONAL_DEBS: "build-essential python-catkin-tools python-pip ssh"  # https://github.com/ros-industrial/industrial_ci/issues/216#issuecomment-331136604
  CATKIN_CONFIG: "-DCMAKE_BUILD_TYPE=Release"
  CCACHE_DIR: "$HOME/.ccache"
  DEBUG_BASH: "true"
  DOCKER_DRIVER: "overlay" # recommended for docker-in-docker
  DOCKER_RUN_OPTS: "-e ${CI_REPOSITORY_URL} -e ${CI_COMMIT_REF_NAME}"
  ROS_PARALLEL_JOBS: "j2"
  TMPDIR: "${CI_PROJECT_DIR}.tmp"
services:
  - docker:dind # enable docker-in-docker
before_script:
  - apk add --update bash coreutils tar
  - git clone https://github.com/ros-industrial/industrial_ci.git .ci
  - source ./.gitlab-ci_util.sh
  - sshkey_gitlabci
  - docker login -u gitlab-ci-token -p $CI_JOB_TOKEN $CI_REGISTRY

indigo:
  script:
    - .ci/gitlab.sh
    - docker push $DOCKER_COMMIT
  variables:
    DOCKER_COMMIT: "$CI_REGISTRY_IMAGE:indigo-$CI_COMMIT_REF_SLUG"
    DOCKER_IMAGE: "rosindustrial/ci:indigo-trusty"
    NOT_TEST_BUILD: "true"
    NOT_TEST_INSTALL: "true"
    ROS_DISTRO: "indigo"
```

ipa-mdl:
`docker commit` can take a long time, especially on slow storage backends! It hashes the contents of the added/updated files.
Perhaps it makes more sense to just add an option to name the container and keep it.
`docker commit / push` could be run in the user script.
Please make sure to remove the image after you have pushed it!
This advice should get added to the documentation.

> Looks like docker is trying to pull gitlab/gitlab-runner-helper:x86_64-e900028d image, but not yet sure why.

This is unrelated.

130s:
Using an image built by this change, I see the following error when starting a container. 
```
$ docker pull registry.gitlab.com/foo/baa:tagoo
tagoo: Pulling from foo/baa
:
bae143c83a43: Already exists 
592e1e3560c1: Pull complete 
Digest: sha256:50bde1dd64cba6d009cd01fcc3c4458dea89847899e56710ac4668d8661b315e
Status: Downloaded newer image for registry.gitlab.com/foo/baa:tagoo
$ docker run -it registry.gitlab.com/foo/baa:tagoo       
/bin/bash: /root/ici/ci_main.sh: No such file or directory
$
```

130s:
>> I'm testing this change (on Gitlab CI) and noticed a job gets stuck at docker commit.
>
> docker commit can take a long time, especially on slow storage backends! It hashes the contents of the added/updated files.

Indeed, it was just taking longer time (and my CI limit time was short. After making it longer the CI job finished). Maybe nice to show a warning message.

AustinDeric:
I am going to close out this PR and @130s is going to open a new one with more concise changes.

Issue #316, opened 2018-08-16
Add clang format check?


130s:
It's already listed in https://github.com/ros-industrial/industrial_ci/issues/98 but I spin it off as a separate ticket.

ipa-mdl:
Sry, clicked wrong

PR #320, opened 2018-09-21
Add clang-format check

ipa-jfh:
This PR tries to solve https://github.com/ros-industrial/industrial_ci/issues/316.

### Logic of new argument:
* **CLANGCHECK** (default: not set. Value range: [`<format-style>`|false]): If set or a .clang-format file is in the repository root, run the clang-format check. If argument is not conform to [`<format-style>`](https://bcain-llvm.readthedocs.io/projects/clang/en/release_38/ClangFormatStyleOptions/#configurable-format-style-options) use the default style. Setting the argument to false will ignore an existing .clang-format file.

### Tested with:
```bash
 roscd my_package
 rosrun industrial_ci run_ci ROS_DISTRO=indigo ROS_REPO=ros-shadow-fixed CLANGCHECK=true 
```
and
```bash 
roscd my_package
clang-format-3.8 -style=llvm -dump-config > .clang-format
rosrun industrial_ci run_ci ROS_DISTRO=indigo ROS_REPO=ros-shadow-fixed 
```

Feedback is welcome :)


industrial_ci/src/tests/clang_format_check.sh
```diff
@@ -0,0 +1,61 @@
...
+  local version="${1:--3.8}" # default = -3.8
+  local ERR=0
+
+  ici_require_run_in_docker # this script must be run in docker
+  
+  ici_time_start clang_install
+    clang_install "$version"
+  ici_time_end # clang_install
+
+  ici_time_start define_clang_format
```
ipa-mdl: No need to time this. I would prefer to branch the format without any function call.

industrial_ci/src/tests/clang_format_check.sh
```diff
@@ -0,0 +1,61 @@
...
+  if [ -f "$TARGET_REPO_PATH/.clang-format" ]; then
+    echo "Found '.clang-format' file."
+    CLANG_FORMAT="file"
+  else    
+    CLANG_FORMAT="$CLANGCHECK" # If not valid Clang uses default
+  fi
+}
+
+function run_clang_format_check() {
+  local version="${1:--3.8}" # default = -3.8
```
ipa-mdl: Pinning the version will not work with other linux distros.

industrial_ci/src/tests/clang_format_check.sh
```diff
@@ -0,0 +1,61 @@
...
+function clang_install() {
+  local version="$1"
+  sudo apt-get update -qq
+  sudo apt-get install -y -qq clang-format"$version"
+}
+
+function define_clang_format() {
+  if [ -f "$TARGET_REPO_PATH/.clang-format" ]; then
+    echo "Found '.clang-format' file."
+    CLANG_FORMAT="file"
```
ipa-mdl: Why not use `CLANG_FORMAT` right away? `CLANGCHECK` is too generic.

industrial_ci/src/ci_main.sh
```diff
@@ -39,6 +39,9 @@ if [ "$PRERELEASE" == true ]; then
 elif [ -n "$ABICHECK_URL" ]; then
   source ${ICI_SRC_PATH}/tests/abi_check.sh
   run_abi_check
+elif [ -f "$TARGET_REPO_PATH/.clang-format" -o -n "$CLANGCHECK" ] && [ "$CLANGCHECK" != "false" ]; then
```
ipa-mdl: This would prevent all other tests in case `.clang_format` is present.
I see two options:

1. Use `CLANG_FORMAT=<style>|file` to eable the test. In case of "file", specifiy the absolute path to `.clang_format`.
2. Move into `source_tests.sh`

Option 1 would enable to allow failures regarding the format.
Option 2 would allow the auto-check (with opt-out).

industrial_ci/src/tests/clang_format_check.sh
```diff
@@ -0,0 +1,61 @@
...
+  
+  ici_time_start run_clang_format_check
+  while read f; do
+    if ! clang-format"$version" -style="$CLANG_FORMAT" "$f" | git diff --exit-code "$f" - ; then
+      ERR=$[$ERR +1]
+    fi
+  done < <(find "$TARGET_REPO_PATH" -name '*.h' -or -name '*.hpp' -or -name '*.cpp') 
+  
+  if [ "$ERR" -ne "0" ]; then
+      echo "Clang format check failed for $ERR file(s)."
```
ipa-mdl: Please use `ici_error` instead.

industrial_ci/src/tests/clang_format_check.sh
```diff
@@ -0,0 +1,61 @@
...
+    clang_install "$version"
+  ici_time_end # clang_install
+
+  ici_time_start define_clang_format
+    define_clang_format
+  ici_time_end # define_clang_format
+  
+  ici_time_start run_clang_format_check
+  while read f; do
+    if ! clang-format"$version" -style="$CLANG_FORMAT" "$f" | git diff --exit-code "$f" - ; then
```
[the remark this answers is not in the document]
ipa-jfh: It is used in the form [`git diff [<options>] --no-index [--] <path> <path>` ](https://git-scm.com/docs/git-diff#git-diff-emgitdiffemltoptionsgt--no-index--ltpathgtltpathgt).
This method also works outside a git repository (and then, one can omit the `--no-index`).

I forgot to add git as an installation dependency, which I will correct in the next update.

industrial_ci/src/tests/clang_format_check.sh
ipa-jfh: I removed the functions in 97bb8dc and b9885f4

industrial_ci/src/tests/clang_format_check.sh
ipa-jfh: I removed the version and use only `clang-format`. I checked that install and execute work.

ipa-jfh:
> It would be good to have at least two test cases for this feature.

526eea6  introduces tests for CLANG_FORMAT_CHECK. 

(I can squash the commits once approved)

130s:
I'm excited for this addition! Travis CI was failing for unrelated reasons so I restarted some failing jobs and now passed. I leave the primary review up to @ipa-mdl.

.travis.yml
```diff
@@ -59,14 +59,19 @@ env:
...
     - ROS_DISTRO=kinetic _EXTERNAL_REPO='ros-controls/ros_control#kinetic-devel' UPSTREAM_WORKSPACE=file ROSINSTALL_FILENAME=ros_control.rosinstall ROS_PARALLEL_TEST_JOBS=-j1
     - ROS_DISTRO=kinetic _EXTERNAL_REPO='ipa320/cob_calibration_data#indigo_dev' ROS_REPO=ros UPSTREAM_WORKSPACE=file AFTER_SCRIPT='./.travis.xacro_test.sh'
     - ROS_DISTRO=indigo _EXTERNAL_REPO='ros/actionlib#38ce66e2ae2ec9c19cf12ab22d57a8134a9285be' ROS_REPO=ros ABICHECK_URL=url ABICHECK_MERGE=true # actual URL will not be used in the case
     - ROS_DISTRO=kinetic _EXTERNAL_REPO='ros-industrial/ros_canopen#0.7.5' ROS_REPO=ros ABICHECK_URL='github:ros-industrial/ros_canopen#0.7.1' ABICHECK_MERGE=false EXPECT_EXIT_CODE=1
     - ROS_DISTRO=kinetic _EXTERNAL_REPO='ros-industrial/ros_canopen#0.7.6' ABICHECK_URL='github:ros-industrial/ros_canopen#0.7.5' ABICHECK_MERGE=false
+
+    # Format tests
+    - ROS_DISTRO=kinetic USE_MOCKUP='industrial_ci/mockups/format_tests' CLANG_FORMAT_CHECK='LLVM'
```
ipa-mdl: To test this properly, I would suggest to add 4 tests in total:
1. pkg_a, fails for file
1. pkg_a, passes for llvm
1. pkg_b, passes for file
1. pkg_b, fails for llvm

industrial_ci/src/tests/clang_format_check.sh
```diff
@@ -0,0 +1,48 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+function run_clang_format_check() {
+  local err=0
+  local path=$TARGET_REPO_PATH
+
+  ici_require_run_in_docker # this script must be run in docker
```
ipa-mdl: I think it would make sense to use the base image for the format test, I will try it later..

industrial_ci/src/tests/clang_format_check.sh
```diff
@@ -0,0 +1,48 @@
...
+  
+  sudo apt-get update -qq
+  sudo apt-get install -y -qq git clang-format
```
ipa-mdl: `git-core` might be enough..

.travis.yml
ipa-jfh: `File` is currently equivalent to llvm.
So you are suggesting I should create the `file` based on some other format and then provide one test-package-mockup that is following that code format and one that is following the llvm format?
(instead of the current mock-ups of `format_tests` and `failing_format_tests`)

ipa-mdl: yes, this way you could test the effect of that option and verify the proper operation of your test.

ipa-mdl:
Code looks good nowm bit the tests fail.

ipa-jfh:
@ipa-mdl do you know where this error came from?
```bash
Function install_clang_format returned with code '0' after 0 min 22 sec 
run_clang_format_check
0.01s>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
Clang format check went successful.
Function run_clang_format_check returned with code '0' after 0 min 0 sec 


The command "industrial_ci/_wrap_test.sh" exited with 1.
0.00s$ if ! [ -z "$POST_PROCESS" ]; then $POST_PROCESS; fi
"post processing reached"


The command "if ! [ -z "$POST_PROCESS" ]; then $POST_PROCESS; fi" exited with 0.
```
https://travis-ci.org/ros-industrial/industrial_ci/jobs/452155703

ipa-mdl:
I would suggest to change the config file a little bit to enforce the failure or use a more complicated example file.

PR #321, opened 2018-09-28
Mark bitbucket.sh also as executeable

LoyVanBeek:
Otherwise, it cannot be used in a bitbucket pipeline


LoyVanBeek:
Apparently not much BitBucket users. But it's a very easy fix luckily.

ipa-mdl:
As alternative, it might be prefixed by `bash`.
But these scripts are meant to be executable.

PR #324, opened 2018-10-27
Travis timeout guard

ipa-mdl:
This patch will print a line to the output after each 9 minutes of "silence" preventing Travis from stopping the build.
In contrast to `travis_wait` the output will not get printed at the end, but line-wise.

This might introduce a small performance penalty, because each output line is processed by bash,
but I guess this is neglectable.

Not sure if this feature needs an opt-in or opt-out..


ipa-mdl:
Feature in action with 10s interval: [Job #1352.9](https://travis-ci.org/ros-industrial/industrial_ci/jobs/447143340#L482) .

ipa-mdl:
> Is there any reason that this function will only be added to Travis CI?

Only Travis seems to stop jobs if no output was received (?).

PR #326, opened 2018-10-30
Fix passing of exit code

ipa-mdl:


ipa-mdl:
This is a hotfix and first tests pass, will merge right away..

PR #327, opened 2018-11-01
Add ROSDEP_OPTIONS option.

130s:
# Motivation
- We can't customize `rosdep install` currently on industrial_ci. So it's good to have an option that allows customization.
- I just faced a situation where some packages pre-installed on a Docker image were older and they caused an issue (too complicated to explain in short. See [the answer in this thread](https://answers.ros.org/question/278847/problem-in-moveit-tutorials-step-movegroup-interface-tutorial/?answer=307347#post-id-307347) where a workaround for me was to re-install those packages from apt repo. With the current industrial_ci I can't find a way to work it around on CI.

# Approach

For `rosdep install`,
- allow to pass options.
- also allow verbose output.


130s:
> Removing `-q` is a bad idea (super verbose output!) and it conflicts with #287

Ok.
(I thought there was a usecase where I wanted to see verbose output of `rosdep install`, but with the following point, I don't think I still want it.)

> The reinstall feature ... it might not resolve these `soname` issues in all cases. (it might not reinstall transient dependencies..).

That is true. I tried but it didn't re-install the transitive dependency.

With above there's nothing good about the approach in this PR. Closing.

ipa-mdl:
Actually, dropping `-q` just adds the summary and the download status.
At the moment this does not have any effect because the output filter is still in place.

I try to silence `dpkg`..

PR #329, opened 2018-11-04
pass OS_* variables into docker

ipa-mdl:
For use with code checkes which do not depend on a specific ROS distros (like #320)


ipa-mdl:
I don't merge this for now. I think `env.sh` should be rewritten to support late evaluation instead.

PR #331, opened 2018-11-13
run_travis fixes

ipa-mdl:
Small fixes for run_travis. Now `run_travis` can parse `industral_ci`'s .travis.yml again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ notifications:
 env:
   global:
     - POST_PROCESS='echo "post processing reached"'
-    - BEFORE_SCRIPT='echo current dir: $(pwd)'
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -16,7 +16,7 @@ notifications:
 env:
   global:
     - POST_PROCESS='echo "post processing reached"'
-    - BEFORE_SCRIPT='echo current dir: $(pwd)'
+    - "BEFORE_SCRIPT='echo current dir: $(pwd)'"
```
130s: Does this need to be documented for the user's doc or not?

ipa-mdl: I don't think so. This is a pretty rare case in which ":<SPACE>" is forcing the line to be intepreted as mapping, at least with pyyaml.

PR #332, opened 2018-11-15
[doc] Clarify clang check config file requirement.

130s:
I don't know if those who use `clang-format` regularly should know already about this, but the file name requirement was a little tricky to me.

Please review @ipa-mdl @ipa-jfh


ipa-mdl:
A link to https://clang.llvm.org/docs/ClangFormat.html should be sufficient, the others can be removed..

ipa-jfh:
Maybe sth like this? (url is shown wrong)

* **CLANG_FORMAT_CHECK** (default: not set. Value range: ``[<format-style>|file]``): If set, run the `clang-format <https://clang.llvm.org/docs/ClangFormat.html>`_ check (Set the argument to ``file`` if the style configuration should be loaded from a ``.clang-format`` file, located in one of the parent directories of the source file).

ipa-mdl:
I like @ipa-jfh's new version, but I would leave the brackets out

130s:
Hope my latest commit looks as exact as @ipa-jfh suggested.

130s:
Please re-review @ipa-mdl @ipa-jfh

ipa-mdl:
I have add some fixes. Please squash all commits.

ipa-jfh:
Looks good to me :)

130s:
Thanks for the review and fixes.

PR #334, opened 2018-11-19
Add .clang-format to LLVM test

wxmerkt:
The WebKit test contains the dumped .clang-format from clang-format-3.8, the LLVM does not - this leads to failure of the Industrial CI with the CLANG_FORMAT_CHECK=file defined if the user workspace contains a differing .clang-format.

We use the standard Industrial CI config and just started testing the format check (we have a custom .clang-format in the root of our repository) - without this change it fails as:

image

### DIFF
--- a/industrial_ci/mockups/format_tests/cpp/LLVM/.clang-format
+++ b/industrial_ci/mockups/format_tests/cpp/LLVM/.clang-format
@@ -1,0 +1,90 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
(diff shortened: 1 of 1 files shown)


wxmerkt:
The default ROS Industrial travis.yml config clones the repository into ``.ci_config`` - which as underneath the regular ROS package gets checked and then throws - is there another workaround that is possible?

ipa-mdl:
> The default ROS Industrial travis.yml config clones the repository into .ci_config
`.industrial_ci` ;)

wxmerkt:
Superseded by #335

PR #335, opened 2018-11-19
fix clang format check crawl

ipa-mdl:
This should skip hidden directories in the target path.

@wxmerkt: please test

### DIFF
--- a/industrial_ci/src/tests/clang_format_check.sh
+++ b/industrial_ci/src/tests/clang_format_check.sh
@@ -38,7 +38,7 @@ function run_clang_format_check() {
     if ! clang-format -style="$CLANG_FORMAT_CHECK" "$file" | git diff --exit-code "$file" - ; then
       err=$[$err +1]
     fi
-  done < <(find "$path"/* -name '*.h' -or -name '*.hpp' -or -name '*.cpp')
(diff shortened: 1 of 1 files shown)


wxmerkt:
@ipa-mdl Thanks for looking into this. This fix works as well.

industrial_ci/src/tests/clang_format_check.sh
```diff
@@ -38,7 +38,7 @@ function run_clang_format_check() {
     if ! clang-format -style="$CLANG_FORMAT_CHECK" "$file" | git diff --exit-code "$file" - ; then
       err=$[$err +1]
     fi
-  done < <(find "$path" -name '*.h' -or -name '*.hpp' -or -name '*.cpp')
+  done < <(find "$path"/* -name '*.h' -or -name '*.hpp' -or -name '*.cpp')
```
wxmerkt: Let's add ``-or -name '*.cc'`` here - some code guides (Google) recommend it for C++ files. What do you think @ipa-mdl? Perhaps even ``*.hh``.

ipa-mdl: Good point. Perhaps even `*.c`

PR #338, opened 2018-12-04
 Accept versatile 'catkin build' option.

130s:
At the moment, no versatile options are available to pass arguments to `catkin build` (only options with specific purposes are there).
`OPT_VI` variable might be the closest, which is, however, not exposed to users.

This change allows configuring 'catkin build'.


130s:
Only hydro CI jobs are failing.
https://travis-ci.org/ros-industrial/industrial_ci/jobs/463128257#L467
```
Function prepare_docker_image returned with code '0' after 1 min 30 sec 
/root/ici/tests/source_tests.sh: line 30: CATKIN_BUILD_OPT: command not found
The command "industrial_ci/_wrap_test.sh" exited with 127.
```

130s:
Closing. Using an existing variable `CATKIN_CONFIG` does the same.

Issue #339, opened 2018-12-08
Specify clang-format-check version


wxmerkt:
The clang-format-check added in #320 uses the system-default clang-format (which changes between operating system versions). It would be great if the version could be specified.

ipa-mdl:
You can specify the version indirectly by changing `ROS_DISTRO. For now it is required to bet set it, but is only used for selecting a Docker image.

Which version do you want to use? And how to install it?

wxmerkt:
We noticed differences in formatting between versions and thus want to specify what the agreed upon standard version is - across distros (we support indigo, kinetic, melodic). Eg for kinetic and melodic we want to specify CLANG_FORMAT_EXECUTABLE=clang-format-3.9 - with this version apt-get installed and used in the formatting command.

ipa-mdl:
The problem is that not all `clang-format` versions are availailable for all Ubuntu versions..
However, I just found out that there is no `clang-format` alias for `indigo`/`trusty`.

In general `CLANG_FORMAT_EXECUTABLE` makes sense to me, espcially if we default it to `clang-format`.
It would be great if you could prepare a pull request for this.

ipa-mdl:
fixed in #340

PR #340, opened 2018-12-10
Allow clang-format version to be set

wxmerkt:
I've tested this on our build server and it works (if the set version exists and can be apt-get installed, of course).

- Relates to #339 

cc: @ipa-mdl


doc/index.rst
```diff
@@ -177,7 +177,7 @@ Note that some of these currently tied only to a single option, but we still lea
...
 * **CCACHE_DIR** (default: not set): If set, `ccache <https://en.wikipedia.org/wiki/Ccache>`_ gets enabled for your build to speed up the subsequent builds in the same job if anything. See `detail. <https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst#cache-build-artifacts-to-speed-up-the-subsequent-builds-if-any>`_
-* **CLANG_FORMAT_CHECK** (default: not set. Value range: ``[<format-style>|file]``): If set, run the `clang-format <https://clang.llvm.org/docs/ClangFormat.html>`_ check. Set the argument to ``file`` if the style configuration should be loaded from a ``.clang-format`` file, located in one of the parent directories of the source file.
+* **CLANG_FORMAT_CHECK** (default: not set. Value range: ``[<format-style>|file]``): If set, run the `clang-format <https://clang.llvm.org/docs/ClangFormat.html>`_ check. Set the argument to ``file`` if the style configuration should be loaded from a ``.clang-format`` file, located in one of the parent directories of the source file. If a particular version of clang-format is desired, please specify via **CLANG_FORMAT_EXECUTABLE** (default: clang-format).
```
ipa-mdl: Please split both and give an example for `CLANG_FORMAT_EXECUTABLE`

industrial_ci/src/tests/clang_format_check.sh
```diff
@@ -21,9 +21,15 @@ function run_clang_format_check() {
 
   DOCKER_IMAGE="$DOCKER_BASE_IMAGE" ici_require_run_in_docker # this script must be run in docker
 
+  # Check whether a specific version of clang-format is desired
```
ipa-mdl: variables names should not start with underscores.
I'd prefer:
```bash
local clang_format_executable=${CLANG_FORMAT_EXECUTABLE:-clang-format}
```

ipa-mdl:
It might make sense to migrate one (passing) test case to indigo, which require the executable to be set.

wxmerkt:
@ipa-mdl Thanks for the review. PR updated and passing test for Indigo added.

ipa-mdl:
Just another idea: What do you think of a `CLANG_FORMAT_VERSION` variables?
```bash
local clang_format_executable="catkin-format${CLANG_FORMAT_VERSION:+-}$CLANG_FORMAT_VERSION"
```

wxmerkt:
I think a ``CLANG_FORMAT_VERSIION`` variable would be workable as well, however, I see other added benefits with specifying the full executable name - users can specify other command line parameters/options beyond the default ``style=file``. Should I still go ahead and amend?

ipa-mdl:
> I see other added benefits with specifying the full executable name - users can specify other command line parameters/options beyond the default style=file. 

This would not work because of the apt install..
It be okay to an a variable for additional options.

> Should I still go ahead and amend?

Yes, if it okay for you.
Specificying the version only would be less error-prone.

wxmerkt:
Done! :-)

industrial_ci/src/tests/clang_format_check.sh
```diff
@@ -22,7 +22,7 @@ function run_clang_format_check() {
   DOCKER_IMAGE="$DOCKER_BASE_IMAGE" ici_require_run_in_docker # this script must be run in docker
 
   # Check whether a specific version of clang-format is desired
-  local clang_format_executable="catkin-format${CLANG_FORMAT_VERSION:+-$CLANG_FORMAT_VERSION}"
+  local clang_format_executable="clang-format${CLANG_FORMAT_VERSION:+-$CLANG_FORMAT_VERSION}"
```
ipa-mdl: @wxmerkt: sry, my bad.

PR #341, opened 2018-12-12
Add a feature to commit Docker image (reopen #319)

130s:
# What does this PR change

Re-opening #319 using the commits from the requester there.
This adds a feature to run `docker commit` from industrial_ci, so that the users can access the built docker image for later usage.

Review comments up to this point https://github.com/ros-industrial/industrial_ci/pull/319#pullrequestreview-180176066 are hopefully addressed.

## Maturity of the suggested feature

At the time of opening this PR, the feature is experimental. We've been actively using on our (private) repos, and I found there are still some basic issues that need to be fixed in order for this feature to be integrated into a full-automated cycle. That said, we've been able to use this feature in semi-automated situation where small manual change to the built image makes the image usable.

Instead of trying to fix all of those issues in a single initial PR, I'd suggest to move on, claiming experimental, merge, open tickets, and keep iterating.

# Review items
- [x] @130s Locally test.
- [x] CI pass.
- [ ] Review changes.


130s:
Not sure why Travis CI didn't grab this. Close-reopening.

doc/index.rst
```diff
@@ -179,6 +179,9 @@ Note that some of these currently tied only to a single option, but we still lea
...
 * **CLANG_FORMAT_CHECK** (default: not set. Value range: ``[<format-style>|file]``): If set, run the `clang-format <https://clang.llvm.org/docs/ClangFormat.html>`_ check. Set the argument to ``file`` if the style configuration should be loaded from a ``.clang-format`` file, located in one of the parent directories of the source file.
 * **CLANG_FORMAT_VERSION** (default: not set): Version of clang-format to install and use (relates to both the apt package name as well as the executable), e.g., ``CLANG_FORMAT_VERSION=3.8``.
+* **COMMIT_IMAGE** (default: false): If set to true, the docker images which contains the build and test will be saved in the outer-layer docker which runs the script.  The commited image is then be available to be pushed to a docker registry.  This behavior enables a building Docker images with CI/CD approach. If set to false, or unset the container will not be commited and is removed.
```
ipa-mdl: This is the outdated documentation..

.travis.yml
```diff
@@ -53,7 +53,7 @@ env:
     - ROS_DISTRO=jade   UPSTREAM_WORKSPACE=file  ROSINSTALL_FILENAME=.ci.rosinstall  # Testing arbitrary file name without ROS_DISTRO suffix. As of 6/3/2016 this fails due to https://github.com/ros-industrial/industrial_core/pull/144#issuecomment-223186764
     - ROS_DISTRO=jade   APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.vet  _DO_NOT_FOLD=true # Passing wrong SKS URL as a break test. Should still pass.
     - ROS_DISTRO=jade   UPSTREAM_WORKSPACE=file ROSINSTALL_FILENAME=.i.do.not.exist EXPECT_EXIT_CODE=1
-    - ROS_DISTRO=kinetic DOCKER_BASE_IMAGE="ros:kinetic-ros-base" NOT_TEST_BUILD='true' DEBUG_BASH='true' VERBOSE_OUTPUT='false'
+    - ROS_DISTRO=kinetic DOCKER_BASE_IMAGE="ros:kinetic-ros-base" NOT_TEST_BUILD='true' DEBUG_BASH='true' VERBOSE_OUTPUT='false' DOCKER_COMMIT="img_temp" POST_PROCESS=`docker image inspect "$DOCKER_COMMIT" --format="$DOCKER_COMMIT: {{.Size}} bytes"`
```
ipa-mdl: Please remove these backticks / replace them with single quotes. I guess the config is invalid and that's why the tests are not run.

industrial_ci/src/docker.sh
```diff
@@ -111,7 +111,8 @@ function ici_run_cmd_in_docker() {
   wait %% || ret=$?
   trap - INT
   if [ -n "$commit_image" ]; then
-    docker commit -m "$_COMMIT_IMAGE_MSG" "$cid" $commit_image > /dev/null
+  	echo "Committing container to tag: '$DOCKER_COMMIT'"
```
ipa-mdl: indention

130s:
Commit updated.

Locally I tested `rerun_ci`, which seems to work fine without `DOCKER_COMMIT` defined (i.e. I see a temporary Docker image commited), but with the arg the behavior is still the same as no arg.

```
$ SRCDIR=/tmp/cws_dr/src/ && mkdir -p $SRCDIR && cd $SRCDIR && git clone https://github.com/ros-drivers/openni2_camera
$ cd /tmp && git clone https://github.com/130s/industrial_ci -b feature/docker_commit
$ /tmp/industrial_ci/industrial_ci/scripts/rerun_ci . ROS_DISTRO=melodic ROS_REPO=ros-shadow-fixed
$ /tmp/industrial_ci/industrial_ci/scripts/rerun_ci . ROS_DISTRO=melodic ROS_REPO=ros-shadow-fixed DOCKER_COMMIT="test_rerun_ci" DOCKER_COMMIT_MSG="testing_rerun_msg"

$ docker images| grep -i openni2                                                                                                                                                                                                                                                                                                          
REPOSITORY                                                          TAG                                                           IMAGE ID            CREATED             SIZE
industrial-ci/rerun_ci/openni2_camera                               66d52ad29be5                                                  efa6e1eb336e        2 hours ago         1.76GB
```

130s:
Figured it out the reason why the specified image isn't committed https://github.com/ros-industrial/industrial_ci/pull/341#issuecomment-450568264. rerun_ci looks for environment variables while I only specified the command args. Will update.

130s:
Please re-review @ipa-mdl.

industrial_ci/scripts/rerun_ci
```diff
@@ -52,26 +52,26 @@ fi
 repo_dir=$(cd "$1" && pwd)
 shift
 
-_COMMIT_IMAGE_MSG="$repo_dir $*"
-env_hash=($(sha256sum <<< "$_COMMIT_IMAGE_MSG"))
-_COMMIT_IMAGE="industrial-ci/rerun_ci/$(basename "$repo_dir"):${env_hash:0:12}"
+if [ -z $DOCKER_COMMIT_MSG ]; then DOCKER_COMMIT_MSG="$repo_dir $*"; fi
```
ipa-mdl: `DOCKER_COMMIT_MSG` and `DOCKER_COMMIT` should not get overwritten from the outside, they are crucial to the argument hashing.
Please remove these if clauses.

ipa-mdl:
>  Could you elaborate how you want us to update rerun_ci

Just remove the if clauses and overwrite `DOCKER_COMMIT` and `DOCKER_COMMIT_MSG` (as before).

> then I thought we'd want rerun_ci script to behave the same. When a user defines the docker image name by DOCKER_COMMIT, wouldn't rerun_ci script save the state into the given image name instead of automatically generated hash value? And if that's correct wouldn't it be ok?

No.
`rerun_ci` is just a little helper to take care of `DOCKER_COMMIT`etc.
The hashing is the main functionality. If you change the image name or the message, it does not work properly anymore.
If hashing is not needed, just use `run_ci`.

130s:
> Just remove the if clauses and overwrite `DOCKER_COMMIT` and `DOCKER_COMMIT_MSG` (as before).

Ok.

In that case I don't see a point in updating the variable names (`_COMMIT_IMAGE` to `DOCKER_COMMIT` etc.) in `rerun_ci`. I've kept the change but let me know if I should revert.

Also updated the document, as I think the new feature added in this PR doesn't work for rerun (I tested it locally. Not sure if that's what's supposed to be though).

ipa-mdl:
Thanks!

> In that case I don't see a point in updating the variable names (_COMMIT_IMAGE to DOCKER_COMMIT etc.) in rerun_ci. I've kept the change but let me know if I should revert.

It is crucial to rename the variables everywhere. `rerun_ci` would not work anymore, it built around this feature.

> Also updated the document, as I think the new feature added in this PR doesn't work for rerun (I tested it locally. Not sure if that's what's supposed to be though).

I don't see the need for an "exception", but we could add a note that `rerun_ci` manages these variables for its special purpose.
This PR does not add a new feature, it just exposes an exisiting feature.

I did not make it public in the first place, because the resulting images are not well-formed.
However, I see 2 legit usecases:
1. `rerun_ci`
2. speedup different [build stages](https://github.com/ros-industrial/industrial_ci/pull/286#issuecomment-395332995] (together with #283)

It could even be pushed to a registry and reused as `DOCKER_IMAGE`, but I would not recommend it.

doc/index.rst
```diff
@@ -691,7 +706,11 @@ The cached images can be listed with
 
   $ rosrun industrial_ci rerun_ci --list
 
+Variables unavailable for rerun
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+`rerun_ci` determins Docker image name. Therefore `DOCKER_COMMIT`, `DOCKER_COMMIT_MSG` do not work.
```
ipa-mdl: Something like
> `rerun_ci` is managing `DOCKER_COMMIT` and `DOCKER_COMMIT_MSG` under the hood, so the user cannot set them.

?

130s:
Updated to address your comment @ipa-mdl.

> It could even be pushed to a registry and reused as DOCKER_IMAGE, but I would not recommend it.

Ha, my team has been happilly doing so for a few months :P Indeed built Docker images currently need some manual update (which I'll open tickets about once this PR finishes) but overall being able to automate Docker build as part of normal CI pipeline w/o maintaining Dockerfiles etc. is very helpful for our usecase.

ipa-mdl:
> Ha, my team has been happilly doing so for a few months :P Indeed built Docker images currently need some manual update

That's great to hear :)
But I still do not recommend it to the average user on Travis CI ;)

PR #352, opened 2019-01-27
 include EOL distros in rosdep update

ipa-mdl:
Addresses https://github.com/ros-infrastructure/rosdep/pull/647
The flag will only be set for EOL distros on systems which support this flag.

### DIFF
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -79,8 +79,17 @@ rosdep --version
 if ! [ -d /etc/ros/rosdep/sources.list.d ]; then
     sudo rosdep init
 fi
-ret_rosdep=1
-rosdep update || while [ $ret_rosdep != 0 ]; do sleep 1; rosdep update && ret_rosdep=0 || echo "rosdep update failed"; done
+
+update_opts=()
+case "$ROS_DISTRO" in
+"hydro"|"jade")
+    if rosdep update --help | grep -q -- --include-eol-distros; then
+      update_opts+=(--include-eol-distros)
+    fi
+    ;;
+esac
+
+ici_retry 2 rosdep update "${update_opts[@]}"
 
 ici_time_end  # setup_rosdep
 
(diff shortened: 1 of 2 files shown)


ipa-mdl:
Perhaps we want to inlcude an opt-out for users of EOL Docker images, which might not ship the new flag.

industrial_ci/src/tests/source_tests.sh
```diff
@@ -79,8 +79,17 @@ rosdep --version
 if ! [ -d /etc/ros/rosdep/sources.list.d ]; then
     sudo rosdep init
 fi
-ret_rosdep=1
-rosdep update || while [ $ret_rosdep != 0 ]; do sleep 1; rosdep update && ret_rosdep=0 || echo "rosdep update failed"; done
+
+update_opts=()
+case "$ROS_DISTRO" in
+"hydro"|"jade")
+    if rosdep update --help | grep -q -- --include-eol-distros; then
```
130s: I found this `if` clause a thoughtful addition :).

In `else` situation for this `if`, I assume `rosdep` is old so all rosdep key still get fetched and no error occurs, thus we don't need `else` handling?

industrial_ci/src/tests/source_tests.sh
```diff
@@ -79,8 +79,17 @@ rosdep --version
...
+      update_opts+=(--include-eol-distros)
+    fi
+    ;;
+esac
+
+ici_retry 2 rosdep update "${update_opts[@]}"
```
130s: Just curious, what happens without retrying here?

ipa-mdl: The retry was in there before..

ipa-mdl: There is a change though: now a failed update will stop the test and not just print a warning..

industrial_ci/src/util.sh
```diff
@@ -172,12 +172,16 @@ function ici_enforce_deprecated {
...
 
   for ((i=1;i<=tries;i++)); do
     "$@" && return 0
+    ret=$?
     sleep 1;
   done
-  error "'$*' failed $tries times"
+
+  ici_color_output ${ANSI_RED} "'$*' failed $tries times"
+  return $ret
```
130s: https://github.com/ros-industrial/industrial_ci/pull/352/files#r251236099

> There is a change though: now a failed update will stop the test and not just print a warning..

I might be misunderstood, but doesn't this line and the subsequent code prevent the ICI from failing?

130s: @ipa-mdl This is me just trying to stay on the safest side. If my understanding above is wrong then we can go ahead (if so please merge) as CI is passing.

ipa-mdl: This line returns a non-zero code, which is treated as error in the calling code.

PR #355, opened 2019-02-06
Run cleanup for apt resource when DOCKER_COMMIT is on.

130s:
## Problem aimed to be addressed
Using a Docker image where cleaning apt resource (? I don't know appropriate term for this) was not done, hash mismatch error can occur (as e.g. [this on askubuntu.com](https://askubuntu.com/questions/41605/trouble-downloading-packages-list-due-to-a-hash-sum-mismatch-error)).

## Changes this PR add
Run apt resource cleanup, only when DOCKER_COMMIT is set, which enables `docker commit` image and allows the users to reuse the image built during the run of industrial_ci.

### DIFF
--- a/industrial_ci/src/ci_main.sh
+++ b/industrial_ci/src/ci_main.sh
@@ -49,6 +49,10 @@ fi
 if [ "${AFTER_SCRIPT// }" != "" ]; then
   ici_time_start after_script
 
+  if [ -z "$DOCKER_COMMIT" ]; then
+      apt-get clean && rm -rf /var/lib/apt/lists/*
+  fi
+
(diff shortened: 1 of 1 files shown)


industrial_ci/src/ci_main.sh
```diff
@@ -49,6 +49,10 @@ fi
 if [ "${AFTER_SCRIPT// }" != "" ]; then
   ici_time_start after_script
 
+  if [ -z "$DOCKER_COMMIT" ]; then
```
miguelprada: Will only get executed if `AFTER_SCRIPT` is defined.

ipa-mdl: and `DOCKER_COMMIT` gets unset before the the script executions anyway.

PR #363, opened 2019-04-21
add options to ignore failing tests in prerelease

### LINKED ISSUE #362: Make tests optional for prerelease tests
`NOT_TEST_BUILD` and `NOT_TEST_DOWNSTREAM` could be used to set `ABORT_ON_TEST_FAILURE_UNDERLAY` and `ABORT_ON_TEST_FAILURE_OVERLAY`.

ipa-mdl:
closes #362 

@wxmerkt: With this fix you can set ABORT_ON_TEST_FAILURE_UNDERLAY=0 to make your test pass

### DIFF
--- a/industrial_ci/src/tests/ros_prerelease.sh
+++ b/industrial_ci/src/tests/ros_prerelease.sh
@@ -67,6 +67,8 @@ function run_in_prerelease_docker() {
     ici_run_cmd_in_docker $DIND_OPTS \
                           -v "$WORKSPACE:$WORKSPACE:rw" \
                           -e TRAVIS \
+                          -e "ABORT_ON_TEST_FAILURE_UNDERLAY=${ABORT_ON_TEST_FAILURE_UNDERLAY:-1}" \
+                          -e "ABORT_ON_TEST_FAILURE_OVERLAY=${ABORT_ON_TEST_FAILURE_OVERLAY:-1}" \
                           "industrial-ci/prerelease" \
                           "$@"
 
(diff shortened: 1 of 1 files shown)


ipa-mdl:
This should het merged into the new version.

> Can we document these 2 variables new to ICI?

I will add some line to the documentation of the prerelease tests.
They have been defined like this in `ros_buildfarm`.


> What happens when a test fails while neither of ABORT_ON_TEST_FAILURE_{Over, UNDER}LAY are passed (i.e. default behavior)? If this changes the default, that might be noted in the document?

TL;DR: This patch does not change the default of `industrial_ci`.

In `ros_buildfarm` these variables default to `0`, so failing tests are treated as errors.
In `industrial_ci` their default is `1`. This PR adds an opt-out for this choice.

130s:
Is there any update?

I might be willing to add documentation. As a "new user" to the recently added changes, I might be able to capture missing info well enough. If that's ok in general, I'm fine in moving forward in merging w/o including more documentation (in that case create follow-up ticket and assign me/anyone appropriate).

ipa-mdl:
> Is there any update?

This needs to be rebased, but it is prioritized

Some notes on these flags could be added here:
https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst#run-ros-prerelease-test
Just to comment on the non-default behavior, which is ([examle](https://travis-ci.com/ros-industrial/industrial_ci/jobs/270291114#L253))
> If you want the script to abort and return a non-zero return code
> you can set the environment variable ABORT_ON_TEST_FAILURE=1.
> You can also set ABORT_ON_TEST_FAILURE_UNDERLAY=1 or
> ABORT_ON_TEST_FAILURE_OVERLAY=1 to only affect a specific workspace.

Issue #364, opened 2019-05-03
setup_rosdep is unstable


ipa-mdl:
Currently, `rosdep update` seems to fail a lot with messages like this:
```
ERROR: unable to process source [https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/base.yaml]:
	<urlopen error ('_ssl.c:710: The handshake operation timed out',)> (https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/base.yaml)
```

At the moment, I would suggest to monitor this issue and restart jobs a lot :-/
We can't do much except for increasing the number of retries (currently just 2 tries in total).

ross-desmond:
I've added a ticket here as well for rosdep ci builds: https://github.com/ros-infrastructure/rosdep/issues/677

gavanderhoorn:
A bit of a "me too", but I just ran into this as well.

Do we have any mitigation in place?

ipa-mdl:
> Do we have any mitigation in place?

We could increase the [number of retries](https://github.com/ros-industrial/industrial_ci/blob/8e81016601c73d191f45f3d16367157c10eedf77/industrial_ci/src/tests/source_tests.sh#L92).
But I have restarted one of my builds >10 times now and still no luck.
It looks like `catkin_lint` has trouble downloading as well.

it works on my laptop, so I guess it it as Travis connectivity issue

gavanderhoorn:
> > Do we have any mitigation in place?
> 
> We could increase the [number of retries](https://github.com/ros-industrial/industrial_ci/blob/8e81016601c73d191f45f3d16367157c10eedf77/industrial_ci/src/tests/source_tests.sh#L92).
> But I have restarted one of my builds >10 times now and still no luck.

Yes, I've restarted a build a large nr of times as well and no change.

> it works on my laptop, so I guess it it as Travis connectivity issue

hm. That is interesting info actually.

ipa-rmb:
Same problem here, I sometimes need to restart so many times until something works.

130s:
This has been happening on Gitlab CI too :/


>> it works on my laptop, so I guess it it as Travis connectivity issue
>
> hm. That is interesting info actually.

Running Python version is different? Just a guess.

ross-desmond:
We (seems like a community effort) have posted support issues with travis and they have come up with a workaround as they investigate the failures.

https://www.traviscistatus.com/incidents/kyf149kl6bvp?u=3g7kt113nmgs

[...]

PR #365, opened 2019-05-06
use host network for travis

ipa-mdl:
possible fix for #364 

For Travis (=fresh VM per build job) this should be safe.
For Gitlab this might not be a good default.


ipa-mdl:
Looks like Travis cannot solve it that quick.

ipa-mdl:
Tests just passed.
It will solve al issues, but most.
Perhaps will get reverted later.

PR #369, opened 2019-05-14
Backport some colcon features

### LINKED ISSUE #258: ABI check feature not functional on ROS Indigo?
Reported in https://github.com/ros-drivers/openni2_camera/pull/75#issuecomment-351324425

### LINKED ISSUE #322: ABI check feature failing without obvious output but "ERROR: no symbols info in the ABI dump"
Of [this PR](https://github.com/ros-drivers/openni2_camera/pull/89), ABI check failed e.g. https://travis-ci.org/ros-drivers/openni2_camera/jobs/437001338
```
Preparing, please wait ...
ERROR: no symbols info in the ABI dump
Function abi_check_libopenni2_camera_nodelet returned with code '7' after 0 min 0 sec 
```
But other than that I failed to find the root cause of the error. There's a lot of warnings during the previous ABI check steps, but I don't see anything obviously failing.

ipa-mdl:
Partial backport of #361 

* align the code/behavior between both version
* shellcheck :)
*  fix #258, fix #322 
* backport of CMAKE_ARGS (highly perferred over `CATKIN_CONFIG`)

In addition it should simplify the review.

# CoS (Condition of Satisfaction, suggested by @130s)
- Code review.
  - [x] @miguelprada 
  - [x] @130s
- (Locally) verify the change.
  - [x] @130s ABI error https://github.com/ros-industrial/industrial_ci/issues/258 -> Ongoing in https://github.com/130s/openni2_camera/pull/2


ipa-mdl:
Due to the many renames it is best to review each commit individually.

ipa-mdl:
@130s: Can you test /review it as well?
These fixes will end up in master/legacy.

130s:
# CoS (Condition of Satisfaction)
- [x] @130s Code review.
- (Locally) verify the change.
  - [ ] @130s ABI error https://github.com/ros-industrial/industrial_ci/issues/258 -> Ongoing in https://github.com/130s/openni2_camera/pull/2

ipa-mdl:
Indigo is now [EOL](https://discourse.ros.org/t/indigo-igloo-officially-eol/9211).

ipa-mdl:
@130s: Can we merge this? Indigo jobs started to fail as a result of (https://github.com/ros/rosdistro/pull/21284)

130s:
I confirmed ABI feature issue is fixed with this PR https://github.com/130s/openni2_camera/pull/2#issuecomment-494781550

Thanks for the great changes and backporting them!

130s:
Tagged [0.8.0](https://github.com/ros-industrial/industrial_ci/releases/tag/0.8.0) mainly in a desire to clearly label a change where a ROS distro is now EOLed but still supported by ici.

miguelprada:
FYI, I just found a small regression introduced in this PR related to an arguably non-standard use-case (i.e. see #232).

Previously the `export` command [here](https://github.com/ros-industrial/industrial_ci/blob/f3edb6d71307c5ce4abd918050eb57ac0fbc4ea0/industrial_ci/src/tests/source_tests.sh#L183) was [masking the (error) return value](https://github.com/koalaman/shellcheck/wiki/SC2155) of `catkin_topological_order`, and the execution continued even with an empty workspace. After bdd783f, this error is caught and the job fails.

I don't think this should be acted upon, since that use case is not very common (and I'm no longer doing that myself). Furthermore, I'm not sure that _trick_ would make sense anymore when #361 is merged. I wanted to leave this comment here just for future reference, in case I stumble into this ever again.

ipa-mdl:
Yes, this slipped through.
We could fix this on demand.
With #361 this problem does not exist anymore, you can just specify `TARGET_WORKSPACE`.
In addition `colcon` works with empty source spaces as well.

PR #373, opened 2019-06-04
Key change and snapshots.ros.org support

ipa-mdl:
supersedes #372, supersedes #370  
~~Not sure if the logic is correct, https://github.com/osrf/docker_images/pull/269 will clarify this..~~

### DIFF
--- a/industrial_ci/src/docker.sh
+++ b/industrial_ci/src/docker.sh
@@ -220,6 +220,14 @@ EOF
 }
 
 function ici_generate_default_dockerfile() {
+  local keycmd
+
+  if [ -n "${APTKEY_STORE_HTTPS}" ]; then
+    keycmd="wget '${APTKEY_STORE_HTTPS}' -O - | apt-key add -"
+  else
+    keycmd="apt-key adv --keyserver '${APTKEY_STORE_SKS}' --recv-key '${HASHKEY_SKS}'"
+  fi
+
   cat <<EOF
 FROM $DOCKER_BASE_IMAGE
 
@@ -231,8 +239,7 @@ RUN apt-get update -qq \
     && apt-get -qq install --no-install-recommends -y apt-utils gnupg wget ca-certificates lsb-release
 
 RUN echo "deb ${ROS_REPOSITORY_PATH} \$(lsb_release -sc) main" > /etc/apt/sources.list.d/ros-latest.list
-RUN apt-key adv --keyserver "${APTKEY_STORE_SKS}" --recv-key "${HASHKEY_SKS}" \
-    || { wget "${APTKEY_STORE_HTTPS}" -O - | apt-key add -; }
+RUN for i in 1 2 3; do { $keycmd; } &&  break || sleep 1; done
 
 RUN sed -i "/^# deb.*multiverse/ s/^# //" /etc/apt/sources.list \
     && apt-get update -qq \
--- a/industrial_ci/src/env.sh
+++ b/industrial_ci/src/env.sh
@@ -25,7 +25,7 @@ if [ ! "$ROS_PARALLEL_TEST_JOBS" ]; then export ROS_PARALLEL_TEST_JOBS="$ROS_PAR
 # .rosintall file name
 if [ ! "$ROSINSTALL_FILENAME" ]; then export ROSINSTALL_FILENAME=".travis.rosinstall"; fi
 # For apt key stores
-if [ ! "$APTKEY_STORE_SKS" ]; then export APTKEY_STORE_SKS="hkp://ha.pool.sks-keyservers.net"; fi  # Export a variable for SKS URL for break-testing purpose.
+if [ ! "$APTKEY_STORE_SKS" ]; then export APTKEY_STORE_SKS="hkp://keyserver.ubuntu.com:80"; fi  # Export a variable for SKS URL for break-testing purpose.
 if [ ! "$HASHKEY_SKS" ]; then export HASHKEY_SKS="0xB01FA116"; fi
 if [ "$USE_DEB" ]; then  # USE_DEB is deprecated. See https://github.com/ros-industrial/industrial_ci/pull/47#discussion_r64882878 for the discussion.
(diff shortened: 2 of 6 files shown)


industrial_ci/src/docker.sh
```diff
@@ -231,8 +231,8 @@ RUN apt-get update -qq \
     && apt-get -qq install --no-install-recommends -y apt-utils gnupg wget ca-certificates lsb-release
 
 RUN echo "deb ${ROS_REPOSITORY_PATH} \$(lsb_release -sc) main" > /etc/apt/sources.list.d/ros-latest.list
-RUN apt-key adv --keyserver "${APTKEY_STORE_SKS}" --recv-key "${HASHKEY_SKS}" \
-    || { wget "${APTKEY_STORE_HTTPS}" -O - | apt-key add -; }
+RUN  { [ -n "${APTKEY_STORE_HTTPS}" ] && wget "${APTKEY_STORE_HTTPS}" -O - | apt-key add -; } \
+    || apt-key adv --keyserver "${APTKEY_STORE_SKS}" --recv-key "${HASHKEY_SKS}"
```
miguelprada: If I understand this correctly, key will be downloaded with `wget` as the first option if `APTKEY_STORE_HTTPS` is provided. While before, the `wget` method was only used as a fallback option in case `hkp` failed.

I know I've had issues with `hkp` in the past, but I'm not sure whether these were related to our corporate network or whether these just happen randomly to everybody. Shouldn't a fallback `wget` attempt with a default URL be kept, just in case?

industrial_ci/src/env.sh
```diff
@@ -109,6 +92,31 @@ if [ -z "$OS_CODE_NAME" ]; then
...
+if [ "$ROS_VERSION_EOL" = true ]; then
+  if [ ! "$HASHKEY_SKS" ]; then export HASHKEY_SKS="AD19BAB3CBF125EA"; fi
+  ROS_REPO_SERVER="http://snapshots.ros.org"
+else
+  if [ ! "$HASHKEY_SKS" ]; then export HASHKEY_SKS="C1CF6E31E6BADE8868B172B4F42ED6FBAB17C654"; fi
+  ROS_REPO_SERVER="http://packages.ros.org"
+fi
+
+# If not specified, use ROS Shadow repository http://wiki.ros.org/ShadowRepository
+if [ ! "$ROS_REPOSITORY_PATH" ]; then
```
miguelprada: Should there be any special handling of the EOL distro case here? Is it reasonable to expect `ros` and `ros-testing` to be present in the snapshots server as well? I guess it may be, since there are still EOL'd distro packages in the current `ros-shadow-fixed` repo.

industrial_ci/src/docker.sh
ipa-mdl: > Shouldn't a fallback wget attempt with a default URL be kept, just in case?

If you have a default URL for the snapshots repo, I can adjust it accordingly.
BTW, the new behavior is the same as in #361.

industrial_ci/src/env.sh
ipa-mdl: I don't know yet. I assume that all of them will get renamed (the repos get signed with the new key anyway).

ipa-mdl: Ah, the layout changes completely, e.g. http://snapshots.ros.org/cturtle/final/ubuntu/
I will adapt the logic.

industrial_ci/src/docker.sh
miguelprada: > If you have a default URL for the snapshots repo, I can adjust it accordingly.

I don't.

> BTW, the new behavior is the same as in #361.

Didn't notice that. It makes sense to keep it simple. I guess `APTKEY_STORE_HTTPS` usage should be added to the #361 docs at some point, so anybody having issues with hkp can know it exists and go figure out the right URL for them.

ipa-mdl: I have implemented a retry mechanism

industrial_ci/src/env.sh
ipa-mdl: > I will adapt the logic.

Done.
In addition non-final snapshot can be used as well

ipa-mdl:
> From my side, it's ok to merge once the packages with the new key are released and the tests pass.
:+1: 

I will rerun the tests before the merge.
In addition I have modified the logic slighlty: the final snapshot gets enforced now.

ipa-mdl:
> In addition I have modified the logic slighlty: the final snapshot gets enforced now.

And added some warnings

miguelprada:
It's not quite true that ROS_REPO is ignored for EOL'd distros, is it? Can't a dated snapshot still be chosen using it?

ipa-mdl:
> It's not quite true that ROS_REPO is ignored for EOL'd distros, is it? Can't a dated snapshot still be chosen using it?

That's right, but in this case they warning won't get printed.

(Not `ROS_REPO` gets ignored, but e.g. `ROS_REPO=ros`)

ipa-mdl:
Most tests pass.
I had to remove the 3 tests withr wrong key server entiries (-> behavior change).

Two jobs are still failing:
1.  `DOCKER_IMAGE='arm32v7/ros:indigo-ros-core' ADDITIONAL_DEBS="build-essential python-catkin-tools python-pip" INJECT_QEMU=arm BEFORE_SCRIPT='[[ $(uname -p) == armv7l ]]'`
2. `ROS_DISTRO=melodic OS_NAME=debian OS_CODE_NAME=stretch AFTER_SCRIPT='grep -q ID=debian /etc/os-release && grep -q VERSION_ID=\"9\" /etc/os-release'`

They will pass again after the Docker images will have been rebuilt (https://github.com/osrf/docker_images/pull/275) and the melodic packages have been rebuilt for Debian stretch.

https://github.com/ros-industrial/industrial_ci/pull/373/commits/ae0474fb5169489042460328b89eef004f3b0a0f must get reverted as soon as indigo repository for trusty was synced to the snapshots server.

ipa-mdl:
Synced to master and created PR to use the snapshot repo for indigo (#378)

PR #378, opened 2019-06-08
Use final snaptshots for indigo and lunar

ipa-mdl:
This reverts commits ae0474fb5169489042460328b89eef004f3b0a0f and d56e1dc1f9fa32c1d5ef7dfdae529dcbec2eb4ce.
In addition, I updated the EOL status of lunar in the documentation.


ipa-mdl:
I had to fix a regression of #373 as well

ipa-mdl:
fix shellcheck warning

ipa-mdl:
> I don't quite grasp the changes to the prerelease tests,

The new EOL system overwrites the repository path, but the tests uses a distro (was: `kinetic`) that is independent of `ROS_DISTRO`, just be able to use `catkin_test_results`.
This lead to using the snapshot repo for a non-EOL distro, which fails.
I figured that we can just use an official image for this task.
In addition I took the chance to use `melodic` and the bundled Docker version.

`ROS_DISTRO=''` just suppresses the warning :)

Issue #376, opened 2019-06-07
Have rosinstall use self-signed certificate


kjeremy:
My repo `ros_pkg_b` depends on `ros_pkg_a` which I have specified in my `.rosinstall` file:

```yaml
- git:
    local-name: pkg_a
    uri: https://ago-git.us.ara.com/jkolb/ros_pkg_a.git
    version: master
```

That git repo uses a self-signed cert and during build I get:

> >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> Starting function 'setup_rosws'
> Writing /root/catkin_ws/src/.rosinstall
> 
> update complete.
>      Performing actions: 
> 
>      Add new elements:
>   pkg_a
> 
> Config changed, maybe you need run wstool update to update SCM entries.
> Overwriting /root/catkin_ws/src/.rosinstall
> 
> update complete.
> HOSTNAME=f87373937503
> OS_NAME=ubuntu
> TARGET_REPO_PATH=/root/src/ros_pkg_b
> TERM=dumb
> ROS_ROOT=/opt/ros/kinetic/share/ros
> OS_CODE_NAME=xenial
> ROS_PACKAGE_PATH=/opt/ros/kinetic/share
> ROS_MASTER_URI=http://localhost:11311
> ROSINSTALL_FILENAME=.rosinstall
> ROS_VERSION=1
> LC_ALL=C.UTF-8
> ICI_SRC_PATH=/root/ici
> LD_LIBRARY_PATH=/opt/ros/kinetic/lib
> CATKIN_PARALLEL_TEST_JOBS=-p4
> APTKEY_STORE_SKS=hkp://ha.pool.sks-keyservers.net
> IN_DOCKER=true
> PATH=/opt/ros/kinetic/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
> ROS_PARALLEL_JOBS=-j8
> UPSTREAM_WORKSPACE=file
> _DO_NOT_FOLD=true
> TARGET_REPO_NAME=ros_pkg_b
> PWD=/
> ROS_PARALLEL_TEST_JOBS=-j8
> LANG=C.UTF-8
> HASHKEY_SKS=0xB01FA116
> ROSLISP_PACKAGE_DIRECTORIES=
> DOCKER_BASE_IMAGE=ubuntu:xenial
[...]

PR #381, opened 2019-06-10
Documentation for gitlab with self-signed certs

kjeremy:
Document some things discovered in #376

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -288,6 +288,20 @@ If your Gitlab CI jobs require access to private repos, additional settings are 
       # gitlab.com:22 SSH-2.0-OpenSSH_7.2p2 Ubuntu-4ubuntu2.2
 
 #. Add a public key (reference for `Gitlab <https://docs.gitlab.com/ce/ssh/README.html#deploy-keys>`__ and for `GitHub <https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys>`__) to the private repos your CI jobs accesses. You may need to ask the admin of that repo.
+#. Make sure that `TMPDIR` is set in your `.gitlab-ci.yml` file so that the SSH agent forwards properly ::
+    
+    # The docker runner does not expose /tmp to the docker-in-docker service
+    # This config ensures that the temp folder is located inside the project directory (e.g. for prerelease tests or SSH agent forwarding)
+    variables:
+      TMPDIR: "${CI_PROJECT_DIR}.tmp"
+#. If using a self-signed certificate you may need to make docker aware of your CA file ::
+    
+    kinetic:
+      script:
+        # Run the gitlab script. Since we are using a self-signed certificate we need to tell docker to
+        # mount the CA file and then pass it to git via the GIT_SSL_CAINFO environment variable.
+        - .industrial_ci/gitlab.sh DOCKER_RUN_OPTS="-v $CI_SERVER_TLS_CA_FILE:$CI_SERVER_TLS_CA_FILE -e GIT_SSL_CAINFO=$CI_SERVER_TLS_CA_FILE"
+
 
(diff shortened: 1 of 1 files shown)


doc/index.rst
```diff
@@ -288,6 +288,20 @@ If your Gitlab CI jobs require access to private repos, additional settings are
       # gitlab.com:22 SSH-2.0-OpenSSH_7.2p2 Ubuntu-4ubuntu2.2
 
 #. Add a public key (reference for `Gitlab <https://docs.gitlab.com/ce/ssh/README.html#deploy-keys>`__ and for `GitHub <https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys>`__) to the private repos your CI jobs accesses. You may need to ask the admin of that repo.
+#. Make sure that `TMPDIR` is set in your `.gitlab-ci.yml` file so that the SSH agent forwards properly ::
```
ipa-mdl: Please make this conditional, e.g. start with `If you are using Docker-in-Docker, .."

doc/index.rst
```diff
@@ -288,6 +288,20 @@ If your Gitlab CI jobs require access to private repos, additional settings are
...
+    
+    # The docker runner does not expose /tmp to the docker-in-docker service
+    # This config ensures that the temp folder is located inside the project directory (e.g. for prerelease tests or SSH agent forwarding)
+    variables:
+      TMPDIR: "${CI_PROJECT_DIR}.tmp"
+#. If using a self-signed certificate you may need to make docker aware of your CA file ::
```
ipa-mdl: This is not used by SSH, so no need to set it (for this purpose).

doc/index.rst
```diff
@@ -288,6 +288,20 @@ If your Gitlab CI jobs require access to private repos, additional settings are
...
+    
+    kinetic:
+      script:
+        # Run the gitlab script. Since we are using a self-signed certificate we need to tell docker to
+        # mount the CA file and then pass it to git via the GIT_SSL_CAINFO environment variable.
+        - .industrial_ci/gitlab.sh DOCKER_RUN_OPTS="-v $CI_SERVER_TLS_CA_FILE:$CI_SERVER_TLS_CA_FILE -e GIT_SSL_CAINFO=$CI_SERVER_TLS_CA_FILE"
```
ipa-mdl: I would prefer to set this automatically in `gitlab.sh`

doc/index.rst
kjeremy: I took that directly from here: https://github.com/ros-industrial/industrial_ci/blob/legacy/.gitlab-ci.yml#L7-L10

kjeremy: I could not get `rosinstall` to clone until I set `TMPDIR` so I assumed that it is required for SSH agent.

doc/index.rst
kjeremy: Me too but I'm struggling modifying the `DOCKER_RUN_OPTS` argument fo `gitlab.sh` and injecting it into the environment for `ci_main.sh`

doc/index.rst
ipa-mdl: I was referring to the SSL settings below

doc/index.rst
ipa-mdl: It looks like this change will break git access to all other servers: https://gitlab.com/ipa-mdl/industrial_ci/-/jobs/229408390, there must be a better way

kjeremy: @ipa-mdl Passing the runner's certs works for my case see https://github.com/ros-industrial/industrial_ci/issues/376#issuecomment-501040493

ipa-mdl: Passing the certificates should not affect SSH. It will still work if you remove these settings.

PR #389, opened 2019-06-27
Lunar is EOL

ipa-mdl:


industrial_ci/src/env.sh
```diff
@@ -101,7 +105,7 @@ function use_snapshot() {
 }
 
 function use_repo_or_final_snapshot() {
-    if [ "$ROS_VERSION_EOL" = true ] && [ "$ROS_DISTRO" != "indigo" ]; then
+    if [ "$ROS_VERSION_EOL" = true ] && [ "$ROS_DISTRO" != "indigo" ] && [ "$ROS_DISTRO" != "lunar" ]; then
```
miguelprada: Sorry for being so late to the party. Maybe I'm not understanding this properly, but aren't these the `final` snapsots for [indigo](http://snapshots.ros.org/indigo/final/) and [lunar](http://snapshots.ros.org/lunar/final/)?

ipa-mdl: They have just been added (https://discourse.ros.org/t/final-lunar-sync/9629/5).
We can remove these clauses, but its not urgent for now.

ipa-mdl:
snapshot will be used in #378

PR #392, opened 2019-07-03
Revert switching to snapshots

ipa-mdl:
Reverts #378 partially, because the final snapshots have been [rolled back](https://discourse.ros.org/t/final-lunar-sync/9629/7).


nuclearsandwich:
Indigo and Lunar final snapshots have been corrected and deployed.

ipa-mdl:
@nuclearsandwich: Thanks! I will test it later :)

PR #393, opened 2019-07-11
Revert revert revert using snapshots

ipa-mdl:
reverts #392


gavanderhoorn:
> Great PR title, BTW!

I don't know, could've used a few more `revert`s.

ipa-mdl:
> I don't know, could've used a few more reverts.

Only if needed ;)

PR #395, opened 2019-07-14
switch to default Travis notification settings

ipa-mdl:
https://docs.travis-ci.com/user/notifications/#default-notification-settings:

> By default, email notifications are sent to the committer and the commit author when they are members of the repository, that is they have
> - push or admin permissions for public repositories.
> - pull, push or admin permissions for private repositories.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,6 @@ services:
   - docker
 # Force travis to use its minimal image with default Python settings
 language: generic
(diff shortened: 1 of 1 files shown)


ipa-mdl:
> My guess is once rebased to the origin's branch to include #394, CI will pass?

Yes, that's why I canceled the build.

Issue #387, opened 2019-06-17
As an ICI develop / ICI user repo maintainer, I want the purpose of branch(es) to be clarified


130s:
# Problem
I found the default branch on GitHub is called `legacy` now. My local repo, which I hadn't updated for months (bad), tells me the default branch used to be `master`. I can't find when this happened nor the decision to do so (aplogies if there's clear notice..). Branch name and usage change, esp. when it's not announced/shared, can confuse developers and often cause unnecessary extra work.

# Possible solution
- Document the usage of branch(es).
- Since the new branching is already in effect, I don't think we need to revert the usage.

130s:
> I can't find when this happened nor the decision to do so (apologies if there's clear notice..). 

Found it. https://github.com/ros-industrial/industrial_ci/issues/358#issuecomment-496425447, I'm sorry again (apparently I'm not catching up with the development on this repo).

That said, the change needs documented at least.

ipa-mdl:
The transition to `legacy` has been [dicussed](https://github.com/ros-industrial/industrial_ci/issues/35) and [announced](https://discourse.ros.org/t/industrial-ci-migration-needed-for-recent-and-future-changes/9462).

For now, all three `branches` point to the same commit.
If you [agree](https://github.com/ros-industrial/industrial_ci/issues/358#issuecomment-501216503):
* `colcon` will get deleted
* `legacy` will point to the old version (default for some time)
* `master` will become the colcon/ROS2-ready version (#361) and eventually will become the default again

`legacy` and `master` will point to each other.

130s:
I see.

> - `legacy` will point to the old version (default for some time)

As it's [announced](https://discourse.ros.org/t/industrial-ci-migration-needed-for-recent-and-future-changes/9462) already I won't request change. But "legacy" is relative term so I'm afraid we might come up with another legacy in the future. I'd go with version number (latest ATM is 0.8.0), although it won't be a big deal at all if `legacy` branch is planned to be deleted at some point soon.

ipa-mdl:
> if legacy branch is planned to be deleted at some point soon

It should not be deleted.
It will be kept as-is without new features.
Issues will be fixed if possible with limited effort, e.g., if distros become EOL.

> But "legacy" is relative term so I'm afraid we might come up with another legacy in the future.

The name refers to legacy systems, which will not be supported by the master branch anymore.

ipa-mdl:
> That said, the change needs documented at least.

#397 will add a hint

[...]

PR #397, opened 2019-07-14
deprecate legacy branch

ipa-mdl:
~~closes~~ #387


ipa-mdl:
This will be accompanied by a ROS discourse post

130s:
> This PR looks

Nevermind mOm

130s:
Also, I think somewhere we should clarify "support" plan (answering questions, future development) assuming `master` will be the default and the maintenance focus. So in`legacy` branch at the top of the doc we may want to put a notice.

ipa-mdl:
>  I think somewhere we should clarify "support" plan (answering questions, future development)

Good point!
I would describe it like this:

* no future development for `legacy`, bug fixing will continue as long as feasible
* support will be minimal: bug fixing and support for features that can only be found in `legacy`.

I will update the PR to reflect at least the first point

PR #400, opened 2019-08-05
Add IMMEDIATE_TEST_OUTPUT option

agutenkunst:
This would enable setting the interleave option.

Reasoning:
If a gtest gets stuck and travis will timeout eventually the output for debugging the reason will be lost.
This PR allows setting the interleave option for continous output during the test execution.

Note that this is not an issue with rostests since they have their own timeout.


ipa-mdl:
I understand the need for this flag.
However, the `legacy` branch is more or less in maintenance-only mode.
New developments should be targeted at the new version (#361, hopefully gets merged in the next days).
It should be quite easy to implement the same in the new structure :)

agutenkunst:
I will adapt as soon as the new version is merged...

ipa-mdl:
#361 got merged..

agutenkunst:
Ready. Please remove legay label.

agutenkunst:
@ipa-mdl I am not sure what is the desired feature for colon users. Having no option to set the output behavior could be undesired...

ipa-mdl:
> I am not sure what is the desired feature for colon users. Having no option to set the output behavior could be undesired...

Yes. That's why I want to have colcon-support as well.
If `INTERLEAVE_OUTPUT_TESTS` is `true`, `console_direct+` should be passed instead of `console_cohesion+`.
I am not sure if we have to  switch to the sequential executor...

We could even disuss (later?) to make `INTERLEAVE_OUTPUT_TESTS=true` the default.

> How about introducing an option `DIRECT_TEST_OUTPUT` which works for both catkin_tools and colon?

The name `DIRECT_TEST_OUTPUT` would work for me as well.
If we decide to make this behavior the default, we could even go with `TESTS_OUTPUT_COHESION` (or alike) as opt-out.

agutenkunst:
As far as I understand the problem is that catkin_tools offers 2 options (interleave, nothing) while colcon offers 3 options (direct, cohesion, nothing) correct?
What would you consider to be the desired default?

Actually I needed the direct output because a test got stuck beyond the travis timeout and I wanted to have more insight. Could we run into mixing test outputs due to parallel test execution if the default is direct output?

ipa-mdl:
> As far as I understand the problem is that catkin_tools offers 2 options (interleave, nothing) while colcon offers 3 options (direct, cohesion, nothing) correct?

colcon's default: nothing (really no useful output at all) 
catkin-tools' default: non-interleaved

I changed colcon to use `console_cohesion` to have the same behavior as with catkin-tools.
(non-interleaved == console_cohesion, interleaved == console_direct).

> What would you consider to be the desired default?

Honestly, I don't know.
For CI, interleaved mode makes sense to spot the errors easier.
But this Travis timeout makes things more complicated.

Could try your faling build with a custom [timeout](http://manpages.ubuntu.com/manpages/trusty/man1/timeout.1.html)?

``` yaml
script:
   - timeout --preserve-status 45m ./.industrial_ci/travis.sh
```
> Could we run into mixing test outputs due to parallel test execution if the default is direct output?

Yes. With colcon and catkin-tools.

agutenkunst:
So I ran some checks if this works as expected. (Note: Before renaming the variable.)
See the various build of https://github.com/agutenkunst/art_interleave_option/branches for the effect.

With the feature
https://github.com/agutenkunst/art_interleave_option/blob/26a7fd5f2849b4741a5562580bf7d7e13c5a3358/test/stuck_test.cpp#L7
gets printed.

> Could try your faling build with a custom [timeout](http://manpages.ubuntu.com/manpages/trusty/man1/timeout.1.html)?
> 
> ```yaml
> script:
>    - timeout --preserve-status 45m ./.industrial_ci/travis.sh
> ```

This results in the same unsatisfying output
```
Subprocess  > art_interleave_option:make `cd /root/target_ws/build/art_interleave_option; catkin build --get-env art_interleave_option | catkin env -si  /usr/bin/make run_tests --jobserver-fds=6,7 -j; cd -`
...industrial_ci is still running...
The command "timeout --preserve-status 10m .industrial_ci/travis.sh" exited with 143.
Done. Your build exited with 1.
```
If you are onboard with **IMMEDIATE_TEST_OUTPUT** I suggest a merge + close.

industrial_ci/src/builders/colcon.sh
```diff
@@ -34,7 +34,12 @@ function builder_run_build {
 function builder_run_tests {
     local extend=$1; shift
     local ws=$1; shift
-    ici_exec_in_workspace "$extend" "$ws" colcon test --event-handlers "${_colcon_event_handlers[@]}" console_cohesion+
+    if [ "$IMMEDIATE_TEST_OUTPUT" == true ]; then
+        _colcon_event_handlers+=("console_direct+")
```
ipa-mdl: This variable is used by subsequent builds as well and should not be modified.

agutenkunst: Yep sry did not notice.

agutenkunst:
Ready for merge

PR #401, opened 2019-08-06
Minor fix in doc

agutenkunst:
Seems like a copy-paste error to me.


ipa-mdl:
This seems to be a duplicate of #399. What changed?

agutenkunst:
I guess I had one for master and one for legacy... Nevermind. Close.

PR #402, opened 2019-08-15
 implemented clang-tidy check

### LINKED ISSUE #398: clang-tidy support?
Sort-of a follow-up to #316.

I've seen some users of `industrial_ci` manually add a clang-tidy run to their CI setup, or decide to add it to the `CMakeLists.txt` itself.

It could be nice if `industrial_ci` could do the hard work and add optional tidy support, similar to how clang-format is supported.

One big difference between tidy and format is that the latter is typically used to change things, which may not fit well with the CI theme in which `industrial_ci` fits.

ipa-mdl:
fixes #398 
needs #361

Can be activated with `CLANG_TIDY=true` and will only for errors.
With `CLANG_TIDY=pedantic`, it will fail for warnings as well.

The output is grouped by build spaces, which corresponds to packages except for `BUILDER=catkin_make`.

ToDo:
- [x] add proper tests
- [x] add documentation


ipa-mdl:
only https://github.com/ros-industrial/industrial_ci/pull/402/commits/8c0152662a3191be24c450419c41fe38d3787711 is part of this PR

ipa-mdl:
@gavanderhoorn: Did you have a chance to test this?

PR #403, opened 2019-08-18
interleave test output

ipa-mdl:
#400 for catkin_tools and colcon, but without opt-out
~~I am not yet sure how to provide an opt-out for colcon (`console_cohesion+`?).~~


ipa-mdl:
I guess this should be merged with #400

ipa-mdl:
included in #400

Issue #406, opened 2019-09-04
Failing jobs due to timeouts - Travis issue, or something else?


gavanderhoorn:
Not sure this is something that we can fix on this side, but I have quite a few jobs that have run to the max time allowed for open-source builds, such as [this one](https://travis-ci.org/ros-industrial/motoman/builds/580651197).

[This one](https://travis-ci.org/ros-industrial/motoman/builds/580723642) got very close, but didn't make it either.

[Here](https://travis-ci.org/ros-industrial/motoman/jobs/580723644), the `rosdep install` step takes 38 minutes in total.

Not sure why builds that used to take 20 minutes are suddenly taking 2.5x as long.

ipa-mdl:
Some weeks ago I saw these delays as well.
According to my tests, the apt download speed just became extremely slow.
With Travis and locally.

One workaround is to switch to a different mirror or the mirror protocol, but it turned out to be quiet unstable.

gavanderhoorn:
Let's see whether it wants to play nice again today.

I didn't really believe we could do something about `apt` being slow here, but at least wanted to track it here.

gavanderhoorn:
Perhaps Travis doesn't like me any more.

Jobs are still timing out everywhere (or not even starting), but the status pages of the service don't show anything out of the ordinary, nor are there massive nrs of tweets with complaints.

musamarcusso:
I had the exact same issue some builds ago: https://travis-ci.org/boschresearch/pcg_gazebo_pkgs/jobs/579431706

I imagine something was happening because I was trying to install a dependency `apt install libspatialindex-dev` using the `BEFORE_SCRIPT` input and after I removed it the builds came back to normal.
I wonder if this has something to do with some APT packages having to be whitelisted for Travis CI (like in the case of `libspatialindex` [here](https://github.com/travis-ci/travis-ci/issues/4560)).
By the way, I still cannot run `apt` in the `BEFORE_SCRIPT`, I just removed it for the time being so my builds don't timeout all the time.

gavanderhoorn:
@musamarcusso: thanks for the input.

Unfortunately I haven't changed anything in any of the builds that are timing out. It's just a regular CI run for a PR that adds some packages (which don't introduce new dependencies).

musamarcusso:
Interesting. I will try to add the package again and see what happens, the timeout issues lasted a couple of days and I eventually gave up.

wxmerkt:
[...]

PR #408, opened 2019-09-06
use Ubuntu mirror for Travis

ipa-mdl:
This is a hotfix for #406  to use the EC2 mirror on Travis.
@gavanderhoorn: please test

The hard-coded mirror is set on some (but not all) Travis machines as well.
For other/custom CI systems, this mirror might not work properly, so it gets only set for Travis.
The introduced variable `_UBUNTU_MIRROR` is not public and therefore not supported apart from internal use.

A fix for `master` is a little more complex and will follow later.

### DIFF
--- a/industrial_ci/src/docker.sh
+++ b/industrial_ci/src/docker.sh
@@ -236,6 +236,7 @@ FROM $DOCKER_BASE_IMAGE
 ENV ROS_DISTRO $ROS_DISTRO
 
 RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
Apologies. I've not yet had an opportunity to test this.

ipa-mdl:
I will close this for now.
I saw that Travis started to provide an apt cache, which we might use as well from Docker.

gavanderhoorn:
Apologies for not having been able to test this.

> I saw that Travis started to provide an apt cache, which we might use as well from Docker.

that would be great.

I'm actually surprised that Travis didn't start providing this earlier.

Or did they always have this and we're just late with noticing it?

ipa-mdl:
> Or did they always have this and we're just late with noticing it?

I just saw some build errors because the cache could not be reached..
It is not set-up for all builds: https://github.com/travis-ci/travis-build/pull/1606

PR #410, opened 2019-09-17
Add set -a flag to export all vars/functions

agutenkunst:
I noticed that if I want to use e.g. ici_exec_in_workspace in an after script I have to 
```
source "${ICI_SRC_PATH}/workspace.sh
```
but I figured that it would be nice to have all the functions within the after script.

Open for discussion.

### DIFF
--- a/industrial_ci/src/ci_main.sh
+++ b/industrial_ci/src/ci_main.sh
@@ -21,6 +21,7 @@
 ## See ./README.rst for the detailed usage.
 
 set -e # exit script on errors
+set -a # export each variable / function (enables usage in after_script)
 if [ "$DEBUG_BASH" ]; then set -x; fi # print trace if DEBUG
 
 # Define some env vars that need to come earlier than util.sh


ipa-mdl:
Actually, this should be supported in the `master` version:

```
industrial_ci/scripts/run_ci AFTER_SCRIPT="ici_exec_in_workspace /opt/ros/melodic \$target_ws/install catkin_topological_order" ROS_DISTRO=melodic
```

However, only selected variables get exported/exposed.

agutenkunst:
Hi @ipa-mdl ,
I am currently adding a coverage check to one of our repos. In order to to so I use a extra script 
in AFTER_SCRIPT. See https://github.com/PilzDE/pilz_robots/compare/feature/coverage_ci


This works only with 
```
source "${ICI_SRC_PATH}/workspace.sh"
```
in the beginning of the script, yet I am not so sure why...

If I don't source the workspace file I get a
```
./.coverage.sh: line 14: ici_exec_in_workspace: command not found
```
(See https://github.com/agutenkunst/pilz_robots/commit/051cdfe71c907a7575add9ce25103690872e4055 with https://travis-ci.org/agutenkunst/pilz_robots/jobs/586091313)

ipa-mdl:
I see.
This will not work with exposing all **variables** either.

To be able to use the ici functions, just source your script: `AFTER_SCRIPT="source ./.coverage.sh"`
This way it will be executed in a subshell and not in a new shell.
Please note: This is a new feature of `master`, and does not work with `legacy`.

ipa-mdl:
@agutenkunst: Did sourcing the file work for you?

PR #411, opened 2019-09-18
pass PATH in run_travis

### LINKED ISSUE #409: Running locally fails if docker was installed via snap?
Running industrial_ci locally failed with
```
docker command not found
```
when I had docker installed via snap. Via apt works fine now...

Can anyone reproduce?

ipa-mdl:
fixes #409, should get backported to `legacy`.

@agutenkunst: please test


agutenkunst:
Luckily I had a fresh system around...

Checked with Ubuntu 18.04 and Docker via snap. 

**Works!** Ready to merge...

PR #419, opened 2019-10-12
filter travis yaml like travis-build

### LINKED ISSUE #417: Backslash notation in .travis.yml not working with local run_travis
I needed https://github.com/PilzDE/pilz_robots/commit/0f0a2810fb813be79de6403053b4638ffec7bb8e
because when running locally via `run_travis` only the first variable would be set. And therefore e.g. there would be no AFTER_SCRIPT

`yamllint` indicates that both versions should be valid.

@ipa-mdl you opinion?

ipa-mdl:
fixes #417 

I am not 100% sure about the regex..
@agutenkunst: please test


agutenkunst:
@ipa-mdl tested: looks good!

ipa-mdl:
I now had a closer look at this regex. It is rather fragile, it might be hard to maintain.
Any other ideas?

One option would be to split the string with [shlex](https://docs.python.org/2.7/library/shlex.html) and trim the elements, but this would be just a hack as well :-/

ipa-mdl:
As a tradeoff, I have now added highlighting for the invalid parts.
This way it is easier to spot mistakes in the config file and the regex.

ipa-mdl:
> I'm curious. In which way you thought fragile?

It is harder to maintain, because the defaults could get changed at any time by Travis CI.
Fuerther, more I had to convert the rather complex regular pattern from Ruby to Python.

ipa-mdl:
Rebased to current master.

Example from #417 
image

PR #423, opened 2019-10-14
readme: repository was migrated to travis.com

gavanderhoorn:
As per subject.

Not sure whether you'd rather have this in `master`, or whether you can 'port' this to that branch yourself.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,8 +1,8 @@
 ================
 Industrial CI
 ================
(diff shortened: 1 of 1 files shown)


ipa-mdl:
Thanks!
I will stop the CI builds for now, otherwise the other repos will be blocked for some hours.

> Not sure whether you'd rather have this in master, or whether you can 'port' this to that branch yourself.

I will port it to the other branches as well.

PR #424, opened 2019-10-14
introduce APT_PROXY with auto-config for Travis

ipa-mdl:
Follow-up of #408 

Looks promising. The test only install `industrial_core` dependencies, but does not build it.


doc/index.rst
```diff
@@ -170,6 +170,7 @@ Note that some of these currently tied only to a single option, but we still lea
 * **ABICHECK_URL** (default: not set): Run binary compatibility check with `ABICC <https://github.com/lvc/abi-compliance-checker>`__. The URL should point to a baseline archive (\*.tar.\*,\*.zip, \*.tgz or \*.tbz2). See more in `the ABI checks section <#abi-checks>`__)
 * **ABICHECK_VERSION** (default: not set): Used only when `ABICHECK_URL` is set. Version name (for display only) of the set of code, which the location is specified in `ABICHECK_URL` of. The version will be automatically read from the URL passed in `ABICHECK_URL` if possible, but for a URL that doesn't point to a version-based file name (e.g. the link for a tagged version on Gitlab doesn't).
 * **ADDITIONAL_DEBS** (default: not set): More DEBs to be used. List the name of DEB(s delimitted by whitespace if multiple DEBs specified). Needs to be full-qualified Ubuntu package name. E.g.: "ros-indigo-roslint ros-indigo-gazebo-ros" (without quotation).
+* **APT_PROXY** (default: TRAVIS_APT_PROXY for Travis CI): Configure APT to use the provided URL as http proxy. An empty string disables the proxy.
```
130s: I've been becoming a novice user for the recent changes, so I might be perfect reviewer for documentation :)

Missing context of this potential new feature:
- What kind of specific content does the URL need to point to? `apt` file?
- In what case this feature can be useful.
- Is there any caviat?

Sorry, as a maintainer I could have traced back PRs and tickets then figure out answers to above, without requesting documenting. But then users may hit the same questions. So I'd rather be an annoyance.

ipa-mdl: > In what case this feature can be useful.

This was an experiment to make use of the Travis Apt package proxy

> What kind of specific content does the URL need to point to? apt file?

To any URI of a proxy server. For Travis it gets set to `$TRAVIS_APT_PROXY`

> Is there any caviat?

Probably. The Travis Apt proxy did/does not work properly. And all non-Travis users, would need to host an apt proxy themselves.

ipa-mdl:
superseded by #691

PR #428, opened 2019-10-24
do not pull images after qemu got injected

ipa-mdl:


ipa-mdl:
Gitlab tests passed: https://gitlab.com/ipa-mdl/industrial_ci/pipelines/91387808

ipa-mdl:
For `master` the same problem exists, but I want to fix in a different way.

Issue #274, opened 2018-04-03
Add support for python3


ipa-mdl:
We should introduce a flag `PYTHON_VERSION` (?) that switches to the `python3` debians.

Some questions:
* Are additonal steps needed for python3 testing?
* Should we provide docker images for python3?

130s:
Just a food for thought; [`rospkg` runs tests on CI against multiple Python versions](https://github.com/ros-infrastructure/rospkg/blob/5e55690e197411c29735632d3628e0b5ca6d6803/.travis.yml).

ipa-mdl:
`master` uses python3 for ROS2.
I don't know what's the status of ROS1 and python3, I guess it is targeted for `noetic`.

130s:
> I guess it is targeted for `noetic`.

Verified that seems to be the case.
- Melodic supports 2.7 (testing against 3.5 is recommended) [from REP0003](https://www.ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023).
- A set of [tutorial for transitioning to Python3 is recently started by OR on ROS wiki](http://wiki.ros.org/UsingPython3).

The tutorial above identifies a few areas that might need changes. Honestly I've never thought about Py3 transition, but I'm afraid there would be a lot of issues than I thought. Hope that won't be the case, but starting to provide CI environment earlier might be helpful for catching issues early? OR already opened up Jenkins job http://wiki.ros.org/UsingPython3/Contributing#Check_the_Noetic_CI_jobs for capturing issues early (although I'm struggling how to interpret its result).

ipa-mdl:
> Melodic supports 2.7 (testing against 3.5 is recommended) from REP0003.

But how to test? And how to tell rosdep?

ipa-mdl:
> But how to test? And how to tell rosdep?

I was able to figue it out, I guess ;)
#429 is based on http://wiki.ros.org/UsingPython3/BuildUsingPython3

I will have a look at the new Noetic CI jobs later

ipa-mdl:
Noetic support was merged: #499

ipa-mdl:
Python 3 is supporedt for ROS1 noetic and ROS2.
Proper support for ROS1 melodic is very hard to implement and most-likely not needed (anymore).

PR #429, opened 2019-10-27
Support for python3 with ROS1

ipa-mdl:
addresses #274 

It looks like ROS1 must be rebuilt from scratch in order to support python3.
So I add a `ROS_FROM_SCRATCH` feature, which uses `rosinstall_generator` to fetch all released sources.

The latter came in handy add support for `noetic` (based on Debian Buster) as well.


ipa-mdl:
Works, but the auto-generation is not as powerful without https://github.com/ros-infrastructure/rosinstall_generator/issues/25 being fixed.
As a fallback `UPSTREAM_WORKSPACE` should point to a (generated) rosinstall file.

arturmiller:
I have tried this out, with one of my repositories, but I have not been successful. I tried to run ROS melodic with Python3 and got as error message: "env: ‘rosdep’: No such file or directory". Am I missing something? Here is the build report: https://travis-ci.org/arturmiller/pybind11_examples/jobs/628120545#L186
And here is my travis.yml: https://github.com/arturmiller/pybind11_examples/blob/fix/ci/.travis.yml

ipa-mdl:
@arturmiller:
If I recall it correctly, I got this error when a dependency pulled in `python-catkin-pkg` for some reason.
This would replace `python3-catkin-pkg` and therefore uninstall `python3-rosdep`.
I will have a look again.

ipa-mdl:
> when a dependency pulled in `python-catkin-pkg`

In your case `python-rospkg`: https://travis-ci.org/arturmiller/pybind11_examples/jobs/628120545#L1154

arturmiller:
Just now I see what it takes to support Python3. Your are right a dependency is not Python3 ready. This is the reason for my error. I was able to create a [minimal example](https://github.com/arturmiller/ros_py3), which is tested with Melodic and Python3.
I think I found a bug. If the project has no dependencies to other ROS packages, rosinstall-generator throws an [error](https://travis-ci.org/arturmiller/ros_py3/jobs/629752310#L739). This is why I have included roscpp as dependency. I don't know if this should be discussed in this PR.
I think this PR is very Important and I'm glad you created it. Many maintainer want to test their projects with Python3 and don't know how. [Python2 retires very soon](https://pythonclock.org/?1).

ipa-mdl:
This is not needed anymore..
noetic is completely supported now and python 3 support for melodic seems  to introduduce more problems than it solves.

Issue #434, opened 2019-11-12
Running CI on bare rosinstall repo


fmessmer:
sing `legacy` branch, I'm trying something similar as already asked in #232 - that's why same title

My repo only contains the following `travis.yml` (includes some debugging options already) and a (~~currently empty~~/**minimal**) `.travis.rosinstall` file (to be filled later with other repos I would like to get built and tested

```
language: generic
services:
  - docker

notifications:
  email:
    on_success: change
    on_failure: always
env:
  global:
    - ROS_REPO=ros
    - DEBUG_BASH=true
    - VERBOSE_OUTPUT=true
    - VERBOSE_TESTS=false
    - NOT_TEST_INSTALL=true
    - UPSTREAM_WORKSPACE=file
    - _GUARD_INTERVAL=10
  matrix:
    - ROS_DISTRO=kinetic
install:
  - git clone https://github.com/ros-industrial/industrial_ci.git .ci_config
script:
  - .ci_config/travis.sh
```

The output (when running `rosrun industrial_ci run_ci ROS_DISTRO=kinetic ROS_REPO=ros DEBUG_BASH=true VERBOSE_OUTPUT=true VERBOSE_TESTS=false NOT_TEST_INSTALL=true UPSTREAM_WORKSPACE=file _GUARD_INTERVAL=10`) on my repo, is the following:

```
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
Starting function 'catkin_build'
++ export TARGET_PKGS
++ export PKGS_DOWNSTREAM
++ echo -e 'TARGET_PKGS: '
TARGET_PKGS: 
++ echo -e 'TARGET_REPO_PATH: /root/src/mojin_ci'
TARGET_REPO_PATH: /root/src/mojin_ci
++ '[' '' == '' ']'
+++ catkin_topological_order /root/src/mojin_ci --only-names
Workspace "/root/src/mojin_ci" seems to not contain any projects. Have you passed the correct path to a catkin workspace?++ TARGET_PKGS=
+ ici_exit
+ local exit_code=3
+ trap - EXIT
+ '[' -n catkin_build ']'
+ '[' 3 -ne 0 ']'
+ color_wrap=31
+ ici_time_end 31 3
+ '[' true ']'
+ '[' true == true ']'
+ set +x
[...]

PR #436, opened 2019-11-18
not fail catkin_topological_order for empty workspace

fmessmer:
this allows to use `legacy` in the usecase mentioned in #434

### DIFF
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -199,9 +199,11 @@ ici_time_start catkin_build
 export TARGET_PKGS
 export PKGS_DOWNSTREAM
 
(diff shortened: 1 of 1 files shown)


fmessmer:
how else could we catch it?

ipa-mdl:
> how else could we catch it?

I don't know a proper way right now.
However, building an empty target workspace is not a supported use case.

fmessmer:
ok, I think I can live with the solutions mentioned in https://github.com/ros-industrial/industrial_ci/issues/434#issuecomment-555032582, i.e. use `master` branch or use `legacy` + add a meta-package....

ipa-mdl:
I will close this PR now, because the `master` branch already offers this functionality already and a (simple) back-port is not possible.

PR #442, opened 2019-11-27
crystal is EOL

ipa-mdl:


ipa-mdl:
>  I see this is WIP.

Yes, because crystal is not yet EOL (?).
(https://github.com/ros/rosdistro/blob/master/index-v4.yaml#L21)

ipa-mdl:
https://github.com/ros/rosdistro/pull/23486

PR #443, opened 2019-12-04
add --recursive for vcs import

fmessmer:
apparently `vcs import` has `--recursive` disabed by default
which is why github submodules are not pulled during setup_ws

from `vcs import --help`:
>  --recursive        Recurse into submodules (default: False)

I hope I covered all occurrences correctly...

### DIFF
--- a/industrial_ci/src/workspace.sh
+++ b/industrial_ci/src/workspace.sh
@@ -79,6 +79,10 @@ function ici_setup_git_client {
   fi
 }
 
+function ici_vcs_import {
+    vcs import --recursive "$@"
+}
+
 function ici_import_repository {
     local sourcespace=$1; shift
     local url=$1; shift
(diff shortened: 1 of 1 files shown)


ipa-mdl:
Overall: :+1: 
I am just hesitating, because this is not the default behavior of `vcs import`.

I see 3 options:
1. merge as-is
2. add variable, default to false (opt-in)
3. add variable, default to true (opt-out)

Option 2 would be the most conservative and might be what users expect.
Option 1 can be migrated into option 3 later.

What do you think?

gavanderhoorn:
The ROS buildfarm defaults to recursive cloning I believe since ros-infrastructure/ros_buildfarm#515.

Might be good to mirror that behaviour here.

fmessmer:
just let me know what you like best and I can try to adjust the pr

ipa-mdl:
@fmessmer: please have a look at https://github.com/ros-industrial/industrial_ci/pull/443/commits/23ed7ff17394a663e03a26173af8381f5bd49310

fmessmer:
@ipa-mdl 
thanks, works as expected :+1: 
(I don't think this PR is responsible for breaking the "Travis CI - Pull Request" check, right?)

fmessmer:
@ipa-mdl are you happy with this PR - or are there any further changes requested?

ipa-mdl:
@fmessmer: I think I broke the tests with my changes. I will provide a slightly different version. And then we can test and merge.

fmessmer:
thanks!
:tada:

PR #448, opened 2019-12-13
install catkin_lint from ROS repository

ipa-mdl:


ipa-mdl:
needs to be back-ported to `legacy` as well.

PR #454, opened 2019-12-19
Support for CI on top of ros2 nightly images

mikaelarguedas:
Hi all, I created a way simpler and stupider version of ici for ROS 2 a while back ([ros2ci](https://github.com/mikaelarguedas/ros2ci/)). Main purpose was to be able to run CI on both released ROS 2 distros and ROS 2 nightly builds to catch breakage of packages against upcoming versions of ROS 2.

Now that ICI supports colcon + ROS 2 (hat tip @ipa-mdl :+1::tada:), I'd like to integrate the support for nightly builds instead of maintaining something on the side.

I'm opening this to start discussion of supporting nightly ROS 2 CI in ICI.
Feature description:
- run jobs based on OSRF's nightly ros2 docker images (Linux only)
- build a "middle-lay" with missing dependencies
- build and test packages on top of it

Is it a use case ICI would be interested in supporting?

---

Content of this PR:
- https://github.com/ros-industrial/industrial_ci/commit/bb03f8451811bd6cf5ab866c34dd2e3b07c89f54: Modify ICI to forward declare the ROS 2 distros as soon as they are [announced / used in the nightly builds](https://github.com/ros2/ros2/pull/829), with a blanket empty "DEFAULT_DOCKER_IMAGE"
- https://github.com/ros-industrial/industrial_ci/commit/c61ecab8324047b177b5d818d4f9fdcbd4fe0981 Extend instead of overriding ROS_PACKAGE_PATH to keep the one set in the upstream image (This should not be needed or merged IMO as thiscan modify behavior for all users and there are likely good reasons to override. This is a hack to work around a current rosdep/ROS2 issue https://github.com/osrf/docker_images/pull/328)

Users wanting to run CI based on nightly builds could modify their travis config as follows:
```diff
 env:
   matrix:
     - ROS_DISTRO="dashing" ROS_REPO=main
     - ROS_DISTRO="eloquent" ROS_REPO=main
+    - DOCKER_IMAGE="osrf/ros2:nightly" UPSTREAM_WORKSPACE="additional_repos.repos"
```

- `osrf/ros2:nightly` being the docker image provided daily by osrf with the content of the latest nightly build
- `additional_repos.repos` the list of upstream packages needed to compile the repository that are NOT part of the nightly image

Example of [repo using this](https://github.com/Karsten1987/confbot_robot/blob/ca85f937cf51b597477837c4a636dad88fc39a20/.travis.yml) and [travis job](https://travis-ci.org/Karsten1987/confbot_robot/jobs/627319651?utm_medium=notification&utm_source=github_status)

### DIFF
--- a/industrial_ci/src/env.sh
+++ b/industrial_ci/src/env.sh
@@ -123,6 +123,10 @@ function set_ros_variables {
     "eloquent")
         ros2_defaults "bionic"
         ;;
+    "foxy")
+        ros2_defaults "bionic"
+        DOCKER_IMAGE=${DOCKER_IMAGE:-osrf/ros2:nightly}
+        ;;
     esac
 
     local prefix=ros
--- a/industrial_ci/src/workspace.sh
+++ b/industrial_ci/src/workspace.sh
@@ -234,15 +234,12 @@ function ici_exec_in_workspace {
 function ici_install_dependencies {
     local extend=$1; shift
     local skip_keys=$1; shift
-    local cmake_prefix_path
-    cmake_prefix_path="$(ici_exec_in_workspace "$extend" . env | grep -oP '^CMAKE_PREFIX_PATH=\K.*')" || true
-
     rosdep_opts=(-q --from-paths "$@" --ignore-src -y)
     if [ -n "$skip_keys" ]; then
       rosdep_opts+=(--skip-keys "$skip_keys")
     fi
     set -o pipefail # fail if rosdep install fails
-    ROS_PACKAGE_PATH=$cmake_prefix_path ici_exec_in_workspace "$extend" "." rosdep install "${rosdep_opts[@]}" | { grep "executing command" || true; }
+    ici_exec_in_workspace "$extend" "." rosdep install "${rosdep_opts[@]}" | { grep "executing command" || true; }
     set +o pipefail
 }
 


industrial_ci/src/env.sh
```diff
@@ -123,6 +123,10 @@ function set_ros_variables {
     "eloquent")
         ros2_defaults "bionic"
         ;;
+    "foxy")
+        ros2_defaults "bionic"
+        DEFAULT_DOCKER_IMAGE=
```
ipa-mdl: Why not
```suggestion
        DEFAULT_DOCKER_IMAGE=osrf/ros2:nightly
```
This way, one can specify `DOCKER_IMAGE=osrf/ros2:nightly` (for rolling nightly version) or `ROS_DISTRO=foxy` (for next release only)

industrial_ci/src/workspace.sh
```diff
@@ -242,7 +242,7 @@ function ici_install_dependencies {
       rosdep_opts+=(--skip-keys "$skip_keys")
     fi
     set -o pipefail # fail if rosdep install fails
-    ROS_PACKAGE_PATH=$cmake_prefix_path ici_exec_in_workspace "$extend" "." rosdep install "${rosdep_opts[@]}" | { grep "executing command" || true; }
+    ROS_PACKAGE_PATH=$cmake_prefix_path:${ROS_PACKAGE_PATH} ici_exec_in_workspace "$extend" "." rosdep install "${rosdep_opts[@]}" | { grep "executing command" || true; }
```
ipa-mdl: `${ROS_PACKAGE_PATH}` should be empty at this point

industrial_ci/src/env.sh
ipa-mdl: Or `DOCKER_IMAGE=osrf/ros2:nightly` for now?

industrial_ci/src/workspace.sh
mikaelarguedas: In the case of the nightly image it is not empty https://github.com/osrf/docker_images/pull/328.
But it is a hack because of a rosdep issue, so I dont think this commit should be merged in industrial_ci, this is only propagating the workaround to have a working example for this discussion

industrial_ci/src/env.sh
mikaelarguedas: > This way, one can specify DOCKER_IMAGE=osrf/ros2:nightly (for rolling nightly version) or ROS_DISTRO=foxy (for next release only)

I am not sure I understood what the intended result is from these 2 different configurations.
What are the 2 scenarios we are trying to cover with this change?
What would the `.travis.yml` line would look like for each?

> Or DOCKER_IMAGE=osrf/ros2:nightly for now?

This will work more consistently than `DEFAULT_DOCKER_IMAGE=osrf/ros2:nightly` as the `DEFAULT_DOCKER_IMAGE` gets overwritten for some `ROS_REPO` values (including the default one)

---

A few hidden assumptions / expectations from my part that came up:

- Clear at first glance
It should be clear to user which job is a nightly build vs a build against a released distribution
in ros2ci it was done by naming the job nightly
```yaml
    - env: JOB_TYPE=nightly
      script:
      - cp additional_repos.repos .ros2ci/
      - .ros2ci/travis.bash $JOB_TYPE
```
Not sure what would be the best for ICI. Maybe having the `DOCKER_IMAGE=osrf/ros2:nightly` at the beginning of the line is enough :man_shrugging: 
- Minimal effort to set up
Should be a one liner for user to set up in their CI system e.g.:
```diff
 env:
   matrix:
     - ROS_DISTRO="dashing" ROS_REPO=main
     - ROS_DISTRO="eloquent" ROS_REPO=main
+    - DOCKER_IMAGE="osrf/ros2:nightly" UPSTREAM_WORKSPACE="additional_repos.repos"
```

- Zero maintenance from users
Once added the line should never need to be updated to keep tracking "ROS 2 master"
  - line should be `ROS_DISTRO`-free
- Clearly separated from release jobs
The job defined will never track a released `ROS_DISTRO` from debians, if users want to add a job for a new distro they should
```diff
 env:
   matrix:
     - ROS_DISTRO="dashing" ROS_REPO=main
     - ROS_DISTRO="eloquent" ROS_REPO=main
+    - ROS_DISTRO="foxy" ROS_REPO=main
     - DOCKER_IMAGE="osrf/ros2:nightly" UPSTREAM_WORKSPACE="additional_repos.repos"
```

ipa-mdl: > I am not sure I understood what the intended result is from these 2 different configurations.

1. test against night build (`DOCKER_IMAGE="osrf/ros2:nightly" UPSTREAM_WORKSPACE="additional_repos.repos`)
2. test against ROS2 foxy (`ROS_DISTRO="foxy" ROS_REPO=main`). Could point to the nightly build right now and can be switched to the testing/main repo as they become available.

industrial_ci/src/workspace.sh
ipa-mdl: I see. Shouldn't this path be included in `CMAKE_PREFIX_PATH`? I stumbled upon this issue as well and `CMAKE_PREFIX_PATH` did the trick in my tests. However, I haven't tested it against the nightly build yet.

industrial_ci/src/env.sh
mikaelarguedas: Thanks for clarifying :+1:

Yes it should work.
Users with all their dependencies in [ros2.repos](https://github.com/ros2/ros2/blame/master/ros2.repos) will be fine, the ones with extra dependencies will need to remove the `UPSTREAM_WORKSPACE` once the packages are available in the distro
```
ROS_DISTRO="foxy" ROS_REPO=main UPSTREAM_WORKSPACE="additional_repos.repos"
```
 to
```
ROS_DISTRO="foxy" ROS_REPO=main
```

industrial_ci/src/workspace.sh
mikaelarguedas: hmm I think rosdep does only look in `ROS_PACKAGE_PATH` and `AMENT_PREFIX_PATH`.
If all the packages in the nightly archive were registered in the ament index it would work (ongoing discussion at https://github.com/ros2/tinyxml_vendor/pull/14)

ipa-mdl: > hmm I think rosdep does only look in ROS_PACKAGE_PATH

That's why I set it to `CMAKE_PREFIX_PATH`, this works (or worked?) for catkin and ament :)

gavanderhoorn: But `rosdep` is not Catkin/Ament, correct?

I haven't checked, but if `rosdep` is used to resolve package dependencies in the scenarios @mikaelarguedas is trying to support, `CMAKE_PREFIX_PATH` may not help.

ipa-mdl: > in the scenarios @mikaelarguedas is trying to support, CMAKE_PREFIX_PATH may not help

I think I already tested it against this scenario, but I will test again :)

ipa-mdl: > But rosdep is not Catkin/Ament, correct?

rosdep is rosbuild-based with some work-arounds for catkin and ament ;)

mikaelarguedas: > That's why I set it to CMAKE_PREFIX_PATH, this works (or worked?) for catkin and ament :)

So I looked a bit into that.
TL;DR this line can stay as is if we remove the COLCON_IGNORE file located at the root of the archive https://github.com/mikaelarguedas/industrial_ci/commit/241216c2206749df874f776fd89eda94040a3296
I dont think this file serve much purpose in this case and if removed it should likely be removed upstream either in the docker image or in the archive generation itself

---

More technical dive:

By setting `ROS_PACKAGE_PATH` to anything, we are basically asking `rosdep` (and under the hood `catkin_pkg`) to crawl these directories looking for catkin packages.
By setting it to the `CMAKE_PREFIX_PATH` that is in general higher up in the hierarchy than `ROS_PACKAGE_PATH` we're adding overhead.

`ament` by opposition uses the ament resource index using the prefixes listed in only `AMENT_PREFIX_PATH` and does not crawl or populate third-party variables like `CMAKE_PREFIX_PATH`. So the ROS2 docker images (e.g. `ros:dashing`) do not set `CMAKE_PREFIX_PATH` or `ROS_PACKAGE_PATH`. Only `AMENT_PREFIX_PATH`.


There seem to be a couple different scenarios for ros2 builds:
  - For ROS2 docker images with only debs: `CMAKE_PREFIX_PATH` is not set
    - So if this works in this case is because the `AMENT_PREFIX_PATH` is used
  - For workspaces built with colcon: `CMAKE_PREFIX_PATH` is set to `AMENT_PREFIX_PATH` (e.g. `/opt/ros/foxy`) for compatibility, **BUT** a `COLCON_IGNORE` file is dropped at the same location. So `catkin_pkg` thus `rosdep` cannot find the packages at that location
  - The nightly image is the same as the previous case + `ROS_PACKAGE_PATH` is set to the share directory (like in ROS1) `/opt/ros/foxy/share` so there is no `COLCON_IGNORE` and it can find the packages by crawling

Ideal outcome IMO: for ROS2 the crawling approach is never used. No need to set or modify `ROS_PACKAGE_PATH` or `CMAKE_PREFIX_PATH`, only use `AMENT_PREFIX_PATH`
Packages in the archive should all be registered in the ament index by either creating a marker file in the resource index or by being ament packages to start with

ipa-mdl: >  only use AMENT_PREFIX_PATH

catkin packages do not set `AMENT_PREFIX_PATH` (?)

> So the ROS2 docker images (e.g. ros:dashing) do not set CMAKE_PREFIX_PATH

I know, and they should not set it. I am reading it from `setup.bash`
This is a temporary hack until rosdep can crawl the ament index.

mikaelarguedas: > catkin packages do not set AMENT_PREFIX_PATH (?)

not that I know of. Even if a tool build tool like colcon did it, I wouldn't expect catkin packages to register themselves in the index. 

>  This is a temporary hack until rosdep can crawl the ament index.

This is now a thing https://github.com/ros-infrastructure/rosdep/pull/699
The nightly image sets the `ROS_PACKAGE_PATH` only because rosdep cannot find non-ament packages like https://github.com/ros2/tinyxml_vendor/pull/14

ipa-mdl: > TL;DR this line can stay as is if we remove the COLCON_IGNORE

I just tested it, the `COLCON_IGNORE` is breaking the work-around.

If `rosdep` now support `AMENT_PREFIX_PATH`, then we don't need to set `ROS_PACKAGE_PATH`.

ipa-mdl: #456 now reverts the work-around properly.

industrial_ci/src/env.sh
```diff
@@ -123,6 +123,10 @@ function set_ros_variables {
     "eloquent")
         ros2_defaults "bionic"
         ;;
+    "foxy")
+        ros2_defaults "bionic"
+        DOCKER_IMAGE=osrf/ros2:nightly
```
ipa-mdl:
```suggestion
        DOCKER_IMAGE=${DOCKER_IMAGE:-osrf/ros2:nightly}
```
This allow sderived images as well.

mikaelarguedas:
> Please rebase this to #456 and commit my suggestion.
If the new version still passes your tests, I will merge both PRs.

Haven't tried this yet but I expect #456 to break the use of nightly images.
If the nightly images set the` ROS_PACKAGE_PATH` it's because some packages like https://github.com/ros2/tinyxml_vendor/pull/14 cannot be found using the `AMENT_PREFIX_PATH` approach at the moment

So for the nightly builds to work we would have to keep populating the ROS_PACKAGE_PATH here for now

industrial_ci/src/workspace.sh
mikaelarguedas: > I just tested it, the COLCON_IGNORE is breaking the work-around.

In what scenario does it break? Can you clarify how it breaks?

If removing it has a negative effect, I'll reject https://github.com/osrf/docker_images/pull/345

industrial_ci/src/env.sh
mikaelarguedas: Thanks!

ipa-mdl:
> Haven't tried this yet but I expect #456 to break the use of nightly images.

I tested it locally with https://github.com/Karsten1987/confbot_robot/blob/ca85f937cf51b597477837c4a636dad88fc39a20/.travis.yml

It seems to work (all tests pass), but it displays a lot of connext-related errors/warnings.

mikaelarguedas:
> It seems to work (all tests pass), but it displays a lot of connext-related errors/warnings.

Oh yeah,  it will work because the `ROS_PACKAGE_PATH` work-around hard coded in the docker image is not being overwritten anymore :+1:

I'll rebase on #456 and force-push on this branch

mikaelarguedas:
> but it displays a lot of connext-related errors/warnings.

Yeah this is another issue with the OR nightly archive, unrelated to this PR

mikaelarguedas:
> I am not sure if we should list foxy in the list of supported distros.

Not sure either. As long as Foxy is not released (or in beta phase) the statement would be misleading as testing against the last nightly != testing against officially released Foxy...

An alternative is to add "ROS 2 nightly builds" to the list of supported distros. It does not qualify as a distro per se but would be more accurate.
I you go that route, linking to an example of usage would be beneficial as it differs from other distros

ipa-mdl:
> An alternative is to add "ROS 2 nightly builds" to the list of supported distros

I'd like to support something like `ROS_DISTRO=nightly`.

> I you go that route, linking to an example of usage would be beneficial as it differs from other distros

For now we should just merge it as-is.
We can update the documentation, once we have tested it further.

---

Please let me know if your test repo works with this version. I will postpone the merge until then.

mikaelarguedas:
> I'd like to support something like ROS_DISTRO=nightly.

Yeah me too, there were some blockers on the docker image side for a while. It may be possible now, would need to check again.

> Please let me know if your test repo works with this version. 

It works for me! 
Thanks for the support on getting this PR ready for merge :+1:

PR #456, opened 2019-12-21
Revert "work-around for chained colcon workspaces"

ipa-mdl:
This reverts commit 09618f49562b1bbdc98b1385d9ef7a560b81ec85.

as discussed in https://github.com/ros-industrial/industrial_ci/pull/454#discussion_r360407669, this work-around is not needed anymore.


ipa-mdl:
`ardent` and `bouncy` ship with an older version (in the snapshot repo), so they would still need this work-around.
The latest version of rosdep can get installed with 
```
ADDITIONAL_DEBS=python3-pip AFTER_INIT="pip3 install rosdep"
```

ipa-mdl:
Included in #454

PR #458, opened 2019-12-25
implement and use true/false helpers

ipa-mdl:
This replaces all input-based conditions  with semantic function.
Fixes some broken code, especially 
```
if [ "$VERBOSE_OUTPUT" != false ]; then
```


ipa-mdl:
This is just an intermediate solution. The `*_or_unset` cases should be implemented with proper default values (something like #452).

PR #461, opened 2020-01-14
adapt test to upstream changes in cob_calibration_data

ipa-mdl:


fmessmer:
reason for upstream change was `catkin_lint`
I didn't want to install the .travis script, thus it could not be executable...

PR #462, opened 2020-01-14
initialize VERBOSE_OUTPUT and VERBOSE_TESTS to false

ipa-mdl:
quick fix as long as #458 is not ready


ipa-mdl:
The failing test is unrelated.

Issue #465, opened 2020-01-18
run_travis fails during setup_upstream_workspace with private repos while run_ci does not


fmessmer:
testing locally, `run_travis` fails to setup_upstream_workspace with the following rosinstall file containing a private repo
```
- git:
    local-name: mojin_supported_robots
    uri: git@github.com:mojin-robotics/mojin_supported_robots.git
    version: indigo-devel
```
while `run_ci` - with the same job options works
is it only me?
any idea?

<details>
<summary>rosrun industrial_ci run_travis 1</summary>

```
( | )fxm@alpamayo:~/projects/default/robot_ws/src/mojin_calibration_data (rosenv_after_script)$ rosrun industrial_ci run_travis 1
Running job 1: /home/fxm/projects/ci/install/lib/industrial_ci/run_ci /home/fxm/projects/default/robot_ws/src/mojin_calibration_data ADDITIONAL_DEBS='apt-utils dialog curl wget' AFTER_SCRIPT='rosenv sh .travis.xacro_test.sh' CATKIN_LINT=pedantic CMAKE_ARGS=-DCMAKE_BUILD_TYPE=Release NOT_TEST_DOWNSTREAM=true ROS_REPO=ros UPSTREAM_WORKSPACE=file VERBOSE_OUTPUT=false VERBOSE_TESTS=false ROS_DISTRO=kinetic
Testing /home/fxm/projects/default/robot_ws/src/mojin_calibration_data
Replacing 'UPSTREAM_WORKSPACE=file' with 'UPSTREAM_WORKSPACE=.travis.rosinstall'

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
Starting function 'prepare_docker_image'
Pulling Docker image 'ros:kinetic-ros-core'...
<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
Function 'prepare_docker_image' returned with code '0' after 0 min 3 sec

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
Starting function 'init'
debconf: delaying package configuration, since apt-utils is not installed
Selecting previously unselected package libapt-inst2.0:amd64.
(Reading database ... 42390 files and directories currently installed.)
[...]

PR #466, opened 2020-01-18
add HOME to whitelist

fmessmer:
fixes https://github.com/ros-industrial/industrial_ci/issues/465


ipa-mdl:
cherry-picked to legacy as well.

PR #469, opened 2020-01-25
pin std_msgs version to indigo-compatible version

ipa-mdl:


ipa-mdl:
The failing test will recover after the merge. It reads the rosinstall file directly from Github.

PR #471, opened 2020-01-26
explicitly set ROS_DISTRO in rosdep update

ipa-mdl:
to take advantage of https://github.com/ros-infrastructure/rosdep/pull/738


ipa-mdl:
Ignoring `ros_control` failures for now..

PR #463, opened 2020-01-16
[WIP][master] pylint check

fmessmer:
still WIP, but I wanted to hear people's thoughts about adding `pylint` to `industrial_ci`

example config, i.e. add to `.travis.yml` of your repo which contains python code
```
env:
  global:
    - PYLINT=true
    - PYLINT_ARGS="--output-format=parseable --reports=no"
    - PYLINT_VERSIONS="pylint pylint3"
```

motivation:
we want to verify that our code is both python2- and python3-compatible!
which is why we need to run both `pylint` (for python2.7) and `pylint3` (for python3.5)

what do you think of this?
any comments on how it should be implemented or how the this first draft should be modified?

PR #477, opened 2020-02-18
[master] pylint check - continued

fmessmer:
this extends #463 
- implement `PYLINT_EXCLUDE` - inspired by https://github.com/ros-industrial/industrial_ci/pull/463#issuecomment-585315152
- refactors the error reporting (per-file folding, error summary)

@ipa-mdl 
could you have a look at this PR?
Unfortunately, I'm using my `master_pylint` branch from #463 in various repos. Thus, I had to provide a new PR from a new branch...but I also cleaned up the commit history a bit

once happy with the feature, I will
- finalize examples and tests
- add documentation to the README


fmessmer:
I'm testing this branch here: https://github.com/ipa320/cob_control/pull/233

fmessmer:
is it possible to use "nested folding"?

fmessmer:
example log of (intentionally) failing job: https://travis-ci.com/ipa320/cob_control/builds/149445930

fmessmer:
Is there anything else I can do to speed-up review/merge this PR?

@agutenkunst @gavanderhoorn 
You've been reacting to #463 - do you have an opinion on this feature?

gavanderhoorn:
I haven't checked the PR in detail, but if a high-level overview of the rationale for treating `pylint` as a special case could be given that would be good.

Same question/comment as https://github.com/ros-industrial/industrial_ci/pull/463#issuecomment-575349222.

fmessmer:
well, I don't know anything about `ament_lint`...
our motivation for adding this to ICI, was that we wanted a CI mechanism that helps us verify that our code is - and stays - py2/py3-compatible....in the end, `pylint` also did find some other bugs in the code that we have missed until then...
it's now another tool that helps us check PRs for correctness automatically, i.e. assist/support QA...

gavanderhoorn:
> well, I don't know anything about `ament_lint`...

`ament_lint` was just an example. It's not about that package specifically.

> our motivation for adding this to ICI, was that we wanted a CI mechanism that helps us verify that our code is - and stays - py2/py3-compatible....in the end, `pylint` also did find some other bugs in the code that we have missed until then...
> it's now another tool that helps us check PRs for correctness automatically, i.e. assist/support QA...

I understand why having a tool like `pylint` in a CI pipeline would be desirable. I'm also not questioning this aspect.

What I am wondering about is whether it would be more efficient to make running something like `pylint` the responsibility of a specific package (say `pylint_ros` or whatever) and then have it export a target/CMake function you can call from the packages you want linted.

This would be similar to how `roslint` works (and incidentily: `ament_lint`).

The net effect is the same: your code gets linted, but:

 - no tool specific support needed in `industrial_ci`
 - adding additional linting / quality checkers would mean adding another package and having your to-be-linted package depend on it and call a new CMake function
 - linting/quality checks can be run easily in a regular workspace (ie: without needing `industrial_ci`)

This is actually part of the rationale for why something like `ament_lint` exists.

I was wondering why we opted to add tool-specific support to `industrial_ci`, hence my comment.

fmessmer:
ping @ipa-mdl 
how to proceed with this feature?

ipa-mdl:
> how to proceed with this feature?

I will have a look soon

fmessmer:
any update on when "soon" will be?
can I assist with something?

fmessmer:
:tada: happy 3-months anniversary :birthday: 
(:point_right: this is a ping :wink:)

fmessmer:
any update on this PR getting reviewed/merged?

fmessmer:
proper solution needed:
`noetic` is python3-only....thus, there is no `pylint3` (for python3) vs. `pylint` (for python2.7) anymore....it's just called `pylint` (for python3)...but `pylint` executable name is used as `cmd` in `PYLINT2_CHECK`...

industrial_ci/src/tests/source_tests.sh
```diff
@@ -97,6 +97,96 @@ function run_clang_tidy_check {
     fi
 }
 
+function install_pylint {
```
ipa-mdl, line 100: What is the purpose of this function?

.travis.pylintrc
```diff
@@ -0,0 +1,378 @@
+[MASTER]
```
ipa-mdl, line 1: Do we need this file at the top level?
Why is it prefixed with ".travis"?
Where are the settings from?

industrial_ci/src/tests/source_tests.sh
```diff
@@ -97,6 +97,96 @@ function run_clang_tidy_check {
...
+        local name=${file#$path/}
+        local exit_code=0
+        run_pylint_file exit_code "$target_ws" "$cmd" "$file" "$name" "${pylint_args[@]}"
+        if [ "$exit_code" -gt "0" ]; then
+            _run_pylint_errors+=("$name")
+        fi
+    done < <(ici_find_nonhidden "$path" "${pylint_find_pattern[@]}")
+}
+
+function run_pylint_check {
```
ipa-mdl, line 152: This function should be refactored so it can be called twice, once for pylint2 and once for pylint3

industrial_ci/src/tests/source_tests.sh
```diff
@@ -97,6 +97,96 @@ function run_clang_tidy_check {
...
+    # shellcheck disable=SC2016,SC2034
+    for p in $PYLINT_EXCLUDE; do pylint_find_pattern+=(-not -path '*$p*'); done
+    pylint_find_pattern+=(-type f -iname '*.py')
+
+    local path; path="$target_ws/src"
+
+    while read -r file; do
+        local name=${file#$path/}
+        local exit_code=0
+        run_pylint_file exit_code "$target_ws" "$cmd" "$file" "$name" "${pylint_args[@]}"
```
ipa-mdl, line 145: Why did you factor this out into a function?
The call looks rather complicated, especially with `_run_pylint_errors`

.travis.pylintrc
fmessmer:
```
pylint --generate-rcfile > .pylintrc
```

it's just been an example to test running pylint against with a given config file

industrial_ci/src/tests/source_tests.sh
fmessmer: `run_pylint` runs pylint/pylint3 in general while `run_pylint_file` runs pylint/pylint3 on a single file....and then adds the file to the errors in case pylint reports an error....

industrial_ci/src/tests/source_tests.sh
fmessmer: this is what `run_pylint` is there for....

industrial_ci/src/tests/source_tests.sh
fmessmer: this installs pylint based on the pylint_cmd requested, i.e. pylint for pylint2 and pylint3 for pylint3

fmessmer:
I had hoped to get some feedback on this feature and how to bulletproof the check, present the results and optimize the performance/configurability....

ipa-mdl:
> proper solution needed:
> `noetic` is python3-only....thus, there is no `pylint3` (for python3) vs. `pylint` (for python2.7) anymore....it's just called `pylint` (for python3)...but `pylint` executable name is used as `cmd` in `PYLINT2_CHECK`...

What do you think of just having `PYLINT_CHECK`?
For melodic and older, it will be python2 and for noetic it will be python3?

And do you really need a ROS set-up for these checks?

fmessmer:
> > proper solution needed:
> > `noetic` is python3-only....thus, there is no `pylint3` (for python3) vs. `pylint` (for python2.7) anymore....it's just called `pylint` (for python3)...but `pylint` executable name is used as `cmd` in `PYLINT2_CHECK`...
> 
> What do you think of just having `PYLINT_CHECK`?
> For melodic and older, it will be python2 and for noetic it will be python3?

I'm already kinda using it this way in my repos...and I mentioned this issue already: https://github.com/ros-industrial/industrial_ci/pull/477#issuecomment-699495254
Haven't changed anything yet as I wanted to get some general feedback on the feature...
Is there anything else in this PR that needs to be optimized/changed, i.e. the general idea of this check or the way the results are presented - e.g. see https://github.com/ipa320/cob_control/pull/233

> ~And do you really need a ROS set-up for these checks?~ (yes)

yes, we discussed this already and tried it without ROS setup

ipa-mdl:
> Is there anything else in this PR that needs to be optimized/changed, i.e. the general idea of this check or the way the results are presented

I do not have a strong opinion about this, because I am not using Python with ROS so much..
So far it looks good to me.

From my point of view, we should clean the code up an then push it out.
And it might manke sense to define a sane set of default arguments,
If the presentation works for you, then there is big chance that it works for others as well.

fmessmer:
one aspect I'm still unclear about is whether I should call pylint on a list of files or call pylint for each file individually
this gives different reports 
however, pylint on a list of files is much faster - it seems to not check the scripts but only the modules though:

small test script:
```
#!/bin/bash

# pylint
python_files=$(find -iname "*.py")
if [ -z "$python_files" ]; then
  printf "\nno python files found, skipping pylint\n"
else
  printf "\n##################################\nPylint ALL:\n##################################\n"
  START=$(date +%s.%N)
  pylint --errors-only --output-format=parseable $python_files
  END=$(date +%s.%N)
  DIFF=$(echo "$END - $START" | bc)
  echo $DIFF
  printf "\n##################################\nPylint Single:\n##################################\n"
  START=$(date +%s.%N)
  for VARIABLE in $python_files
  do
    pylint --errors-only --output-format=parseable $VARIABLE
  done
  END=$(date +%s.%N)
  DIFF=$(echo "$END - $START" | bc)
  echo $DIFF
fi
```

fmessmer:
ref https://docs.pylint.org/en/1.6.0/run.html

I think modules and scripts need to be handled differently....i.e. on modules, I can run pylint on multiple py files and for scripts, I have to call pylint on each file separately/explicitly....apparently pylint tries to convert each file into a module and then for script other files in the script folder get lost....

ipa-mdl:
> for scripts, I have to call pylint on each file separately/explicitly

More importantly, you might have to [add the scripts folder to PYTHONPATH](https://docs.pylint.org/en/1.6.0/faq.html#can-i-give-pylint-a-file-as-an-argument-instead-of-a-module).

For packages, only the packages are needed.

fmessmer:
To be honest, I'm close to abandon this PR/feature...it frustrates me more and more the longer I work on it...
...the way it is, it has found various issues in our code for many months already...but it's super-slow...and if I change it to running pylint against all files at the same time this is against the way to represent the errors per file we once discussed...and I don't understand pylint well enough in order to guarantee it's not missing anything when done that way...
in case there are specific changes requested then let me know...

ipa-mdl:
Let me play around with this a little bit :)
I am just busy with #572 right now

fmessmer:
ok...thanks!
after some more testing - to the best of my knowledge - I haven't found any errors/issues that would be important to us that are not found by pylint when run against all files at once - in contrary to running pylint against each file individually (as this PR is doing it atm)

thus, I'd go for the speed-up achieved running pylint against all files at once
let me know once you did your testing and made up your own thoughts...

fmessmer:
:skull: please have a look at #612

fmessmer:
closing in favor of #612

PR #506, opened 2020-05-06
Fix foxy

### LINKED ISSUE #505: Foxy nightly rosdep update cannot find distro
### Bug Description
During my **foxy** nightly [build](https://travis-ci.com/github/Briancbn/packml_ros2/jobs/328472167), 

`rosdep update` failed with the message 
```
Requested distribution "foxy" is not in the index.
```
[Here](https://github.com/Briancbn/packml_ros2/blob/62e7c7423b4b49c91f0a21c9858b9fa01006eefc/.travis.yml) is my `.travis-ci.yml` configuration

### Proposed Solution

I did some investigation, it seems that after #471 is merged, the nightly build will have this error .The environment variable `ROS_DISTRO` is foxy, but it is not really a ros distro released . Should we temporarily treat **foxy** as an exception here, or just revert the original PR for the moment.

ipa-mdl:
adresses #505


ipa-mdl:
This bypasses #471, if a custom `ROSDISTRO_INDEX_URL` was set in a Docker image (e.g. with `osrf/ros2:nightly`)

PR #509, opened 2020-05-12
vcstool is always python3

ipa-jba:
There is no working way to install `python-vcstool`. It is always `python3-vcstool`. This change affects only pre-noetic distros

### DIFF
--- a/industrial_ci/src/workspace.sh
+++ b/industrial_ci/src/workspace.sh
@@ -87,7 +87,7 @@ function ici_import_repository {
     local sourcespace=$1; shift
     local url=$1; shift
 
-    ici_install_pkgs_for_command vcs "${PYTHON_VERSION_NAME}-vcstool"
+    ici_install_pkgs_for_command vcs python3-vcstool
 
     IFS=" " read -r -a parts <<< "$(ici_resolve_scheme "$url")" # name, type, url, version
 
@@ -111,7 +111,7 @@ function ici_import_file {
         bsdtar -C "$sourcespace" -xf "$file"
         ;;
     *)
-        ici_install_pkgs_for_command vcs "${PYTHON_VERSION_NAME}-vcstool"
+        ici_install_pkgs_for_command vcs python3-vcstool
         ici_setup_git_client
         ici_vcs_import "$sourcespace" < "$file"
     ;;
@@ -132,7 +132,7 @@ function ici_import_url {
         processor=(bsdtar -C "$sourcespace" -xf-)
         ;;
     *)
-        ici_install_pkgs_for_command vcs "${PYTHON_VERSION_NAME}-vcstool"
+        ici_install_pkgs_for_command vcs python3-vcstool
         ici_setup_git_client
         processor=(ici_vcs_import "$sourcespace")
     ;;


ipa-mdl:
I am not sure if I understand the issue..
Your change enforces `python3-vcstool`, which is not what was intended.

ipa-jba:
The issue is that python-vcstool cannot be installed in Ubuntu:bionic (anymore):

```
 The following packages have unmet dependencies:
  python-vcstool : Depends: python2 (< 2.8) but it is not installable
                   Depends: python2 (>= 2.7) but it is not installable
                   Depends: python2:any (>= 2.6.6-7~) but it is not installable
```

Feel free to propose another fix, we are currently running CI from my fork, since a broken CI impacts development process significantly.

ipa-mdl:
Ah, I see :)
For some reason Travis did not notify me about the failing master cron jobs..

I will merge this, but rephrase it shlightly.
The only downside is that it pulls in python3, but I don't see another solution for now.

gavanderhoorn:
Could this change be the cause of https://github.com/ros-industrial-consortium/tesseract/pull/292 failing?

The error message is:

```
Starting function 'setup_upstream_workspace'
E: Failed to fetch http://packages.ros.org/ros/ubuntu/pool/main/p/python3-vcstool/python3-vcstool_0.2.8-1_all.deb  404  Not Found [IP: 64.50.236.52 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

That file indeed does not exist.

But seeing as the version is the only difference, I'm thinking this may be something else.

ipa-jba:
Since I started this PR I had a (really short) look into it. You are trying to download the version 0.2.8-1 from the `ROS_REPO=ros` but the version is currently available only in the ros-testing repository.

In the package index of "ros"(http://packages.ros.org/ros/ubuntu/dists/bionic/main/binary-amd64/Packages) the link points to 0.2.7.

I'm not sure whats wrong here, but since the same versions are used for python-vcstool (0.2.8 in testing, 0.2.7 in "ros") I'm with you in assuming there is another issue

Briancbn:
@gavanderhoorn I faced this error multiple times in travis-ci yesterday as well, not just `python3-vcstool`, but also `python3-catkin-pkg ` [job1](https://travis-ci.com/github/ros-industrial/industrial_ci/jobs/335758370) [job2](https://travis-ci.com/github/ros-industrial/industrial_ci/jobs/335758376) [job3](https://travis-ci.com/github/ros-industrial/industrial_ci/jobs/335758384)

but after restarting these job in my own travis-ci, the problem doesn't appear.

gavanderhoorn:
I'm getting rather confused: retrying the URL (and looking at the directory index) now does work.

Somehow the missing file is now there.

Looks to me to be unrelated to this PR.

ipa-mdl:
> Looks to me to be unrelated to this PR.

Yes, I guess the ROS repository server had a problem with the sync..

gavanderhoorn:
Looks like there are more ppl who run into this: [Different versions of python3-catkin-pkg-modules are distributed on packages.ros.org](https://answers.ros.org/question/352605) on ROS Answers.

PR #524, opened 2020-06-07
implement embeddable hooks

ipa-mdl:


industrial_ci/src/util.sh
```diff
@@ -76,11 +76,17 @@ function _sub_shell() (
...
+  local name_embed="${name}_EMBED"
 
   local script=${!name}
   if [ -n "$script" ]; then
     ici_run "$1" _sub_shell "$script"
   fi
+
+  local script_embed=${!name_embed}
+  if [ -n "$script_embed" ]; then
+    ici_run "$1" eval "$script_embed"
```
ipa-mdl, line 88:
```suggestion
    ici_run "$1_embed" eval "$script_embed"
```

ipa-mdl:
The idea is to offer `<HOOK>_EMBED` next to  `<HOOK>`.
If both are present, first non-embed variant will be run first.

Example:

```
AFTER_INIT="[install framework..]"
AFTER_INIT_EMBED="source /opt/framework/sourceme.bash param1 param2"
````

**Comment by tylerjw** [comment]
I tested locally and this works except for one weird sideafect... with this config:

```
    AFTER_INIT="./.ci/install_dependencies.sh" \
    AFTER_INIT_EMBED="source /opt/intel/openvino/bin/setupvars.sh" \
    AFTER_AFTER_INIT="./.ci/setup_known_hosts.sh"
```

It runs in this order:
1. AFTER_INIT
2. AFTER_AFTER_INIT
3. AFTER_INIT_EMBED

It is just fine for my use-case, it is just not the order I would expect.

ipa-mdl:
> It is just fine for my use-case, it is just not the order I would expect.

The order is wrong..

ipa-mdl:
@tylerjw: please try again

tylerjw:
Works as expected now!  Thank you.  This will greatly simplify my workflow for local testing.

PR #539, opened 2020-07-03
readme: mention noetic

gavanderhoorn:
As per subject.

It's supported since #499.


gavanderhoorn:
Not sure why the shellcheck tests fail.

Briancbn:
I am facing the same problem. Not sure why.
It seems that the shellcheck version on alpine docker is 0.7.1 same as my com, but it gives out this error strangely

Adding additional disable comments help solve the error.
```bash
# shellcheck disable=SC1090
# shellcheck disable=SC1091
source "$1"/setup.bash
```

PR #540, opened 2020-07-05
Fix Shellcheck error

Briancbn:
Signed-off-by: Chen Bainian <brian97cbn@gmail.com>
Fix Shellcheck error in Github Action


Briancbn:
Shellcheck fails in Travis CI (shellcheck 0.6.0). Still need `shellcheck disable=SC1090`.

ipa-mdl:
Ah okay, I will fix it later.. it should go into one line then..

ipa-mdl:
Or we can just drop the shellcheck job from Travis CI

PR #543, opened 2020-07-05
Update libseccomp2 to fix statx syscall error

### LINKED ISSUE #542: Travis CI Foxy: RCC (the QT resource compiler) couldn’t find files
Here is my compilation error log
```
AutoRcc subprocess error

------------------------

The rcc process failed to compile

  "SRC:/resources/resource.qrc"

into

  "BIN:/traffic-editor_autogen/3YJK5W5UP7/qrc_resource.cpp"

Command

-------

/usr/lib/qt5/bin/rcc -name resource -o /root/upstream_ws/build/traffic_editor/traffic-editor_autogen/3YJK5W5UP7/qrc_resource.cpp /root/upstream_ws/src/traffic_editor/traffic_editor/resources/resource.qrc

Output

------

RCC: Error in '/root/upstream_ws/src/traffic_editor/traffic_editor/resources/resource.qrc': Cannot find file 'icons/vertex.svg'
```

It seems that qt5 rcc fails in both Foxy main and testing and it is happening to multiple packages of mine.
I tried the Foxy docker image locally, there isn't a compilation problem. 
Same problem doesn't occur in Gitlab CI either. I suspect it is due to something related to the docker image used by Travis.

Briancbn:
Signed-off-by: Chen Bainian <brian97cbn@gmail.com>
Fix #542 
This problem has been bugging me for some time since the foxy-nightly docker
This fix will update libseccomp2 to higher than 2.3.3.
This will enable the docker-ce 18.06.0 run in Travis CI to allow statx syscall, which will fix the Qt5 RCC not found error in Foxy.
Reference: https://travis-ci.community/t/docker-build-environments/7216/4

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -27,7 +27,7 @@ export PYTHONUNBUFFERED=${PYTHONUNBUFFERED:1}
 export _FOLDING_TYPE=travis
 
 # Update libseccomp to allow statx syscalls
-(sudo apt update && sudo apt install -y libseccomp2) &> "$(mktemp)"
+sudo apt-get update -qq && sudo apt-get install -y -qq libseccomp2
 
 if [ "$ABICHECK_MERGE" = "auto" ]; then
   export ABICHECK_MERGE=false
--- a/travis.sh
+++ b/travis.sh
@@ -26,6 +26,9 @@ export TARGET_REPO_NAME=${TRAVIS_REPO_SLUG##*/}
 export PYTHONUNBUFFERED=${PYTHONUNBUFFERED:1}
 export _FOLDING_TYPE=travis
 
+# Update libseccomp to allow statx syscalls
+(sudo apt update && sudo apt install -y libseccomp2) &> "$(mktemp)"
+
 if [ "$ABICHECK_MERGE" = "auto" ]; then
   export ABICHECK_MERGE=false
   [ "$TRAVIS_PULL_REQUEST" = "false" ] || ABICHECK_MERGE=true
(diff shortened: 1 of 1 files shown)


travis.sh
```diff
@@ -26,6 +26,9 @@ export TARGET_REPO_NAME=${TRAVIS_REPO_SLUG##*/}
 export PYTHONUNBUFFERED=${PYTHONUNBUFFERED:1}
 export _FOLDING_TYPE=travis
 
+# Update libseccomp to allow statx syscalls
+(sudo apt update && sudo apt install -y libseccomp2) &> "$(mktemp)"
```
ipa-mdl, line 30: Why do you forward it to a temporary file?
And why do yu need to to update first?

ipa-mdl, line 30:
```suggestion
sudo apt-get -qq install -y libseccomp2"
```

Briancbn: Without updating libseccomp2 will be at 2.3.2, which will block statx syscall.

ipa-mdl: I meant the "apt-get update" :)

ipa-mdl: Not all versions of apt-get accept "-qq" after install, if I recall it correctly

ipa-mdl: Just read that Travis does not update automatically (any more?)

Briancbn: Not that I know of. It is included inside the man page tho.

ipa-mdl:
I will merge this right after #544

PR #555, opened 2020-08-31
Install catkin_lint from pip

### LINKED ISSUE #554: outdated catkin_lint
It seems like there's no way to specify a version of catkin lint and that industrial ci always uses the version from python, which is over 2 years old. According to the [catkin_lint repo](https://github.com/fkie/catkin_lint), its recommended to use the versions provided in the [official PPA](https://packages.ubuntu.com/catkin-lint)(which contain the most updated version).

wmmc88:
Closes #554 by install catkin_lint from pip rather than apt.


ipa-mdl:
Thanks for your contribution!
In general, it is a good idea to switch to the officially recommended version.
I am hesitating now , because it might make a lot of builds fail out of the blue.
Our test is failing as well now..

wmmc88:
Test 20's failure has been present in master(for the last 10 days) and is unrelated to my change. Looks like it's trying to install python3-wxgtk4 when it doesnt exist for kinetic (Ubuntu 16.04).

The other two failing tests are actually do the catkin lint catching issues that it previously missed(due to using outdated version). Personally, I think that a lot of builds "suddenly failing" should not be a problem because people's CI should not be pinned to master branch.... it should be pinned to a certain commit of industrial ci (ex. A release tag). This is the common method used in github actions (and I assume should be best practice for other ci). Repo maintainers should be upgrading the version of industrial ci in their ci config files whenever they see fit, as opposed to just blindly using master (and possibly causing random build failures from that). Perhaps I should change the docs to reflect that? Thoughts on this?

I could introduce a CATKIN_LINT_VERSION parameter with like RECOMMENDED and LEGACY, so that the legacy behaviour could be used if someone chose to? I wouldn't recommend doing this though because even the version of catkin lint shipped with melodic is two years old and riddled with many bugs. In fact a lot of false positive from catkin_lint in ubuntu 18.04 is actually what prompted me to make this update.

ipa-mdl:
> Test 20's failure has been present in master(for the last 10 days) and is unrelated to my change.

Yes, I saw that, but it is not what I meant.
I will have a look.

> The other two failing tests are actually do the catkin lint catching issues that it previously missed(due to using outdated version).

So these have to be fixed in this PR.

> Personally, I think that a lot of builds "suddenly failing" should not be a problem because people's CI should not be pinned to master branch.... it should be pinned to a certain commit of industrial ci (ex. A release tag).

This is not our approach. We recommend using the master, because we are constantly reacting to upstream changes.

> I could introduce a CATKIN_LINT_VERSION parameter with like RECOMMENDED and LEGACY, so that the legacy behaviour could be used if someone chose to?

That should not be necessary. The current approach is to use the version bundled with ROS, because most developers use this version. Or used it during development for an outdated, but still maintained version.
Just imagine to lint old `indigo` code with the latest version of `catkin_lint`.

A different version can be installed in an `AFTER_INIT_SCRIPT`. In this case, just the PPA needs to be added.

> I wouldn't recommend doing this though because even the version of catkin lint shipped with melodic is two years old and riddled with many bugs.

ROS melodic itself is two years old..

> In fact a lot of false positive from catkin_lint in ubuntu 18.04 is actually what prompted me to make this update.

If the version is that outdated, you could ask the maintainer for an update as well.
If there is a strong demand, we can make the latest version the default.
For now, I could offer a "ici_add_ppa" command that makes it easy to setup PPAs in the `AFTER_INIT_SCRIPT`

gavanderhoorn:
@wmmc88 wrote:
> even the version of catkin lint shipped with melodic is two years old

There seems to be some misunderstanding here.

There is no "version shipped with Melodic" (nor with any other version of ROS).

`catkin_lint` is a stand-alone Python package which was imported from time-to-time by the OSRF repository maintainers *for convenience*. It's not a ROS package, so there is no `ros-kinetic-catkin-lint` or `ros-melodic-catkin-lint`, only `python-catkin-lint`. And that package is built by @roehling in his PPA.

See ros/rosdistro#24251.

industrial_ci/src/tests/source_tests.sh
```diff
@@ -23,6 +23,8 @@
 # (As of version 0.4.4 most of them are defined in env.sh).
 
 function install_catkin_lint {
+    ici_apt_install software-properties-common
```
ipa-mdl, line 26:
```suggestion
    ici_install_pkgs_for_command apt-add-repository software-properties-common
```

industrial_ci/src/tests/source_tests.sh
```diff
@@ -23,6 +23,8 @@
...
+    ici_asroot apt-add-repository ppa:roehling/ros -y -u # -u for versions pre-Ubuntu 18.04
```
ipa-mdl, line 27: This could be made a function (in workspace.sh): `ici_add_apt_repo ppa:roehling/ros`
This would even work from any hook script.

industrial_ci/src/tests/source_tests.sh
```diff
@@ -23,6 +23,8 @@
 # (As of version 0.4.4 most of them are defined in env.sh).
 
 function install_catkin_lint {
+    ici_install_pkgs_for_command apt-add-repository software-properties-common
+    ici_asroot apt-add-repository ppa:roehling/ros -y -u # -u for versions pre-Ubuntu 18.04
```
ipa-mdl, line 27: using a PPA would fail for Debian

industrial_ci/src/tests/source_tests.sh
```diff
@@ -23,6 +23,8 @@
...
     ici_install_pkgs_for_command catkin_lint "${PYTHON_VERSION_NAME}-catkin-lint"
```
ipa-mdl, line 28: From Ubuntu 20.04 on the package was renamed `catkin-lint`.

ipa-mdl (changes requested):
This feature would need to handle a lot of cases.
And it would break all Debian builds that configure `catkin-lint`.

industrial_ci/src/tests/source_tests.sh
wmmc88: I could just install it from pip. would that be okay? I know generally this is frowned upon for ros, but since this isn't a ros package and is in an isolated ci environment, I think it'd be a valid solution to fix the debian problem.

ipa-mdl: You mean, you want to revert https://github.com/ros-industrial/industrial_ci/pull/448? ;)
But the main problem persists: The user might use the catkin_lint version from the ROS repository..
Right now I do not see a string demand for the the latest catkin_lint version.

And you can always install it with pip, e.g. in `AFTER_INIT`.
`ici_install_pkgs_for_command` will not install anything, if the command is already available.
It just exists for the purpose of installing a different version beforehand.

gavanderhoorn: > Right now I do not see a strong demand for the the latest catkin_lint version.

I could add a :+1: here (as in: I'd like to use the latest version).

It's the reason I opened ros/rosdistro#24251.

There's little value in having a linter available if you don't have access to the latest version.

ipa-mdl: I am open to reverting #448 (for master only), but someone needs to fix the catkin_lint issues and document it properly ;)
But we might want to run a test beforehand to see how many regressions we will get.

As an alternative, `AFTER_INIT='pip install catkin-lint'` (not tested..) should work for everyone right now.

ipa-mdl: `AFTER_INIT='pip install catkin-lint'` would be an opt-in, `ADDITIONAL_DEBS=python-catkin-lint` would be an opt-out (if we make pip the default).

gavanderhoorn: >  but someone needs to fix the catkin_lint issues and document it properly ;)

Which issues do you refer to here?

> As an alternative, `AFTER_INIT='pip install catkin-lint'` (not tested..) should work for everyone right now.

I agree, but seeing as `CATKIN_LINT` is something users can "just enable", it seems like not too strange for them to expect that would lint their packages using the latest released version.

Right now it doesn't do that, but only because OR has stopped importing the `.deb` from @roehling's PPA. That's not very clear.

wmmc88: I can change this pr to instead revert #448, fix catkin_lint issues in tests triggered by the new version, and update docs with the addition deb opt-out. 

I'm very busy rn, so I'll push this till the weekend 😅

ipa-mdl: > This opt-out needs a ici_install_pypi_for_command (or some better name) to not install catkin_lint, if it is already installed.

I doesn't have to be a separate command for now, but it would be appreciated..

industrial_ci/src/tests/source_tests.sh
roehling: If it makes things easier, there's a transitional `python3-catkin-lint` package that pulls `catkin-lint`. The package is guaranteed to be available in the next Debian stable release Bullseye (to provide an upgrade path from Buster), so all Ubuntu releases until at least Ubuntu 21.04 will still have that package, too (assuming the next Debian release happens mid-2021).

ipa-mdl: I think for now we will use thy pypi version, because it should work for all OS (Debian/Ubuntu, back to Ubuntu trusty).

I am not 100% familiar with the Debian update policy, but if I recall it correctly only bug fixes are allowed for stable distros.
So, the user would - again - not get the latest version of `catkin-lint` after some time, correct?

roehling: That is true. PyPI is the best choice if you're looking for the latest version in an OS independent manner. My comment was mostly aiming at my PPA, which admittedly only works for Ubuntu releases.

roehling:
While you're still working on this, I'd like to point out the support for [configuration files](https://github.com/fkie/catkin_lint/blob/master/catkin_lint.pod#configuration-file-format) that has been added in version 1.6.9, which might be useful for you.

wmmc88:
> While you're still working on this, I'd like to point out the support for [configuration files](https://github.com/fkie/catkin_lint/blob/master/catkin_lint.pod#configuration-file-format) that has been added in version 1.6.9, which might be useful for you.

I believe that configuration files in a repo should be picked up automatically since it will appear on the ROS PACKAGE PATH.

industrial_ci/src/tests/source_tests.sh
wmmc88: @ipa-mdl Is what I most recently pushed match what you're looking for?

wmmc88:
@ipa-mdl Is there anything that I need to change in this pr?

industrial_ci/mockups/legacy_catkin_lint_testpkg/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+cmake_minimum_required(VERSION 2.8.3)
```
ipa-mdl, line 1: This additional package is not needed..
There are other ways to check the version.

industrial_ci/src/workspace.sh
```diff
@@ -85,6 +89,11 @@ function ici_install_pkgs_for_command {
   ici_exec_for_command "$command" ici_apt_install "$@"
 }
 
+function ici_install_python_pkgs_for_command {
```
ipa-mdl, line 92: python package is ambiguous, I would prefer `ici_install_pypi_pkgs_for_command`

industrial_ci/src/workspace.sh
```diff
@@ -85,6 +89,11 @@ function ici_install_pkgs_for_command {
...
+    local command=$1; shift
+  ici_exec_for_command "$command" ici_pip_install "$@"
```
ipa-mdl, line 94: whitespace

ipa-mdl (changes requested):
Only some minor issues left :+1: 
I will try to work on this tomorrow and push some changes.

industrial_ci/src/workspace.sh
```diff
@@ -85,6 +89,11 @@ function ici_install_pkgs_for_command {
   ici_exec_for_command "$command" ici_apt_install "$@"
 }
 
+function ici_install_python_pkgs_for_command {
+    local command=$1; shift
```
ipa-mdl, line 93:
```suggestion
  local command=$1; shift
```

industrial_ci/src/workspace.sh
ipa-mdl, line 92:
```suggestion
function ici_install_pypi_pkgs_for_command {
```

industrial_ci/src/tests/source_tests.sh
```diff
@@ -23,7 +23,7 @@
 # (As of version 0.4.4 most of them are defined in env.sh).
 
 function install_catkin_lint {
-    ici_install_pkgs_for_command catkin_lint "${PYTHON_VERSION_NAME}-catkin-lint"
+    ici_install_python_pkgs_for_command catkin_lint "catkin-lint"
```
ipa-mdl, line 26:
```suggestion
    ici_install_pypi_pkgs_for_command catkin_lint "catkin-lint"
```

doc/index.rst
```diff
@@ -167,7 +167,7 @@ Note that some of these currently tied only to a single option, but we still lea
...
-* **CATKIN_LINT** (default: not set. Value range: [true|pedantic]): If ``true``, run `catkin_lint <http://fkie.github.io/catkin_lint/>`__ with ``--explain`` option. If ``pedantic``, ``catkin_lint`` command runs with ``--strict -W2`` option, i.e. more verbose output will print, and the CI job fails if there's any error and/or warning occurs.
+* **CATKIN_LINT** (default: not set. Value range: [true|pedantic]): If ``true``, run `catkin_lint <http://fkie.github.io/catkin_lint/>`__ with ``--explain`` option. If ``pedantic``, ``catkin_lint`` command runs with ``--strict -W2`` option, i.e. more verbose output will print, and the CI job fails if there's any error and/or warning occurs. Industrial CI uses the `latest version available from pypi <https://pypi.org/project/catkin-lint/>`__. If the older version in the `ros repository <http://packages.ros.org/ros/ubuntu/pool/main/c/catkin-lint/>`__ is required, :code:`AFTER_INIT='ici_install_pkgs_for_command catkin_lint "${PYTHON_VERSION_NAME}-catkin-lint"'` can be added to the CI Config.
```
ipa-mdl, line 170:
```suggestion
* **CATKIN_LINT** (default: not set. Value range: [true|pedantic]): If ``true``, run `catkin_lint <http://fkie.github.io/catkin_lint/>`__ with ``--explain`` option. If ``pedantic``, ``catkin_lint`` command runs with ``--strict -W2`` option, i.e. more verbose output will print, and the CI job fails if there's any error and/or warning occurs. Industrial CI uses the `latest version available from pypi <https://pypi.org/project/catkin-lint/>`__. If the older version in the `ros repository <http://packages.ros.org/ros/ubuntu/pool/main/c/catkin-lint/>`__ is required, :code:`ADDITIONAL_DEBS='python-catkin-lint"'` can be added to the CI Config.
```

industrial_ci/mockups/industrial_ci_testpkg/CMakeLists.txt
```diff
@@ -7,7 +7,9 @@ endif()
 
 find_package(catkin REQUIRED COMPONENTS rospy_tutorials rostest)
 
-catkin_package()
+catkin_package(
+  CATKIN_DEPENDS rospy_tutorials
```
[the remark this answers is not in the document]
wmmc88: this was one of the things that the newest catkin_lint complained about. Not sure if its necessary

doc/index.rst
```diff
@@ -167,7 +167,7 @@ Note that some of these currently tied only to a single option, but we still lea
...
-* **CATKIN_LINT** (default: not set. Value range: [true|pedantic]): If ``true``, run `catkin_lint <http://fkie.github.io/catkin_lint/>`__ with ``--explain`` option. If ``pedantic``, ``catkin_lint`` command runs with ``--strict -W2`` option, i.e. more verbose output will print, and the CI job fails if there's any error and/or warning occurs.
+* **CATKIN_LINT** (default: not set. Value range: [true|pedantic]): If ``true``, run `catkin_lint <http://fkie.github.io/catkin_lint/>`__ with ``--explain`` option. If ``pedantic``, ``catkin_lint`` command runs with ``--strict -W2`` option, i.e. more verbose output will print, and the CI job fails if there's any error and/or warning occurs. Industrial CI uses the `latest version available from pypi <https://pypi.org/project/catkin-lint/>`__. If the older version in the `ros repository <http://packages.ros.org/ros/ubuntu/pool/main/c/catkin-lint/>`__ is required, :code:`ADDITIONAL_DEBS='python-catkin-lint"'` can be added to the CI Config.
```
ipa-mdl, line 170:
```suggestion
* **CATKIN_LINT** (default: not set. Value range: [true|pedantic]): If ``true``, run `catkin_lint <http://fkie.github.io/catkin_lint/>`__ with ``--explain`` option. If ``pedantic``, ``catkin_lint`` command runs with ``--strict -W2`` option, i.e. more verbose output will print, and the CI job fails if there's any error and/or warning occurs. Industrial CI uses the `latest version available from pypi <https://pypi.org/project/catkin-lint/>`__. If the older version in the `ros repository <http://packages.ros.org/ros/ubuntu/pool/main/c/catkin-lint/>`__ is required, :code:`ADDITIONAL_DEBS='python-catkin-lint'` can be added to the CI Config.
```

industrial_ci/mockups/industrial_ci_testpkg/CMakeLists.txt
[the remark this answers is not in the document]
ipa-mdl: This is almost impossible to do right with package format 1..
I just migrated it to packages format 2.

ipa-mdl (approved):
@wmmc88: Please review my changes.
If you approve, I will squash-merge after the test jobs have passed.

wmmc88 (commented):
@ipa-mdl LGTM!

doc/index.rst
```diff
@@ -167,7 +167,7 @@ Note that some of these currently tied only to a single option, but we still lea
...
-* **CATKIN_LINT** (default: not set. Value range: [true|pedantic]): If ``true``, run `catkin_lint <http://fkie.github.io/catkin_lint/>`__ with ``--explain`` option. If ``pedantic``, ``catkin_lint`` command runs with ``--strict -W2`` option, i.e. more verbose output will print, and the CI job fails if there's any error and/or warning occurs.
+* **CATKIN_LINT** (default: not set. Value range: [true|pedantic]): If ``true``, run `catkin_lint <http://fkie.github.io/catkin_lint/>`__ with ``--explain`` option. If ``pedantic``, ``catkin_lint`` command runs with ``--strict -W2`` option, i.e. more verbose output will print, and the CI job fails if there's any error and/or warning occurs. Industrial CI uses the `latest version available from pypi <https://pypi.org/project/catkin-lint/>`__. If the older version in the `ros repository <http://packages.ros.org/ros/ubuntu/pool/main/c/catkin-lint/>`__ is required, :code:`ADDITIONAL_DEBS='python-catkin-lint'` can be added to the CI Config.
```
gavanderhoorn, line 170: IIUC, this:

> If the older version in the `ros repository <http://packages.ros.org/ros/ubuntu/pool/main/c/catkin-lint/>`__ is required

will not be true any more in the near future.

"the latest version" of `catkin_lint` will be imported into the ROS package repositories again.

https://github.com/ros-infrastructure/reprepro-updater/pull/81

ipa-mdl: > "the latest version" of catkin_lint will be imported into the ROS package repositories again.

But not for EOL distros, and it might lag behind again until another PR is filed and merged.
Actually, the upcoming import made me merge this PR right away..

> will not be true any more in the near future.

Do do have an idea how we could rephrase it to be more accurate?

Issue #459, opened 2019-12-29
Skip launching docker image


tylerjw:
I would like to let gitlab handle pull the image and running CI inside it.  This might already exist and I just haven't figured out how to do it.

----------
Edit:
Reading the code I found the test around the variable `IN_DOCKER`.  If my usecase is normal enough maybe I could make a small PR to the documentation to document setting it to skip launching docker?

ipa-mdl:
> Reading the code I found the test around the variable IN_DOCKER

Please be aware that setting `IN_DOCKER=false` bypasses some code, e.g. for setting up the ROS repository.

I have plans to refactor this part of the code.
Eventually, `IN_DOCKER` should get replaced by something else as well.
I could even imagine that we introduce an option to choose the "executor".

>  If my usecase is normal enough maybe I could make a small PR to the documentation to document setting it to skip launching docker?

`IN_DOCKER` is not documented, because it could lead to unintended side effects, especially if someone would use it locally or even worse with - for some reason - password-less sudo.

I would prefer a more fail-safe option: some special handing in `gitlab.sh`.
For example a flag, which would test if the scripts is run in a disposable environment and would then set `IN_DOCKER=true`.

tylerjw:
One more comment on this is that I've found it useful to use `IN_DOCKER` locally to allow me to manage the docker launch and some environment issues I haven't found a way around otherwise.  For context here is my notes on doing that:

```bash
# pull and open docker image
docker pull moveit/moveit:melodic-ci
docker run -it -v $HOME/.ssh:/root/.ssh-keys:ro -v $HOME/.ccache:/root/.ccache --cap-add=SYS_PTRACE --security-opt seccomp=unconfined moveit/moveit:melodic-ci /bin/bash

# configure .ssh folder
mkdir -p ~/.ssh
cp -r /root/.ssh-keys/* ~/.ssh/
chown -R $(id -u):$(id -g) ~/.ssh

# Update and Setup enviroment
apt update -qq
apt dist-upgrade -qq --yes
apt install -qq --yes openssh-client apt-utils gnupg wget ca-certificates lsb-release dirmngr vim
echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
eval $(ssh-agent -s) && ssh-add ~/.ssh/id_rsa

# Setup test workspace
CATKIN_WS=$HOME/tmp/ws_test/
mkdir -p $CATKIN_WS/src
cd $CATKIN_WS/src
[...]

PR #557, opened 2020-09-01
Add option to disable running industrial_ci in a docker image

lsolanka:
This might sound strange at first sight, but sometimes there's no need to run the whole build in a docker container spawned by `industrial_ci` itself. This is because many pipelines already run inside a docker container which is configurable by the user.

Also, there's another, technical reason, related to bitbucket pipelines: In pipelines when run as a docker service (i.e. docker inside docker), the support for ccache doesn't work because it is not possible to mount a pipeline build cache from outside the repository clone directory into the `industrial_ci` docker container. This is because security restrictions in pipelines prevent this: it is only possible to mount anything into a docker container from inside the `$BITBUCKET_CLONE_DIR`, but that is already mounted read-only in `industrial_ci`.

Hope the explanation is clear, I'm curious if this option would make sense to be merged.

### DIFF
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -101,7 +101,7 @@ function run_source_tests {
     # shellcheck disable=SC1090
     source "${ICI_SRC_PATH}/builders/$BUILDER.sh" || ici_error "Builder '$BUILDER' not supported"
 
-    if [ -z "$DO_NOT_RUN_IN_DOCKER" ]; then
+    if [ -z "$DISABLE_DOCKER_FOR_SOURCE_TESTS" ]; then
         ici_require_run_in_docker # this script must be run in docker
     fi
     upstream_ws=~/upstream_ws
--- a/industrial_ci/src/tests/source_tests.sh
+++ b/industrial_ci/src/tests/source_tests.sh
@@ -101,7 +101,9 @@ function run_source_tests {
     # shellcheck disable=SC1090
     source "${ICI_SRC_PATH}/builders/$BUILDER.sh" || ici_error "Builder '$BUILDER' not supported"
 
-    ici_require_run_in_docker # this script must be run in docker
+    if [ -z "$DO_NOT_RUN_IN_DOCKER" ]; then
+        ici_require_run_in_docker # this script must be run in docker
+    fi
     upstream_ws=~/upstream_ws
     target_ws=~/target_ws
     downstream_ws=~/downstream_ws


ipa-mdl:
> Hope the explanation is clear

Mostly :)

>  it is only possible to mount anything into a docker container from inside the $BITBUCKET_CLONE_DIR, but that is already mounted read-only in industrial_ci.

Gitlab has similar restrictions. Did you try using a subdirectory as `CCACHE_DR`?

> I'm curious if this option would make sense to be merged

No, at least not now.
For two reasons:
1. it already exists (as an undocumented, unsupported feature)
2. bypassing `ici_require_run_in_docker` might have unintended side-effects, if used in a wrong way

I would rather like to split the processes (inside/outside Docker) properly. (see https://github.com/ros-industrial/industrial_ci/issues/459 as well).
Eventually, I want to get rid of the Docker build step completely (almost there!) and then make industrial_ci runnable in any Debian-based container (like we do it for Github Actions already), with some optional bootstrapping (for non-Docker services like Travis).

So in the future there should be no need for such a flag anymore..
If you really need to skip the Docker part now, just set  `IN_DOCKER=true` to bypass it.

https://github.com/ros-industrial/industrial_ci/blob/d5a4d278f03fb8a3da6ec992db4766821a951b8d/industrial_ci/src/docker.sh#L36

tylerjw:
I run this in gitlab ci and have gotten around docker using IN_DOCKER.  For ccache I did it this way:

```
  cache:
    key: project_ccache  # rename this
    paths:
      - .ccache/
  script:
    # industrial_ci cache settings
    - export CCACHE_DIR=$PWD/.ccache
    - mkdir -p $CCACHE_DIR
    - git clone --quiet --depth 1 https://github.com/ros-industrial/industrial_ci .industrial_ci -b master
    - .industrial_ci/gitlab.sh
```

ipa-mdl:
> Eventually, I want to get rid of the Docker build step completely (almost there!) 

Completed in #558

lsolanka:
> For two reasons:
> 
> 1. it already exists (as an undocumented, unsupported feature)
> 2. bypassing `ici_require_run_in_docker` might have unintended side-effects, if used in a wrong wa

Makes sense, the `IN_DOCKER` option somehow slipped my attention, even though I was looking straight at it. My bad!

lsolanka:
> I run this in gitlab ci and have gotten around docker using IN_DOCKER. For ccache I did it this way:
> 
> ```
>   cache:
>     key: project_ccache  # rename this
>     paths:
>       - .ccache/
>   script:
>     # industrial_ci cache settings
>     - export CCACHE_DIR=$PWD/.ccache
>     - mkdir -p $CCACHE_DIR
>     - git clone --quiet --depth 1 https://github.com/ros-industrial/industrial_ci .industrial_ci -b master
>     - .industrial_ci/gitlab.sh
> ```

Thanks, we use bitbucket and there was some issue with permissions due to docker userns remapping. But we managed to get around it.

There's no need for this PR, as I see it.

ipa-mdl:
There is now a new option ISOLATION=shell (https://github.com/ros-industrial/industrial_ci/pull/572).

PR #558, opened 2020-09-01
get rid of docker build

ipa-mdl:
This PR removes the docker build step.
Instead, everything that needs to be set up (ROS repos etc.) is handled in `ici_init_apt`.


industrial_ci/src/workspace.sh
```diff
@@ -60,9 +60,28 @@ function ici_pip_install {
...
+
+      ici_retry 3 eval "$keycmd"
+      echo "deb ${ROS_REPOSITORY_PATH} $(lsb_release -sc) main" > "/etc/apt/sources.list.d/ros${ROS_VERSION}-latest.list"
+    fi
+    sed -i "/^# deb.*multiverse/ s/^# //" /etc/apt/sources.list
+
     ici_asroot apt-get update -qq
 
-    local debs_default=(build-essential)
+    local debs_default=(apt-utils ca-certificates build-essential)
```
Briancbn, line 84: This is causing ca-certificates error in gitlab-ci private repo pipeline when setting up `ca-certificate`
log:
```
Setting up ca-certificates (20201027ubuntu0.18.04.1) ...
Updating certificates in /etc/ssl/certs...
ln: failed to create symbolic link '/etc/ssl/certs/ACCVRAIZ1.pem': Read-only file system
dpkg: error processing package ca-certificates (--configure):
 installed ca-certificates package post-installation script subprocess returned error exit status 1
```
gitlab config: note that `melodic` and `eloquent-no-rmf` will error
```yml
workflow:
  rules:
    - if: $CI_MERGE_REQUEST_ID
    - if: $CI_COMMIT_TAG
    - if: $CI_COMMIT_BRANCH

image: docker:git
services:
  - docker:dind
before_script:
  - apk add --update bash coreutils tar curl
  - git clone --quiet --depth 1 https://github.com/ipa-mdl/industrial_ci .industrial_ci -b feature/omni-init
cache:
  key: ${CI_JOB_NAME}
  paths: 
    - .ccache/

variables:
  TMPDIR: "${CI_PROJECT_DIR}.tmp"
  DOCKER_RUN_OPTS: "-v /etc/ssl/certs:/etc/ssl/certs:ro"
  CCACHE_DIR: "${CI_PROJECT_DIR}/.ccache"

melodic:
  variables:
    ROS_DISTRO: "melodic"
    ROS_REPO: "main"
  script: 
    - .industrial_ci/gitlab.sh

eloquent-no-rmf:
  variables:
    ROS_DISTRO: "eloquent"
  script:
    - .industrial_ci/gitlab.sh
  allow_failure: true

foxy-rmf:
  variables:
    ROS_DISTRO: "foxy"
    DOCKER_IMAGE: "registry.gitlab.com/rosi-ap/rosiap_docker_images:rmf-foxy-nightly"
    AFTER_INIT_EMBED: "source /opt/rmf/install/setup.bash"
  script: 
    - echo "xxxxxxxxxx" | docker login -u rosi --password-stdin registry.gitlab.com
    - .industrial_ci/gitlab.sh
```

Since the **DOCKER_RUN_OPTS** in the tutorial is set as `:ro`.

There is not problem with public repository in gitlab_ci.

Briancbn: Simply changing `DOCKER_RUN_OPTS` from `:ro` to `:rw` will fix the error for shared runner and my specific runner.

Also add in additional volumn paths inside `config.toml` will fix it.

Should I try using the **IN_DOCKER** option as well?

Briancbn:
**IN_DOCKER** runs well on **Gitlab CI** private repo with non-privileged specific runner setup, ccaching also works.
However, if you have **private docker image** that is in need of **access token or key** to download (like me). I currently have no solution.

This PR should solve #459 for **specific non-privileged docker gitlab runner**. Haven't tested on Travis and Github Action yet. I think this documentation should be one issue in it self as mentioned in https://github.com/ros-industrial/industrial_ci/issues/459#issuecomment-577937824 for potential safety reason.

Here is my **gitlab-ci.config** for melodic

```yml
before_script:
  - apt-get update
  - apt-get install -y git
  - git clone --quiet --depth 1 https://github.com/ipa-mdl/industrial_ci .industrial_ci -b feature/omni-init
cache:
  key: ${CI_JOB_NAME}
  paths: 
    - .ccache/

variables:
  IN_DOCKER: "true"
  CCACHE_DIR: "${CI_PROJECT_DIR}/.ccache"

melodic:
  image: ubuntu:18.04
  variables:
    ROS_DISTRO: "melodic"
    ROS_REPO: "main"
  script: 
    - .industrial_ci/gitlab.sh
```

Here is my config.yoml
```
[[runners]]
  name = "xxxxxxxxxx"
  url = "xxxxxx"
  token = "xxxxxxxx"
  executor = "docker"
  [runners.custom_builder_dir]
  [runners.cache]
    [runners.cache.s3]
    [runners.cache.gcs]
  [runners.docker]
    tls_verify = false
    image = "alpine:3.4"
    privileged = false
    disable_entrypoint_overwrite = false
    oom_kill_disable = false
    disable_cache = false
    volumes = ["/certs/client", "/cache"]
    shm_size = 0
```

industrial_ci/src/workspace.sh
ipa-mdl: >   DOCKER_RUN_OPTS: "-v /etc/ssl/certs:/etc/ssl/certs:rw"

I would advise against that!
This would update the host's certificates from the Docker container.
Why do you need this?

ipa-mdl: To be safe, I will rewrite it to only update/install the certificates, if /etc/ssl/certs is not populated

Briancbn: I now understand the problem that I was facing. 

I actually don't have a self-signed certificate inside `/etc/ssl/certs`, so I don't really need `DOCKER_RUN_OPTS: "-v /etc/ssl/certs:/etc/ssl/certs:ro"`. 

It didn't hurt to have this line inside in the past. During `docker build` you have access to the `/etc/ssl/certs` with the privileged runner. Now it is giving out error, due to this additional variable.

Actually, now I realize without the latest changes can work, as well. Sorry about the confusion. Hopefully, this won't break someone else's setup where **a self-signed certificate was used**, since I don't have existing setup to test with.

I think this is ready to merge.

PR #572, opened 2020-12-15
Implement ISOLATATION={docker,shell} option

ipa-mdl:
* ISOLATION={docker,shell}, works for Github Actions and GitLab CI
* alternatove TEST= option, as well for external test scripts
* BUILDER can now be external, too
* BUILDER can be set for ROS prerelease test
* `set -u` fos bash 4.4+ (earlier versions treat array differently)


ipa-mdl:
@Briancbn: please try with a non-privileged runner

ipa-mdl:
Tested  Github Action, Travis CI, Gitlab.com, custom Gitlab and run_ci.

ipa-mdl:
Documentation will follow later.

Briancbn:
Sorry, wasn't able to review this in time.

An error pop up after this PR for the ones using `AFTER_INIT_EMBED` to set custom source environment.

When it is sourcing the custom workspace in custom `DOCKER_IMAGE`. This error will pop up.

```
/opt/rmf/install/setup.bash: line 11: COLCON_TRACE: unbound variable
```

This seems to be the effect of `set -u` now by default in both the `AFTER_*` and `AFTER_*_EMBED` (they are exactly the same now), where variable `COLCON_TRACE` doesn't exist yet causing -u to exit.

If I source `/opt/ros/foxy/setup.bash`, the following error will occur

```
/opt/ros/foxy/setup.bash: line 8: AMENT_TRACE_SETUP_FILES: unbound variable
```

Solution 1, maybe set +u before running the embed script?

Solution 2, I will use `ici_source_setup` instead in my `AFTER_INIT`

ipa-mdl:
Ah, the new `-u` setting affects the  `*_EMBED` scripts.
While solution 2 is very clean and would allow hardening the rest of the scrips, it might not be possible to use it in all cases.
I will implement solution 1 for you to test.

ipa-mdl:
> I will implement solution 1 for you to test.

I had second thoughts and went on with a compromise (#581).
Sourcing a ROS environment is not a supported use case for the `EMBED`  scripts.

However, if someone has to enable unset variable for third-party scripts, `ici_with_unset_variables` can be used.

Briancbn:
> However, if someone has to enable unset variable for third-party scripts, `ici_with_unset_variables` can be used.

This is definitely the better choice.

PR #571, opened 2020-12-14
Port travis test cases to github actions

Briancbn:
Port tests from Travis to Github Actions
- [x] Enable wrap test for external repository
- [x] Add in test for ARM
- [x] Add in test for run_travis
- [x] Port all travis tests.

PR #576, opened 2020-12-17
migrate GitHub action to composite action (bash)

ipa-mdl:
A bash-based action should be easier to maintain.


ipa-mdl:
Tested: https://github.com/ipa-mdl/ros_canopen/runs/1576163323
Works as well with #571: https://github.com/ipa-mdl/industrial_ci/actions/runs/430234007

gavanderhoorn:
@ipa-mdl: seeing as I'm currently migrating lots of things to GH Actions: could you briefly describe the benefits of the changes in this PR? For users, that is?

Or are these purely developer-facing changes?

ipa-mdl:
> could you briefly describe the benefits of the changes in this PR? For users, that is?

None.

>Or are these purely developer-facing changes?

Yes.
It is just easier to maintain and allows us to add GitHub Action specific features / optimizations in the future.

The first version was written in Javascript, because the composite action was not available back then.

PR #579, opened 2021-01-12
configure EOL distros for prerelease

ipa-mdl:
as introduced in https://github.com/ros-infrastructure/ros_buildfarm/pull/684


ipa-mdl:
Does not solve the issue

ipa-mdl:
`ROSDISTRO_INDEX_URL` needs to be renamed.

PR #581, opened 2021-01-17
Deal with unset variables in EMBED scrips

ipa-mdl:
As reported in https://github.com/ros-industrial/industrial_ci/pull/572#issuecomment-761718057


ipa-mdl:
@Briancbn: please test with `ici_with_unset_variables`.

Briancbn:
> @Briancbn: please test with `ici_with_unset_variables`.

It works. Thanks!

PR #582, opened 2021-01-18
Script to parse github action workflows

floweisshardt:
This PR introduces a script that parses a github action workflow from `.github/workflows/industrial_ci_action.yml` and then calls `run_ci` with this configuration.
The script is the equivalent to `run_travis` and has the same terminal usage and arguments.

### DIFF
--- a/industrial_ci/src/util.sh
+++ b/industrial_ci/src/util.sh
@@ -305,7 +305,11 @@ function ici_parse_jobs {
       _ici_parse_jobs_res="$3";;
   "true")
       _ici_parse_jobs_res="0";;
+  "True")
+      _ici_parse_jobs_res="0";;
   "false")
+      _ici_parse_jobs_res="1";;
+  "False")
       _ici_parse_jobs_res="1";;
   *)
       if ! [[ "$_ici_parse_jobs_res" =~ ^[0-9]+$ ]]; then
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
I haven't had time to test it yet, but ever since I've come across [nektos/act](https://github.com/nektos/act) I've wanted to test it with `industrial_ci`.

agutenkunst:
@gavanderhoorn I tried it a while ago but had problems mounting the right folders. 
See https://github.com/nektos/act/issues/410 

But I think it is worth a second maybe more educated shot.

ipa-mdl:
`run_gitlab_actions` is a copy of `run_travis`. I would prefer to write a proper python module.
This could be used as well for Gitlab CI.

industrial_ci/src/util.sh
```diff
@@ -305,8 +305,12 @@ function ici_parse_jobs {
       _ici_parse_jobs_res="$3";;
   "true")
       _ici_parse_jobs_res="0";;
+  "True")
```
ipa-mdl, line 308: We use `true` and `false` everywhere else.
Instead of accepting the Python syntax, it should be converted in the calling script.

ipa-mdl (changes requested):

ipa-mdl:
> I would prefer to write a proper python module.

I migrated `run_travis` in #717

PR #588, opened 2021-01-22
Refactor handling of docker settings

ipa-mdl:
* `ici_forward_mount` to setup mount from variable or file name. ensures that path is absolute (@gavanderhoorn: FYI)
* `ici_forward_variable` passes variables down to the isolated environment

### DIFF
--- a/industrial_ci/src/isolation/docker.sh
+++ b/industrial_ci/src/isolation/docker.sh
@@ -20,15 +20,31 @@ export DOCKER_COMMIT_MSG=${DOCKER_COMMIT_MSG:-}
 export DOCKER_COMMIT_CREDENTIALS=${DOCKER_COMMIT_CREDENTIALS:-}
 export DOCKER_PULL=${DOCKER_PULL:-true}
 
-# ici_docker_forward_mount options VARNAME rw/ro [PATH]
-function ici_docker_forward_mount() {
-  local -n _ici_parse_mount_dir_res=$1
-  local p=${!2:-}
+# ici_forward_mount VARNAME/FILE rw/ro [PATH]
+function ici_forward_mount() {
+  local p=$1
+  local v=
+  if ! [ -e "$1" ]; then
+    v=$1
+    p=${!1:-}
+  fi
   if [ -n "$p" ]; then
     local p_abs
     p_abs=$(readlink -m "$p")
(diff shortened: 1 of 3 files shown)


industrial_ci/src/isolation/docker.sh
```diff
@@ -20,13 +20,40 @@ export DOCKER_COMMIT_MSG=${DOCKER_COMMIT_MSG:-}
 export DOCKER_COMMIT_CREDENTIALS=${DOCKER_COMMIT_CREDENTIALS:-}
 export DOCKER_PULL=${DOCKER_PULL:-true}
 
+# ici_forward_mount VARNAME/FILE rw/ro [PATH]
+function ici_forward_mount() {
```
gavanderhoorn, line 24: I expect it's all completely clear to you, but I'd really benefit from some more explanation of what this does, and when I should use it (if at all).

ipa-mdl: > when I should use it (if at all).

It is used implicitly for  variables that contain path, which have to be mounted into docker.
So it must be used in the industrial_ci code whenever a mount is needed.
It is an extension of #572, so the mount will be skipped for ISOLATION=shell.
Users don't have to do anything to enable it.

All path variables can now be specified relative to the CI build folder as well.

> some more explanation of what this does

`ici_forward_mount` collects the Docker options that are needed for the mount (`-v ...`) and for passing/rewriting the variable (`-e ...`). Beforehand these  options were composed individually all over the code...
As an additional feature, the path gets resolved to an absolute path as it is required by Docker.

gavanderhoorn: Great.

Good comments to add to the code? :)

ipa-mdl: > Good comments to add to the code? :)

Yes, but not now..
The comments of all functions are quite outdated, so it need a general overhaul.
And I would like to document these things for contributors.
There are actually two versions of `ici_forward_mount`, so the underlying isolation concepts needs to be documented as well.

Actually, this needs to be done for all new (internal) features since we branched off from `legacy`.

ipa-mdl: -> https://github.com/ros-industrial/industrial_ci/issues/589

Issue #590, opened 2021-01-23
Update Github Action template


ipa-mdl:
The current matrix expansion approach works, but it is not the intended usage and results in a warning.

![image](https://user-images.githubusercontent.com/727227/105580320-f825db80-5d8b-11eb-90bf-923ebb48f8f3.png)
> Matrix options must only contain primitive values

gavanderhoorn:
This should be equivalent:

```yaml
jobs:
  industrial_ci:
    strategy:
      matrix:
        ros_distro: [ melodic ]
        ros_repo: [ main, testing ]

    ...

    steps:
      - ...

      - uses: ros-industrial/industrial_ci@master
        env:
          ROS_DISTRO: ${{ matrix.ros_distro }}
          ROS_REPO: ${{ matrix.ros_repo }}

      - ...
```

I haven't really found a way yet to provide 'defaults' in the `matrix` section.

Could set `env.ROS_REPO` to `main` perhaps and then rely on the `matrix` expansion to override it when desired?

ipa-mdl:
Yes, this is what I had in mind.
I just don't like the redundancy..

We could extend the action to read the matrix variable and then deal with  the defaults (from `env.*`).

ipa-mdl:
@gavanderhoorn: 

How do you like this option (#591)?
```yaml
jobs:
  industrial_ci:
    strategy:
      matrix:
        ROS_DISTRP: [ melodic ]
        ROS_REPO: [ main, testing ]

    ...

    steps:
      - ...

      - uses: ros-industrial/industrial_ci@master
        with
          matrix: ${{ toJSON(matrix) }}        
      - ...
```

gavanderhoorn:
Hm.

I actually liked the fact you have to explicitly tell each step (and in this case the `industrial_ci` step) where the values come from.

It's *really* obvious reading the `.yml` that step gets two env vars: `ROS_DISTRO` and `ROS_REPO`, and their values come from the `matrix` stanza earlier in the file.

With your proposal, you have to understand how `toJSON(..)` works and what gets passed exactly.

That's of course OK, but if we can get something equivalent (and I believe the default way to do this is) and it's easier to grok, it may be preferable over saving a line (or two, or three ..).

[...]

PR #591, opened 2021-01-23
Add matrix input to Github Action

ipa-mdl:
```yaml
jobs:
  industrial_ci:
    strategy:
      matrix:
        ROS_DISTRO: [melodic]
        ROS_REPO: [testing, main]
    runs-on: ubuntu-latest
    steps:
      - ...
      - uses: 'ros-industrial/industrial_ci@master'
        with:
          matrix: ${{toJSON(matrix)}}
```

`$` can be escaped with `$$`.
Without escaping, the variable substitution will happen immediately.

### DIFF
--- a/.github/action.sh
+++ b/.github/action.sh
@@ -24,4 +24,10 @@ export TARGET_REPO_PATH=$GITHUB_WORKSPACE
 export TARGET_REPO_NAME=${GITHUB_REPOSITORY##*/}
 export _FOLDING_TYPE=github_actions
 
+if [ -n "$INPUT_CONFIG" ]; then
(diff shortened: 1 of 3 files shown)


ipa-mdl:
Not sure, if `matrix` is the right name. `config` might be better.

gavanderhoorn:
> `$` can be escaped with `$$`.
> Without escaping, the variable substitution will happen immediately.

can or *must*?

The latter would be unfortunate.

ipa-mdl:
> > $ can be escaped with $$.
> > Without escaping, the variable substitution will happen immediately.

> can or must?

Depends ;)
It works the same as for Gitlab CI.
For `BEFORE/AFTER_*` scripts you must escape it, for configuration variable early substitution  should be fine.

I could rewrite it to not support substitution at all, but then there might be no way to have substitution for Github Actions.

ipa-mdl:
I just renamed the input to `config` and decided to not change the template yet (to be discussed in #590).

PR #594, opened 2021-01-25
Initial support for automatic setup of GitLab CI credentials

stertingen:


ipa-mdl:
While it might not be necessary for Gitlab CI, such a credential helper might be good for Github access tokens.

ipa-mdl:
With #613, we could just prepare `.gitconfig` and `.gitcredentials` outside of the `industrial_ci` run (e.g. from gitlab.sh) and then just copy the credentials. 
I don't think that passing the token as an environment variable is more secure than storing it in a file (in the context of CI tests!).

If we automate this in the CI-specific entry point, then it still works with `run_ci`..

Issue #578, opened 2021-01-11
Collect JUnit XML test results for GitLab


stertingen:
GitLab has a feature to collect JUnit XML test results and display them in the Pipeline UI: https://docs.gitlab.com/ee/ci/unit_test_reports.html

Currently, this is not directly supported/documented by the CI, but the following `.gitlab-ci.yml` does the job:

```yaml
image: docker:git

variables:
  DOCKER_RUN_OPTS: "-v ${CI_PROJECT_DIR}/.build:/root/target_ws/build" # Share build space

services:
  - docker:19.03.5-dind

before_script:
  - apk add --update bash coreutils tar grep
  - git clone --quiet --depth 1 https://github.com/ros-industrial/industrial_ci .industrial_ci -b master

# Collect test results by default
default:
  artifacts:
    when: always
    reports:
      junit: ${CI_PROJECT_DIR}/.build/**/test_results/**/*.xml

kinetic:
  script: .industrial_ci/gitlab.sh
  variables:
    ROS_DISTRO: "kinetic"
```

It places the build space for the target workspaces in the outer docker container and mounts it into the inner container. The extra mount seems a little bit hacky to me, but it works fine.

I thought that others might profit from this, so this could be added to the README or the example config.

ipa-mdl:
Very nice!

>  The extra mount seems a little bit hacky to me

It is not too hacky.. just one of the downside of using Docker (or containers in general).
With https://github.com/ros-industrial/industrial_ci/pull/572, you can even avoid the mount :)

> I thought that others might profit from this, so this could be added to the README or the example config.

Definitely! I will add it to https://gitlab.com/ipa-mdl/industrial_ci
Anyway, the example config is a little bit outdated now and needs some overhaul (like most of the documentation..).

ipa-mdl:
I have now added support for mounting the workspaces automatically (https://github.com/ros-industrial/industrial_ci/pull/572).
```yaml
test_junit:
  script: .industrial_ci/gitlab.sh
  variables:
    ROS_DISTRO: noetic
    BASEDIR: ${CI_PROJECT_DIR}/.build
  artifacts:
    when: always
    reports:
      junit: ${BASEDIR}/**/test_results/**/*.xml
```

(results: https://gitlab.com/ipa-mdl/industrial_ci/-/pipelines/240764976/test_report)

stertingen:
> 
> 
> I have now added support for mounting the workspaces automatically (#572).
> 
> ```yaml
[...]

PR #595, opened 2021-01-27
Copy target repo if BASEDIR is set

ipa-mdl:
Fix BASEDIR issue with Docker (https://github.com/ros-industrial/industrial_ci/issues/578#issuecomment-767451862)

Test case (failing): https://gitlab.com/ipa-mdl/industrial_ci/-/jobs/989291921
Fix (passing): https://gitlab.com/ipa-mdl/industrial_ci/-/jobs/989415411


stertingen:
This is also relevant for shell isolation.

PR #609, opened 2021-02-03
configurable base_workspace

fmessmer:
we use docker images that already have some workspace set up on the image
we want to use that workspace in addition to `/opt/ros/$ROS_DISTRO` instead of using a source `UPSTREAM_WORKSPACE` (which has to be build first during the ici run)

thoughts?


ipa-mdl:
Good idea!

However, we cannot name it  `BASE_WORKSPACE`:
*  `_WORKSPACE` would imply the workspace definition options.
* `base_ws` is used in the ABI checks
* `/opt/ros/$ROS_DISTRO/` is not a workspace, but an install space.

One idea would be `ROS_ROOT`, but it might be set already for rosbuild compliancy.
Other idea is `EXTEND_FROM`, which would point to the install space.

Apart from the name, there are some minor issus, which can be addresses easily:
* default should be set in `run.sh`, because it depends on `ROS_DISTRO`
* it should be optional (not set if ROS_DISTRO is not set/empty)
* it must be used everywhere (not only source_tests.sh)
* documentation needs to be updated

fmessmer:
I rebased onto latest master and tried to consider your comments....I'm not sure I covered all of them, though

industrial_ci/src/tests/abi_check.sh
```diff
@@ -187,7 +187,11 @@ function run_abi_check() {
         export PATH="/usr/lib/ccache:$PATH"
     fi
 
-    extend="/opt/ros/$ROS_DISTRO"
+    if [ -n "$EXTEND_FROM" ]; then
```
ipa-mdl, line 190: this decision should be made in `run.sh`. And then just used in all of these places.

ipa-mdl (changes requested):

industrial_ci/src/tests/abi_check.sh
ipa-mdl:
```
if [ -n "$ROS_DISTRO:-" ]; then
    EXTEND_FROM=${EXTEND_FROM:-/opt/ros/$ROS_DISTRO}
fi
```
in run.sh

and then
```
extend=${EXTEND_FROM:?}
```

fmessmer: done

industrial_ci/src/run.sh
```diff
@@ -41,6 +41,10 @@ if [ "$DEBUG_BASH" = true ]; then set -x; fi # print trace if DEBUG
 
 ici_configure_ros
 
+if [ -n "$ROS_DISTRO:-" ]; then
```
ipa-mdl, line 44: Sorry, my bad!

```suggestion
if [ -n "${ROS_DISTRO:-}" ]; then
```

industrial_ci/src/tests/ros_prerelease.sh
```diff
@@ -111,9 +111,11 @@ function run_ros_prerelease() {
     ici_run "prepare_prerelease_workspaces" prepare_prerelease_workspaces "$WORKSPACE" "$reponame" "$(basename "$TARGET_REPO_PATH")"
     ici_run 'generate_prerelease_script' sudo -EH -u ci generate_prerelease_script.py "${ROSDISTRO_INDEX_URL}" "$PRERELEASE_DISTRO" default "$OS_NAME" "$OS_CODE_NAME" "${OS_ARCH:-amd64}" --build-tool "$BUILDER" --level "$downstream_depth" --output-dir "$WORKSPACE" --custom-repo "$reponame::::"
 
+    extend=${EXTEND_FROM:?}
+
```
ipa-mdl, line 115:
```suggestion
```

industrial_ci/src/tests/ros_prerelease.sh
```diff
@@ -111,9 +111,11 @@ function run_ros_prerelease() {
...
     local setup_sh=
-    if [ -f "/opt/ros/$ROS_DISTRO/setup.sh" ]; then
-        setup_sh=". /opt/ros/$ROS_DISTRO/setup.sh && "
+    if [ -f "$extend/setup.sh" ]; then
```
ipa-mdl, line 117:
```suggestion
    if [ -f "${EXTEND_FROM:?}/setup.sh" ]; then
```

industrial_ci/src/tests/ros_prerelease.sh
```diff
@@ -111,9 +111,11 @@ function run_ros_prerelease() {
...
+        setup_sh=". $extend/setup.sh && "
```
ipa-mdl, line 118:
```suggestion
        setup_sh=". $EXTEND_FROM/setup.sh && "
```

industrial_ci/src/tests/ros_prerelease.sh
ipa-mdl, line 115:
```suggestion
```

doc/index.rst
```diff
@@ -185,6 +185,7 @@ Note that some of these currently tied only to a single option, but we still lea
 * **DOCKER_RUN_OPTS** (default: not set): Used to specify additional run options for Docker.
 * **DOWNSTREAM_CMAKE_ARGS** (default: not set): Addtional CMake arguments for downstream `workspace <#workspace-management>`__.
 * **DOWNSTREAM_WORKSPACE** (default: not set): Definition of downstream `workspace <#workspace-management>`__.
+* **EXTEND_FROM** (default: ``/opt/ros/$ROS_DISTRO``): Path to a workspace to be used as an underlay of the workspaces being set up be ICI, e.g. a workspace provided by a custom docker image
```
ipa-mdl, line 188: We could just call it `UNDERLAY`.
"extend" was meant as "extend the chain", but it might not make much sense to users.

ipa-mdl:
LGTM now.
I just have second thoughts about the name..
Please test it with one of your repositories.

ipa-mdl (approved):

fmessmer:
yes, works for me...
do you still want me to rename `EXTEND_FROM` to `UNDERLAY`?

ipa-mdl:
> do you still want me to rename EXTEND_FROM to UNDERLAY?

I am in favor of  `UNDERLAY`, because it is a little bit shorter and it might be easier to understand.

fmessmer:
renaming done

ipa-mdl:
@fmessmer: I have squashed the commits and rebased to master. Please test it once again before the merge.

fmessmer:
all good, thanks for iterating!
I guess you'll take over the rebase-squash-merge from here to sync it with the other PRs...

Issue #618, opened 2021-02-10
ici action fails with jq command not found


fmessmer:
for the following config:
```
name: CI

on:
  push:
  pull_request:
#  schedule:
#    - cron: "0 0 * * 0" # every Sunday at midnight

jobs:
  industrial_ci:
    if: github.event.pull_request.draft == false  # only run on non-draft PRs
    runs-on: self-hosted
    timeout-minutes: 60

    env:
      ADDITIONAL_DEBS: 'apt-utils curl dialog'
      CATKIN_LINT: pedantic
      CATKIN_LINT_ARGS: '--ignore description_boilerplate'
      CCACHE_DIR: /github/home/.ccache  # Directory for ccache (and how we enable ccache in industrial_ci)
      CMAKE_ARGS: -DCMAKE_BUILD_TYPE=Release
      PARALLEL_TESTS: 1
      PYLINT_ARGS: '--output-format=parseable --errors-only'
      PYLINT_CHECK: true
      ROS_REPO: main
      SECRET_AVAILABLE: ${{secrets.GITHUBPAT}}
      UNDERLAY: '/root/git/care-o-bot/devel'  # workspace on docker image

    strategy:
      fail-fast: false
      matrix:
        include:
          - { ROS_DISTRO: kinetic,
              DOCKER_IMAGE: 'ghcr.io/mojin-robotics/mojin-ros:kinetic-full',
              TARGET_WORKSPACE: 'mojin_atf_kevin'
            }

    steps:
      - uses: actions/checkout@v2
        with:
          lfs: true
          ssh-known-hosts: ''

      # This step will fetch/store the directory used by ccache before/after the ci run
      - uses: actions/cache@v2
        with:
          path: ${{ env.CCACHE_DIR }}
          key: ccache-${{ matrix.ROS_DISTRO }}

      - uses: docker/login-action@v1
        if: ${{ env.SECRET_AVAILABLE }} # prevents failure notifications if secret GITHUBPAT is not available
        with:
          registry: ghcr.io
          username: mojin-robotics-ci
          password: ${{ secrets.GITHUBPAT }}

      - uses: 'ros-industrial/industrial_ci@master'
        if: ${{ env.SECRET_AVAILABLE }} # prevents failure notifications if secret GITHUBPAT is not available
        with:
          config: ${{toJSON(matrix)}}
```

I get the following failure
```
Run ros-industrial/industrial_ci@master
  with:
    config: {
[...]

PR #619, opened 2021-02-10
Make GitHub action fail gracefully if jq is not installed

ipa-mdl:
Fix for #618


ipa-mdl:
@fmessmer: Please review.

fmessmer:
> @fmessmer: Please review.

can you explain why ici should not install `jq` if it makes use of it?
otherwise, I'll test this warning when I focus on self-hosted runners again...for now I installed it manually to make it work temporarily....

ipa-mdl:
> can you explain why ici should not install jq if it makes use of it?

I don't know how.. this is running in the host environment, isn't it?

PR #626, opened 2021-02-17
Always copy TARGET_REPO_PATH if it contains BASEDIR

ipa-mdl:
As reported in https://github.com/ros-industrial/industrial_ci/pull/595#issuecomment-779171664
This paths creates the temporary folder, if BASEDIR is a sub-directory of TARGET_REPO_PATH or if they are equal.
Gitlab CI passed: https://gitlab.com/ipa-mdl/industrial_ci/-/pipelines/257815706

@stertingen: please have a look

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -82,7 +82,7 @@ test_junit_docker:
 
 test_junit_shell:
   image: ros:noetic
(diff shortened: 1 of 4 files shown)


stertingen:
Fixed it, thanks!

PR #628, opened 2021-02-24
Default BUILDER to catkin_tools for noetic as well

ipa-mdl:


ipa-mdl:
I would rather like to find a work-around for the `ROS_PACKAGE_PATH` chaining with colcon.

ipa-mdl:
Not needed anymore (#630)

PR #633, opened 2021-02-25
Fix CMP0048 warning

stertingen:


stertingen:
See also: https://wiki.ros.org/noetic/Migration#Increase_required_CMake_version_to_avoid_author_warning

stertingen:
For some reson, I cannot trigger the warning in the CI. The warning appers when I run catkin on the industrial_ci package in a clean ROS workspace.

ipa-mdl:
Thank you much for this fix! 

I wonder, if it would make more sense to  just go with the first commit and test indigo against a mock package with old CMake version.

I doubt that any user still needs to run `industrial_ci` ON a Ubuntu 14.04 host.
And even if, it is possible to call the scripts directly (that's how I am using them..)
We can migrate the mock packages as well (except for the one, which is uses for indigo as well -> `cmake_policy(SET CMP0048 NEW)`). They are not built by default, so no warning should be raised anyway.

If you agree, please remove  the last commit and apply "Fix CMP0048 without breaking indigo" to the test package only.
I can fix the tests afterwards.
I want to migrate most of them to `melodic` and `noetic` anyway.

stertingen:
Like this?

ipa-mdl:
> Like this?

Exactly!

ipa-mdl:
Gitlab: https://gitlab.com/ipa-mdl/industrial_ci/-/pipelines/261981850

ipa-mdl (approved):

PR #651, opened 2021-04-08
cleanup

rhaschke:
This is just some minor cleanup I had lying around: Everywhere else variables are directly initialized as well.


ipa-mdl (changes requested):
Why is this a cleanup? It breaks shellcheck with [SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155).

https://github.com/ros-industrial/industrial_ci/pull/598 will make it two lines.

rhaschke:
I didn't know about SC2155. Sorry for the noise.

PR #655, opened 2021-04-10
Improve clang-tidy check

### LINKED ISSUE #645: clang-tidy only packages with changes
I'd like to run clang-tidy on only packages with changes since some commit or branch?

I started down the path of creating something to do this with this package: https://github.com/tylerjw/moveit_ci_tools

That combined with the binary `ament_clang_tidy` built from source from the `ament_lint` package will do basically what I need.  One limitation though is I have to make a manual change to `ament_clang_tidy` to get it to work with clang-tidy-10.

All this is to say, is there a way you can think of that I could filter the packages that the clang-tidy test in industrial_ci runs on to only ones with changes since some point in the git history (to cut down on the time needed to run it in each pr)?

rhaschke:
This is a further improvement over #649, including:
- Filtering for changed files based on value of `CLANG_TIDY_BASE_REF` (further simplified wrt. #649):
  - not set: check all source files
  - set to _base ref_: filter for changes between _base ref_ and `HEAD`
  - set to `HEAD`: never check (because there are no changes)
  - set to `auto`: automatically set `CLANG_TIDY_BASE_REF` to _base ref_ of a pull request, else to a push commit's ref (`HEAD`).
- Avoid double application of `clang-tidy` with colcon builder
  `colcon` creates a cumulative `compile_commands.json` in the `build` folder, which was triggering an extra run of clang-tidy.
- Install `git` for use in `CLANG_TIDY_BASE_REF` filter
  Not all docker images might have that on board...
- Support `CLANG_TIDY_BASE_REF=auto` retrieving the correct base ref automatically (in Travis + GHA)
- Improve output of clang-tidy
- Actually split work across multiple jobs
  This was obviously intended, but only a single job was run (`xargs` option `-n` was missing)

Example runs can be found here:
- [with everything filtered](https://github.com/rhaschke/rviz/runs/2313129434?check_suite_focus=true)
- [some failures](https://github.com/rhaschke/rviz/runs/2315179748?check_suite_focus=true)
- [successfull pass on all files](https://github.com/ubi-agni/agni_tf_tools/runs/2314516252?check_suite_focus=true)

Fixes #645


tylerjw:
>     * Install `git` for use in `CLANG_TIDY_BASE_REF` filter

Awesome, thank you for doing this.  I had just noticed this issue this morning and was going to update the other PR for this.  Thank you for fixing all the other issues too.

tylerjw:
I looked at the test that is failing and don't understand it.

rhaschke:
Configuring `DEBUG_BASH=true` revealed the problem of the faililng CI job. Should be fixed now.
@tylerjw, note that I changed the semantics of `CLANG_TIDY_BASE_REF` as indicated in the [main comment](https://github.com/ros-industrial/industrial_ci/pull/655#issue-612914423).

travis.sh
```diff
@@ -32,6 +32,9 @@ if [ "$ABICHECK_MERGE" = "auto" ]; then
   export ABICHECK_MERGE=false
   [ "$TRAVIS_PULL_REQUEST" = "false" ] || ABICHECK_MERGE=true
 fi
+if [ "$CLANG_TIDY_BASE_REF" = "auto" ]; then
```
ipa-mdl, line 35: This should be added to Github action as well, if possible

industrial_ci/src/tests/source_tests.sh
```diff
@@ -79,11 +97,13 @@ function run_clang_tidy_check {
     local clang_tidy_args=()
     ici_parse_env_array clang_tidy_args CLANG_TIDY_ARGS
 
-    ici_run "install_clang_tidy" ici_install_pkgs_for_command clang-tidy clang-tidy "$(apt-cache depends --recurse --important clang  | grep "^libclang-common-.*")"
+    ici_run "install_clang_tidy" ici_install_pkgs_for_command clang-tidy \
+        clang-tidy "$(apt-cache depends --recurse --important clang  | grep "^libclang-common-.*")" \
+        "$(test -n "$CLANG_TIDY_BASE_REF" && echo git)"
```
ipa-mdl, line 102: To install `git` please use `ici_setup_git_client`.

travis.sh
ipa-mdl: > Both behaviours can easily be created by automatically setting the variable in GHA or shell, e.g. for GHA:

Fair enough.

industrial_ci/src/tests/source_tests.sh
tylerjw: One thing I just noticed.  Should we add an option to specify the version of clang-tidy?  There is an option to specify the version of clang-format.  I'm not sure if it matters as much with clang-tidy as I'm not sure if there are similar changes.  

Also, if we allow specifying the version how do we deal with someone wanting to specify a new version of clang-tidy on an old version of ROS where that version isn't available for install?  That being said, maybe ignore this comment for now.

ipa-mdl: > Should we add an option to specify the version of clang-tidy? There is an option to specify the version of clang-format.

This is up to you. If you need it, just add it.

> Also, if we allow specifying the version how do we deal with someone wanting to specify a new version of clang-tidy on an old version of ROS where that version isn't available for install?

This is just not supported. Similar to clang-format.

tylerjw: For our use-case the default versions of clang-tidy that are installed work the same for our config.  So this is a non-issue I think.  Maybe this is much less of an issue than with different versions of clang-format.

tylerjw: > For our use-case the default versions of clang-tidy that are installed work the same for our config. So this is a non-issue I think. Maybe this is much less of an issue than with different versions of clang-format.

Sorry to derail this comment on this line, I should have just used a normal comment for this.

industrial_ci/src/tests/source_tests.sh
```diff
@@ -33,14 +33,31 @@ function run_clang_tidy {
...
+    # create an array of all files listed in $db filtered by the source tree
     mapfile -t files < <(grep -oP "(?<=\"file\": \")($regex)(?=\")" "$db")
+    local num_all_files; num_all_files="${#files[@]}"
+    if [ -n "$CLANG_TIDY_BASE_REF" ] ; then
+        echo "Filtering for files that actually changed since $CLANG_TIDY_BASE_REF"
+        # Need to run git in actual source dir:  $files[@] refer to source dir and $PWD is read-only
+        local src_dir
+        src_dir=$(grep -oP "(?<=CMAKE_HOME_DIRECTORY:INTERNAL=).*" "$build/CMakeCache.txt")
+        pushd "$src_dir" > /dev/null || true
+        if git fetch -q origin "$CLANG_TIDY_BASE_REF" ; then  # git might fail, e.g. operating on catkin_tools_prebuild
```
ipa-mdl, line 49: stderr should be written `/dev/null` to get rid of https://github.com/rhaschke/rviz/runs/2313129434?check_suite_focus=true#step:7:209:

```
 fatal: not a git repository (or any parent up to mount point /home/runner)
 Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

rhaschke: Done. Corresponding example build is retriggered: https://github.com/rhaschke/rviz/runs/2327098690

industrial_ci/src/tests/source_tests.sh
rhaschke: Sure. Done.

rhaschke:
I also augmented the docs with some examples how the new variable `CLANG_TIDY_BASE_REF` could be configured.

industrial_ci/src/tests/source_tests.sh
```diff
@@ -80,10 +98,11 @@ function run_clang_tidy_check {
     ici_parse_env_array clang_tidy_args CLANG_TIDY_ARGS
 
     ici_run "install_clang_tidy" ici_install_pkgs_for_command clang-tidy clang-tidy "$(apt-cache depends --recurse --important clang  | grep "^libclang-common-.*")"
+    test -n "$CLANG_TIDY_BASE_REF" && ici_setup_git_client
```
ipa-mdl, line 101: This will fail, if `CLANG_TIDY_BASE_REF` is empty (`set -e`...)
Either `test -z "$CLANG_TIDY_BASE_REF" || ici_setup_git_client` or
```
if [ -n "$CLANG_TIDY_BASE_REF" ]; then
  ici_setup_git_client
fi
```
I prefer the latter, because it is more expressive.

doc/index.rst
```diff
@@ -175,6 +175,13 @@ Note that some of these currently tied only to a single option, but we still lea
 * **CLANG_FORMAT_CHECK** (default: not set. Value range: [``<format-style>``|``file``]): If set, run the `clang-format <https://clang.llvm.org/docs/ClangFormat.html>`__ check. Set the argument to ``file`` if the style configuration should be loaded from a ``.clang-format`` file, located in one of the parent directories of the source file.
 * **CLANG_FORMAT_VERSION** (default: not set): Version of clang-format to install and use (relates to both the apt package name as well as the executable), e.g., ``CLANG_FORMAT_VERSION=3.8``.
 * **CLANG_TIDY** (default: not set. Value range: [``true``|``pedantic``]): If set, run `clang.tidy <https://clang.llvm.org/extra/clang-tidy/>`__ to check the code in all packages and fail in case of errors. If ``pedantic``, warnings will be treated as errors as well.
+* **CLANG_TIDY_BASE_REF** (default: not set.): If set, clang-tidy tests will be performed on files only that changed since the given ref. If not set, clang-tidy checks are performed on all files.
```
ipa-mdl, line 178: nitpick: please put it if after `CLANG_TIDY_ARGS` (here and in the next two chunks)

industrial_ci/src/tests/source_tests.sh
rhaschke: I see. Fixed.

doc/index.rst
rhaschke: I'm not quite sure what you meant by "the next two chunks". But I guess you want to have the variables sorted alphabetically.
Done.

rhaschke:
@ipa-mdl, is there anything missing for this to be merged?

ipa-mdl:
> is there anything missing for this to be merged?

Just my review..

industrial_ci/src/env.sh
```diff
@@ -47,6 +47,7 @@ export CLANG_FORMAT_CHECK=${CLANG_FORMAT_CHECK:-}
 export CLANG_FORMAT_VERSION=${CLANG_FORMAT_VERSION:-}
 
 export CLANG_TIDY=${CLANG_TIDY:-false}
+export CLANG_TIDY_BASE_REF=${CLANG_TIDY_BASE_REF:-}
 export CLANG_TIDY_ARGS=${CLANG_TIDY_ARGS:-}
```
ipa-mdl, line 51:
```suggestion
export CLANG_TIDY_ARGS=${CLANG_TIDY_ARGS:-}
export CLANG_TIDY_BASE_REF=${CLANG_TIDY_BASE_REF:-}
```

industrial_ci/src/isolation/docker.env
```diff
@@ -13,6 +13,7 @@ CFLAGS
 CLANG_FORMAT_CHECK
 CLANG_FORMAT_VERSION
 CLANG_TIDY
+CLANG_TIDY_BASE_REF
 CLANG_TIDY_ARGS
```
ipa-mdl, line 17:
```suggestion
CLANG_TIDY_ARGS
CLANG_TIDY_BASE_REF
```

industrial_ci/src/tests/source_tests.sh
```diff
@@ -33,14 +33,31 @@ function run_clang_tidy {
     local db=$4
     shift 4
 
+    local build; build="$(dirname "$db")"
+    local name; name="$(basename "$build")"
+    ici_time_start "clang_tidy_check_$name"
+
+    # create an array of all files listed in $db filtered by the source tree
     mapfile -t files < <(grep -oP "(?<=\"file\": \")($regex)(?=\")" "$db")
+    local num_all_files; num_all_files="${#files[@]}"
```
ipa-mdl, line 42:
```suggestion
    local num_all_files="${#files[@]}"
```

ipa-mdl (changes requested):
Just some minor issues, the rest looks good.

rhaschke:
Thanks for your suggestions. I applied them as fixups, such that we have a clean history again.

ipa-mdl (approved):

PR #659, opened 2021-04-12
Don't run (failing!) cron-job workflow on a fork

rhaschke:
Why do you need a scheduled workflows at all? Each push is validated anyway. Scheduled workflows only make sense if you depend on external resources, so maybe `external.yml` but not main?
The core issue is of course is that [we cannot opt-in or opt-out to GHA on forks](https://github.community/t/stop-github-actions-running-on-a-fork/17965/4).

### DIFF
--- a/.github/workflows/legacy.yml
+++ b/.github/workflows/legacy.yml
@@ -7,6 +7,7 @@ on:
 # https://stackoverflow.com/questions/58798886/github-actions-schedule-operation-on-branch
 jobs:
   shellcheck:
+    if: github.repository == ‘ros-industrial/industrial_ci’ # don't run on forks
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
@@ -16,6 +17,7 @@ jobs:
         with:
           args: /bin/sh -c "shellcheck -x *.sh industrial_ci/scripts/*_ci industrial_ci/src/*.sh industrial_ci/src/*/*.sh"
   industrial_ci:
+    if: github.repository == ‘ros-industrial/industrial_ci’ # don't run on forks
     strategy:
       matrix:
         env:


ipa-mdl:
> Why do you need a scheduled workflows at all? Each push is validated anyway. Scheduled workflows only make sense if you depend on external resources, so maybe external.yml but not main?

On way to many occasions something got changed in the upstream tools.
For example at some point EOL distros get removed and copied to the snapshots repo.
Or rosdep got updated to not download EOL distros anymore.
Or shellchecks gets updated and shows new errors.

I could wait for the users to complain or I just fix the failing cron job ;)

> Don't run (failing!) cron-job workflow

Normally, they should not fail.

> The core issue is of course is that we cannot opt-in or opt-out to GHA on forks.

You can opt-out, but this involves disabling every workflow manually.

ipa-mdl (approved):
Nice solution!
I think this should be added to all cronjobs, but with the option to run it manually.
I will update it later.

rhaschke:
> You can opt-out, but this involves disabling every workflow manually.

I didn't find a (Web)GUI option to do so. Or are you talking about modifying the yaml configs as I did?

gavanderhoorn:
> > You can opt-out, but this involves disabling every workflow manually.
> 
> I didn't find a (Web)GUI option to do so. Or are you talking about modifying the yaml configs as I did?

There is UI to (en|dis)able individual workflows. See [Disabling and enabling a workflow](https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow).

ipa-mdl:
> I think this should be added to all cronjobs, but with the option to run it manually.

Ah, this would disable the builds on push/pr as well..

ipa-mdl:
For now I merged this as-is, because it disable at least the legacy cron-job, which will fail, if it is out of sync.
Not sure what do about the others.

One option is to specify `if: github.repository == ‘ros-industrial/industrial_ci || github.event_name != 'schedule'`, but this has to be done for each job :/

rhaschke:
Yeah, that's not nice at all. Better they fix it upstream in github actions. Could you endorse the corresponding request [here](https://github.community/t/stop-github-actions-running-on-a-fork/17965/4)?

PR #664, opened 2021-04-21
Skip BASEDIR, CCACHE_DIR and industrial_ci in ici_import_directory

### LINKED ISSUE #661: Large copy when BASEDIR inside repo dir on GitLab
As a result of these lines:

https://github.com/ros-industrial/industrial_ci/blob/bfc68c8a3d5adf12ea4074c2149582398b5b6b54/industrial_ci/src/ci_main.sh#L44-L48

This `.gitlab-ci.yml` results in a huge copy when trying this:

```
stages:
  - test

test:
  except:
    - schedules
  stage: test
  variables:
    ROS_DISTRO: melodic
    ISOLATION: shell
    # repos file containing only exact hash of repos (this almost never changes)
    UPSTREAM_WORKSPACE: 'upstream-fixed.repos'
    BASEDIR: ${CI_PROJECT_DIR}/.workspaces
    CCACHE_DIR: ${CI_PROJECT_DIR}/.ccache
  cache:
    when: always
    key:
      files:
        - upstream-fixed.repos
    paths:
      - .ccache/
      - .workspaces/
  script:
    - git clone --quiet --depth 1 https://github.com/ros-industrial/industrial_ci .industrial_ci -b master
    - .industrial_ci/gitlab.sh
```

This seemed clever until I realized that when the BASEDIR is defined, the target repo directory is copied into it's own directory.  This results in a huge (and time consuming) copy in this use case.  Maybe the copy could somehow skip the BASEDIR if it is defined within the target repo directory.

This is because of the (imo lame and broken) policy of GitLab only caching directories inside the repo directory.

The workaround I tried was setting the base_dir out of the repo, but still caching in the repo and moving it out before starting industrial ci.  The issue with that is that it updates the timestamps so the build upstream workspace step still happens as if the cache wasn't there.

Do you know a workaround for this or is there someway I could update this step to make it not copy CCACHE or BASEDIR if it is within the target repo.

ipa-mdl:
reverts #595 and #626

fixes #661

Gitlab: https://gitlab.com/ipa-mdl/industrial_ci/-/pipelines/290145549

### DIFF
--- a/industrial_ci/src/util.sh
+++ b/industrial_ci/src/util.sh
@@ -405,5 +405,4 @@ function ici_make_temp_dir {
 }
 
 # shellcheck disable=SC1090
-
 ici_source_component _FOLDING_TYPE folding
--- a/industrial_ci/src/util.sh
+++ b/industrial_ci/src/util.sh
@@ -404,13 +404,6 @@ function ici_make_temp_dir {
   ici_cleanup_later "$ici_make_temp_dir_res"
 }
 
-function ici_relocate_target_path {
-  local tmp_src
-  ici_make_temp_dir tmp_src
-  cp -a "$TARGET_REPO_PATH" "$tmp_src/"
-  export TARGET_REPO_PATH
-  TARGET_REPO_PATH="$tmp_src/$(basename "$TARGET_REPO_PATH")"
-}
-
 # shellcheck disable=SC1090
+
 ici_source_component _FOLDING_TYPE folding
--- a/industrial_ci/src/workspace.sh
+++ b/industrial_ci/src/workspace.sh
@@ -197,17 +197,11 @@ function  ici_import_directory {
     mkdir "$target"
     local args=()
     for p in "$BASEDIR" "$CCACHE_DIR" "$(readlink -m "$ICI_SRC_PATH/../..")"; do
-        if [ -n "$p" ]; then
-            args+=("(" "!" -path "$p" ")")
+        if [[ $p/ ==  $dir/?* ]]; then
+            args+=("--exclude=.${p#$dir}")
         fi
     done
-    while read -r f; do
-        local ft
-        ft=$target/$(basename "$f")
-        echo "$f -> $ft"
-        cp -a "$f" "$ft"
-    done < <(find "$dir" -maxdepth 1 -mindepth 1 "${args[@]}")
-
+    tar c "${args[@]}" -C "$dir" . | tar x -C "$target"
 }
 
 function ici_prepare_sourcespace {
--- a/industrial_ci/src/workspace.sh
+++ b/industrial_ci/src/workspace.sh
@@ -190,8 +190,24 @@ function ici_import_url {
 function  ici_import_directory {
     local sourcespace=$1; shift
     local dir=$1; shift
-    rm -rf "$sourcespace:?/$(basename "$dir")"
-    cp -a "$dir" "$sourcespace"
+    local target
+    target=${sourcespace:?}/$(basename "$dir")
+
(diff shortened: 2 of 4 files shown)


industrial_ci/src/util.sh
```diff
@@ -404,13 +404,6 @@ function ici_make_temp_dir {
...
-function ici_relocate_target_path {
-  local tmp_src
-  ici_make_temp_dir tmp_src
-  cp -a "$TARGET_REPO_PATH" "$tmp_src/"
-  export TARGET_REPO_PATH
-  TARGET_REPO_PATH="$tmp_src/$(basename "$TARGET_REPO_PATH")"
-}
-
 # shellcheck disable=SC1090
+
```
ipa-mdl, line 408:
```suggestion
```

ipa-mdl:
@stertingen: Please test this with your system as well.

industrial_ci/src/workspace.sh
```diff
@@ -190,8 +190,24 @@ function ici_import_url {
...
+            args+=("(" "!" -path "$p" ")")
+        fi
+    done
+    while read -r f; do
+        local ft
+        ft=$target/$(basename "$f")
+        echo "$f -> $ft"
+        cp -a "$f" "$ft"
+    done < <(find "$dir" -maxdepth 1 -mindepth 1 "${args[@]}")
+
```
ipa-mdl, line 210:
```suggestion
```

industrial_ci/src/workspace.sh
```diff
@@ -190,8 +190,23 @@ function ici_import_url {
...
+        if [ -n "$p" ]; then
+            args+=("(" "!" -path "$p" ")")
+        fi
+    done
+    while read -r f; do
+        local ft
+        ft=$target/$(basename "$f")
+        echo "$f -> $ft"
+        cp -a "$f" "$ft"
+    done < <(find "$dir" -maxdepth 1 -mindepth 1 "${args[@]}")
```
tylerjw, line 209: I'll test this but I suspect this will not work if the user places the BASEDIR or CCACHE_DIR at any depth except right below the repo directory.  That is why I switched to the `tar` command over `find`.  For example if the user does something like this:

```
CCACHE_DIR: .cache/ccache
BASEDIR: .cache/workspaces
```

In my usecase I only wanted to cache the upstream workspace and not all the data in the basedir so placing everything in a `.cache` directory and caching that didn't work well for my use case.  However, this still revealed the issue with find and the inability to deal with directories that were deeper than just the first level.  Maybe there is a way to do this with find, I just don't know.  Find seems to have it's own language spec that is very confusing to me.

ipa-mdl: > I'll test this but I suspect this will not work if the user places the BASEDIR or CCACHE_DIR at any depth except right below the repo directory. 

Ah, yes,  I think you are correct.

> That is why I switched to the tar command over find

It should be possible with `find` as well.
I will give it a try.

ipa-mdl: > It should be possible with find as well.
> I will give it a try.

It is possible with path pruning and special handling of directories.
But tar is simpler and will perform better.
And since this is now run in the target container a suitable tar version should always ne pre-installed.

@tylerjw: Please review/try the updated version.

stertingen:
> @stertingen: Please test this with your system as well.

A quick test passed, LGTM.

industrial_ci/src/workspace.sh
tylerjw: Tested and this seems to work for me.  Thank you for following up on this!

PR #665, opened 2021-04-23
run_clang_tidy: fix changed-files filter

rhaschke:
Fixup a bug introduced in #655: non-cpp files might get clang-tidy checked:
https://github.com/ros-planning/moveit_task_constructor/runs/2423567108

If no source files are found in `compile_commands.json`, the subsequent git filtering just returns *all* changed files. 
However these shouldn't be clang-tidy checked!


ipa-mdl (approved):

rhaschke:
Thanks for quickly merging!

Issue #668, opened 2021-05-03
UNDERLAY feature causes `build_upstream_workspace` to fail to find catkin CMake module


JStech:
We attempted to use the UNDERLAY feature in MoveIt Calibration. It seemed to work [initially](https://github.com/ros-planning/moveit_calibration/actions/runs/683966589), but recently every CI run was failing in the `build_upstream_workspace` step with the error `The catkin CMake module was not found,...` (for instance, [here](https://github.com/ros-planning/moveit_calibration/runs/2387071599)). Reverting the use of the UNDERLAY feature fixed this error.

The error also occurred locally, when running this command:
```
rosrun industrial_ci rerun_ci src/moveit_calibration \
  ROS_DISTRO=melodic ROS_REPO=ros \
  DOCKER_IMAGE='moveit/moveit:melodic-source' \
  UPSTREAM_WORKSPACE='upstream.rosinstall' \
  BEFORE_INSTALL_UPSTREAM_DEPENDENCIES_EMBED='set +u && source /root/ws_moveit/install/setup.bash && set -u' \
  AFTER_RUN_TARGET_TEST='./.ci.prepare_codecov' \
  TARGET_CMAKE_ARGS="-DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_FLAGS='--coverage' -DCMAKE_CXX_FLAGS='--coverage'" \
  ADDITIONAL_DEBS='lcov' \
  CCACHE_DIR="${HOME}/.ccache" \
  BASEDIR=${HOME}/.work
```

fmessmer:
I guess, I intended UNDERLAY to be used **instead** of UPSTREAM_WORKSPACE in https://github.com/ros-industrial/industrial_ci/pull/609

I haven't looked into how both could be made to work together...

gavanderhoorn:
It seems something isn't entirely correct in the environment setup:

> WARNING: Your workspace is not extending any other result space, but it is set to use a `linked` devel space layout. This requires the `catkin` CMake package in your source space in order to be built.

JStech:
Tagging @tylerjw because he's much more familiar with MoveIt's use of `industrial_ci` than I am.

ipa-mdl:
> I haven't looked into how both could be made to work together...

This should not be a problem..

> It seems something isn't entirely correct in the environment setup:

I had similar issues in some tests, but I don't recall why I got this error message.

ipa-mdl:
[...]

PR #669, opened 2021-05-04
Test that UNDERLAY contains a setup.bash

ipa-mdl:
This tests `UNDERLAY` to prevent from issues like #668.
The test was moved after the init script to allow for underlays created in a script.

Setting `UNDERLAY=/opt/ros/$ROS_DISTRO` will only work if a ROS-based Docker images is used.

@fmessmer, @tylerjw, @JStech : Please test

### DIFF
--- a/industrial_ci/src/run.sh
+++ b/industrial_ci/src/run.sh
@@ -62,7 +62,7 @@ ici_run "init" ici_init_apt
 
 if [ -n "${UNDERLAY:-}" ]; then
     if [ ! -f "$UNDERLAY/setup.bash" ]; then
-        ici_error "UNDERLAY '$UNDERLAY'  does not contain a setup.bash"
+        ici_error "UNDERLAY '$UNDERLAY' does not contain a setup.bash"
     fi
 else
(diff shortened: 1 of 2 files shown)


tylerjw (approved):
Thank you for this, I tested this in a github action and see the failure with the message as expected.

industrial_ci/src/run.sh
```diff
@@ -64,6 +60,16 @@ ici_source_component TEST tests
 
 ici_run "init" ici_init_apt
 
+if [ -n "${UNDERLAY:-}" ]; then
+    if [ ! -f "$UNDERLAY/setup.bash" ]; then
+        ici_error "UNDERLAY '$UNDERLAY'  does not contain a setup.bash"
```
ipa-mdl, line 65:
```suggestion
        ici_error "UNDERLAY '$UNDERLAY' does not contain a setup.bash"
```

JStech:
Thanks for the fix--it now gives a clear error message when I run locally with an incorrect `UNDERLAY`.

PR #675, opened 2021-05-12
Mark ROS2 dashing and eloquent as EOL

ipa-mdl:


ipa-mdl:
Actually, `dashing` is not yet EOL, only `eloquent`.

PR #681, opened 2021-05-29
Remove hydro test from legacy cron job

ipa-mdl:
The Ubuntu archive repository does not host the packages for precise anymore.


ipa-mdl:
Test are failing because of #683

PR #682, opened 2021-05-29
Remove hydro test job and add note about APT issue

ipa-mdl:


ipa-mdl:
Test are failing because of #683

Issue #624, opened 2021-02-15
rosenv does not correctly consider CMAKE_PREFIX_PATH in AFTER_SCRIPT


fmessmer:
I made the following observation:
- using Github Actions (does NOT happen on travis)
- for `noetic` :point_right: default builder is `colcon` (does NOT happen on `kinetic` :point_right: default builder `catkin-tools`)
- using `UNDERLAY` option with custom docker image
- `AFTER_SCRIPT: 'rosenv bash .travis.xacro_test.sh'` 
  (see https://github.com/ipa320/cob_common/blob/kinetic_dev/.travis.yml#L12)

for some reason in GHA noetic job, the test fails because the resources for the xacro test cannot be found, although they should be findable in the `$CMAKE_PREFIX_PATH`

```
   Starting function 'after_script'
  Testing /home/runner/work/mojin_common/mojin_common/mojin_description/urdf/base_theta/base_theta.urdf.xacro
  Testing /home/runner/work/mojin_common/mojin_common/mojin_description/urdf/mrl_base/base.gazebo.xacro
  Testing /home/runner/work/mojin_common/mojin_common/mojin_description/urdf/mrl_base/base.urdf.xacro
  resource not found: cob_description
  ROS path [0]=/opt/ros/noetic/share/ros
  ROS path [1]=/root/target_ws/install/mojin_common/share
  ROS path [2]=/root/target_ws/install/mojin_utils/share
  ROS path [3]=/root/target_ws/install/mojin_msgs/share
  ROS path [4]=/root/target_ws/install/mojin_description/share
  ROS path [5]=/opt/ros/noetic/share
```

the missing resource `cob_description` is part of the `robot_ws` - see `CMAKE_PREFIX_PATH` below

I'm not sure whether my `AFTER_SCRIPT` is not colcon-ready or whether `rosenv` is doing something incorrect

EDIT:
here is the output of `AFTER_SCRIPT: 'rosenv && echo "$CMAKE_PREFIX_PATH" && echo "$UNDERLAY"'`
- for kinetic:
   ```
  Starting function 'after_script'
  /root/target_ws/install:/root/git/app_ws/install:/root/git/rosjava_ws/install:/root/git/mojin_ws/install:/root/git/nav_ws/install:/root/git/robot_ws/install:/opt/ros/kinetic:/root/git/care-o-bot/devel
  /root/git/care-o-bot/devel
[...]

PR #694, opened 2021-06-08
Make catkin_tools the default builder for ROS1 noetic

### LINKED ISSUE #629: catkin_tools for ROS noetic
> > EDIT: As workaround, we could set catkin_tools as default for noetic. Works fine for me.
> 
> @stertingen: I tried it with a dummy package and it did not work for me.

_Originally posted by @ipa-mdl in https://github.com/ros-industrial/industrial_ci/issues/475#issuecomment-785158785_

### LINKED ISSUE #666: colcon test-result missing in pre-release job on Noetic
I get this log at the bottom of a pre-release job on noetic for moveit:

```
2021-04-30T17:05:46.0978485Z Summary: 29 packages finished [21min 44s]
2021-04-30T17:05:46.0979913Z   5 packages had stderr output: moveit_commander moveit_ros_move_group moveit_ros_planning moveit_ros_planning_interface pilz_industrial_motion_planner
2021-04-30T17:05:46.0981585Z   5 packages had test failures: moveit_commander moveit_ros_move_group moveit_ros_planning moveit_ros_planning_interface pilz_industrial_motion_planner
2021-04-30T17:05:46.1465236Z # END SUBSECTION
2021-04-30T17:05:46.3839544Z /tmp/tmp.LDeA5yYfRC
2021-04-30T17:05:46.3840294Z # END SECTION
2021-04-30T17:05:46.3840598Z 
2021-04-30T17:05:46.3840973Z Build step 10
2021-04-30T17:05:46.3841269Z 
2021-04-30T17:05:46.3841652Z Build step 11
2021-04-30T17:05:46.3842245Z # BEGIN SECTION: Create collated test stats dir
2021-04-30T17:05:46.3900397Z # END SECTION
2021-04-30T17:05:46.3901118Z 
2021-04-30T17:05:46.3901489Z 
2021-04-30T17:05:46.3902409Z Test results of underlay workspace
2021-04-30T17:05:46.3903097Z 
2021-04-30T17:05:46.7367815Z 'colcon test-result' not available. Please make sure the necessary extension is installed to output the test result summary.
2021-04-30T17:05:46.7368722Z 
2021-04-30T17:05:46.7368965Z 
2021-04-30T17:05:46.7387241Z Skipping empty overlay workspace
```

The pre-release docker image should probably have colcon installed in it?

ipa-mdl:
While I would like to use colcon for ROS1 as well, it has several issues.
I hope that this will not break too many builds.
`BUILDER=colcon` is still supported, but now as an opt-in.

closes #629 and closes #666

@wxmerkt: Please test, if this resolves your build issues.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -60,9 +60,10 @@ jobs:
 
           # Are CXXFLAGS correctly passed? These tests should fail due to -Werror (exit code is for catkin tools: 1 and for colcon: 2)
           - {ROS_DISTRO: melodic, CXXFLAGS: "-Werror", EXPECT_EXIT_CODE: 1, TARGET_WORKSPACE: 'industrial_ci/mockups/industrial_ci_testpkg'}
-          - {ROS_DISTRO: noetic, CXXFLAGS: "-Werror", EXPECT_EXIT_CODE: 2, TARGET_WORKSPACE: 'industrial_ci/mockups/industrial_ci_testpkg'}
+          - {ROS_DISTRO: noetic, CXXFLAGS: "-Werror", EXPECT_EXIT_CODE: 1, TARGET_WORKSPACE: 'industrial_ci/mockups/industrial_ci_testpkg'}
           - {ROS_DISTRO: melodic, CMAKE_ARGS: -DCMAKE_CXX_FLAGS="-Werror", EXPECT_EXIT_CODE: 1, TARGET_WORKSPACE: 'industrial_ci/mockups/industrial_ci_testpkg'}
-          - {ROS_DISTRO: noetic, CMAKE_ARGS: -DCMAKE_CXX_FLAGS="-Werror", EXPECT_EXIT_CODE: 2, TARGET_WORKSPACE: 'industrial_ci/mockups/industrial_ci_testpkg'}
+          - {ROS_DISTRO: noetic, CMAKE_ARGS: -DCMAKE_CXX_FLAGS="-Werror", EXPECT_EXIT_CODE: 1, TARGET_WORKSPACE: 'industrial_ci/mockups/industrial_ci_testpkg'}
+          - {ROS_DISTRO: noetic, BUILDER: colcon, CMAKE_ARGS: -DCMAKE_CXX_FLAGS="-Werror", EXPECT_EXIT_CODE: 2, TARGET_WORKSPACE: 'industrial_ci/mockups/industrial_ci_testpkg'}
 
           - {ROS_DISTRO: melodic}
           - {ROS_DISTRO: melodic, ROS_REPO: main}
--- a/industrial_ci/src/ros.sh
+++ b/industrial_ci/src/ros.sh
@@ -56,7 +56,6 @@ function _set_ros_defaults {
         _ros1_defaults "bionic"
         ;;
     "noetic")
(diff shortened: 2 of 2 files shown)


wxmerkt:
Thank you for the PR - I am still experimenting locally. I believe my build issues may be more related to sourcing order of the `upstream_ws` rather than the builder. In particular, I noticed that the work-around for https://github.com/ros-industrial/industrial_ci/issues/624 in `workspace.sh` does not repopulate the `PYTHONPATH`. Fixing this, we could forgo this PR.

wxmerkt:
I can confirm that with a custom hook in `AFTER_RUN_TARGET_TEST` we are able to run tests in pure CMake projects [1]. `catkin-tools` works best in terms of discovery of pure CMake packages in `upstream_ws` and for our use cases on ROS Noetic.

[1] The reason to use a custom hook is that `catkin_tools` uses `--catkin-make-args` to `run_tests` whereas `catkin_make_isolated` and `colcon` use `--make-args`. Thus, this is an upstream limitation from a different behaviour present in `catkin_tools`.

ipa-mdl:
> catkin_tools uses --catkin-make-args to run_tests

I am okay with changing it to something else..

https://github.com/ros-industrial/industrial_ci/blob/56adb8d0a9e689fabc4fceb1711af936e823a395/industrial_ci/src/builders/catkin_tools.sh#L56

ipa-mdl:
> does not repopulate the PYTHONPATH. Fixing this, we could forgo this PR.

We could run all scripts, as it it would be done by setup.bash.

wxmerkt:
> I am okay with changing it to something else..

It should probably work with `catkin build --make-args run_tests` -- however, this would be a different behaviour than running `catkin run_tests` which expands to `catkin build --catkin-make-args run_tests`. For now, my workaround is to explicitly specify the manual test commands as an `AFTER_RUN_TARGET_TESTS` script (including sourcing & calling `make test` in the corresponding build directory).

> We could run all scripts, as it it would be done by setup.bash.

This might be a good idea. Are there any potential downsides to this?

rhaschke:
I have seen issues with `colcon` on ROS Noetic as well: [`run_target_test` wasn't finding plugins](https://github.com/ros-planning/moveit/runs/3035216279). Explicitly switching to `catkin_tools` resolved the problem. 

Maybe the `colcon` issue is caused by not sourcing the just-built install-space for testing?
I had a similar problem when experimenting with `catkin_make_isolated_devel`: 
There, it was important to [adapt the `extend` space](https://github.com/ros-industrial/industrial_ci/pull/702/files#diff-5e76c8105e5f56060828806723ab264894086450dfe89cc8c367fd2a3184d043R37-R38).

ipa-mdl:
> There, it was important to adapt the extend space compared to the code of catkin_make_isolated.

I will add this to all builder!

ipa-mdl:
@wxmerkt: please test #704

wxmerkt:
Hi @ipa-mdl 
Thank you for looking into this. I tested it on the two issues I had before:

1. Not finding pure CMake packages (REP-136) in an upstream_workspace from the downstream workspace - this still fails with colcon using #704
2. Running tests in a pure CMake project (works, but also worked before with colcon but not with catkin-tools -- this is not a industrial_ci issue but catkin-tools issue)

PR #700, opened 2021-07-15
Add support for BUILDER=catkin_tools_devel

ipa-mdl:

PR #702, opened 2021-07-15
Add support for BUILDER=catkin_make_isolated_devel

rhaschke:
Same as #700, but for `catkin_make_isolated`.
Final test build is running: https://github.com/ubi-agni/moveit/runs/3081336989?check_suite_focus=true

### DIFF
--- a/industrial_ci/src/builders/catkin_make_isolated_devel.sh
+++ b/industrial_ci/src/builders/catkin_make_isolated_devel.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+# Copyright (c) 2019, Mathias Lüdtke
+# All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# shellcheck source=industrial_ci/src/builders/catkin_make_isolated.sh
+source "${ICI_SRC_PATH}/builders/catkin_make_isolated.sh"
+
+ici_warn "BUILDER=catkin_make_isolated_devel should only be used in addition to the other non-devel builders"
+
+function ici_extend_space {
+    echo "$1/devel"
+}
+
+function builder_run_build {
+    local opts=()
+    _append_job_opts opts PARALLEL_BUILDS 0
+    _run_catkin_make_isolated all "${opts[@]}" "$@"
+}


industrial_ci/src/builders/catkin_make_isolated_devel.sh
```diff
@@ -0,0 +1,39 @@
...
+    _append_job_opts opts PARALLEL_BUILDS 0
+    _run_catkin_make_isolated all "${opts[@]}" "$@"
+}
+
+function builder_run_tests {
+    local ws=$2
+    local opts=()
+    _append_job_opts opts PARALLEL_TESTS 1
+    # to run tests, we need to source the just-built devel-space
+    _run_catkin_make_isolated run_tests "$(ici_extend_space "$ws")" "$ws" "${opts[@]}"
```
ipa-mdl, line 38: The same fix might be needed in general -> catkin_make_isolated.sh

ipa-mdl:
I fixed some whitespace issues.
Does it make sense to have both, `catkin_make_isolated_devel` and `catkin_tools_devel`? `catkin_make_isolated_devel`  might be better to spot hidden issues.

industrial_ci/src/builders/catkin_make_isolated_devel.sh
rhaschke: Hm. I did test this actually. I will trigger a MoveIt build with all these variants to check :wink:

rhaschke: https://github.com/ubi-agni/moveit/actions/runs/1037050795

rhaschke:
I would say, it doesn't harm to have both options :wink:
The output of catkin_make_isolated is much more verbose. On the MoveIt side, I'm tempted to use catkin_tools_devel, which creates more compact output (which is similar to the other outputs as well).

ipa-mdl:
> I would say, it doesn't harm to have both options wink

Yes and no. They have to be maintained ;)
And they should have a purpose.
I did not mean to support the devel space in general.

I am thinking of providing just `catkin_tools_isolated_devel`. This would allow to test the devel space, but with isolation as an additional safe-guard.

ipa-mdl:
> I am thinking of providing just catkin_tools_isolated_devel. This would allow to test the devel space, but with isolation as an additional safe-guard.

#703

rhaschke (commented):
> They have to be maintained :wink:

Sure, maintenance is always an issue. But there is not so much to maintain for both of these options, is it?
They essentially inherit from their base BUILDERs.

> And they should have a purpose.
> I did not mean to support the devel space in general.

I think it would be useful to allow testing of common usage patterns. Both, `catkin_make_isolated` (used by ROS buildfarm) and `catkin_tools` with `devel` and `install` space `are` frequently used. As they provide different verbosity, I consider both of them useful.

ipa-mdl:
> As they provide different verbosity, I consider both of them useful.

Okay, lets merge this (#702) and #703 after #704

This way we have 6 options:  {catkin_make,catkin_tools}[_isolated[_devel]]

ipa-mdl:
I prepared {catkin_make,catkin_tools}[_devel] as well, but I am not yet convinced that we should offer this option.

ipa-mdl:
@rhaschke: I have rebased your commit to the latest master and squashed my online changes (whitespace, warning).

rhaschke:
Looks good to me.

PR #703, opened 2021-07-16
Add support for BUILDER=catkin_tools_isolated_devel

ipa-mdl:
Middle ground between #700  and #702

### DIFF
--- a/industrial_ci/src/builders/catkin_tools_isolated_devel.sh
+++ b/industrial_ci/src/builders/catkin_tools_isolated_devel.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+# Copyright (c) 2021, Mathias Lüdtke
+# All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


rhaschke (commented):
:-1:
`catkin build --isolate-devel` doesn't solve the issues addressed by #700 and #702.
The key point is to **not** use an `install` space at all. However, this option still installs.

ipa-mdl:
> catkin build --isolate-devel doesn't solve the issues addressed by #700 and #702.
> The key point is to not use an install space at all. However, this option still installs.

I tested it locally and it fails with the same error as #700.

rhaschke (approved):
Sorry. I missed the point that you also dropped the `--install` option.
I think the option `--isolate-devel` is optional. I wasn't even aware of it before.

ipa-mdl:
> I think the option --isolate-devel is optional. I wasn't even aware of it before.

This default is a merged devel space (similar to `catkin_make`), the isolated mode will result the same layout as `catkin_make_isolated`.
This should catch issues with implicit dependencies between packages of the same repository/workspace, but I am not 100% sure.

ipa-mdl:
`--isolate-devel`:
```
$ tree -d .base/target_ws/devel/moveit_core/
.base/target_ws/devel/moveit_core/
├── include
│   └── moveit
├── lib
│   ├── moveit_core
│   ├── pkgconfig
│   └── python3
│       └── dist-packages
│           └── moveit
└── share
    └── moveit_core
        └── cmake

rhaschke (approved):

PR #704, opened 2021-07-16
Source test workspace before running tests

ipa-mdl:
@rhaschke: Please have a look. I ported some changes from #702 as well.

### DIFF
--- a/industrial_ci/src/builders/catkin_make.sh
+++ b/industrial_ci/src/builders/catkin_make.sh
@@ -23,6 +23,12 @@ function _append_job_opts() {
         _append_job_opts_res+=("-j$jobs" "-l$jobs")
     fi
 }
+function _run_catkin_make () {
+    local target=$1; shift
+    local extend=$1; shift
+    local ws=$1; shift
+    ici_exec_in_workspace "$extend" "$ws" catkin_make "$@" --make-args "$target"
+}
 
 function builder_setup {
     ici_install_pkgs_for_command catkin_make "ros-${ROS_DISTRO}-catkin"
@@ -33,15 +39,15 @@ function builder_run_build {
     local ws=$1; shift
     local opts=()
     _append_job_opts opts PARALLEL_BUILDS 0
-    ici_exec_in_workspace "$extend" "$ws" catkin_make --make-args install "${opts[@]}" "$@"
+    _run_catkin_make install "$extend" "$ws" "${opts[@]}" "$@"
 }
 
 function builder_run_tests {
     local extend=$1; shift
     local ws=$1; shift
-    local opts=()
+    local opts=(--force-cmake)
     _append_job_opts opts PARALLEL_TESTS 1
-    ici_exec_in_workspace "$extend" "$ws" catkin_make --make-args run_tests "${opts[@]}" "$@"
+    _run_catkin_make run_tests "$(ici_extend_space "$ws")" "$ws" "${opts[@]}" "$@"
 }
 
 function builder_test_results {
--- a/industrial_ci/src/builders/catkin_make_isolated.sh
+++ b/industrial_ci/src/builders/catkin_make_isolated.sh
@@ -28,7 +28,7 @@ function _run_catkin_make_isolated () {
     local target=$1; shift
     local extend=$1; shift
     local ws=$1; shift
-    ici_exec_in_workspace "$extend" "$ws" catkin_make_isolated --build-space "$ws/build" --devel-space "$ws/devel" --install-space "$ws/install" --make-args "$target" "$@"
+    ici_exec_in_workspace "$extend" "$ws" catkin_make_isolated --build-space "$ws/build" --devel-space "$ws/devel" --install-space "$ws/install" "$@" --make-args "$target"
 }
 
 function builder_setup {
@@ -46,9 +46,9 @@ function builder_run_build {
 function builder_run_tests {
     local extend=$1; shift
     local ws=$1; shift
-    local opts=()
+    local opts=(--force-cmake)
     _append_job_opts opts PARALLEL_TESTS 1
-    _run_catkin_make_isolated run_tests "$extend" "$ws" "${opts[@]}"
+    _run_catkin_make_isolated run_tests "$(ici_extend_space "$ws")" "$ws" "${opts[@]}"
 }
 
 function builder_test_results {
--- a/industrial_ci/src/builders/catkin_tools.sh
+++ b/industrial_ci/src/builders/catkin_tools.sh
@@ -59,7 +59,7 @@ function builder_run_tests {
         opts+=(-i)
     fi
     _append_job_opts opts PARALLEL_TESTS 1
-    ici_exec_in_workspace "$extend" "$ws" catkin build --catkin-make-args run_tests -- "${opts[@]}" --no-status
+    ici_exec_in_workspace "$(ici_extend_space "$ws")" "$ws" catkin build --catkin-make-args run_tests -- "${opts[@]}" --no-status
 }
 
 function builder_test_results {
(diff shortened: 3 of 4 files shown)


rhaschke:
I triggered a test: https://github.com/ubi-agni/moveit/actions/runs/1037270646

ipa-mdl:
```
Test node [moveit_ros_planning_interface/python_move_group.py] does not exist or is not executable
```
It looks like the moveit test cannot be run from install space.

ipa-mdl:
Apparently, for catkin_make_isolated, it is necessary to source the devel space and to force cmake:
https://github.com/ros-infrastructure/ros_buildfarm/blob/833969e9f9a4e907059c4377472ae0d018f2c884/scripts/devel/build_and_test.py#L124

rhaschke (approved):
Looks good to me.

ipa-mdl:
Something went wrong in my previous rebase..
I fixed it now: pass --force-cmake, use devel space for tests

ipa-mdl:
@rhaschke: Please trigger the test again to see, if it works as expected.

rhaschke:
Retriggered: https://github.com/ubi-agni/moveit/actions/runs/1037270646

rhaschke (approved):
Didn't you want to merge this before #702 - as mentioned in https://github.com/ros-industrial/industrial_ci/pull/702#issuecomment-881413990?
This is actually a prerequisite for #702.

ipa-mdl:
> Didn't you want to merge this before #702 - as mentioned in #702 (comment)?

I factored out the other fixes into #705, so it was okay to merge #702 right away ^^

> This is actually a prerequisite for #702.

This PR fixes all catkin_make* variants, so the order of merging is not too important.

ipa-mdl:
> Retriggered: https://github.com/ubi-agni/moveit/actions/runs/1037270646

The colcon test fails now, but as far as I can tell we did not change anything for colcon recently.

rhaschke:
> The colcon test fails now, but as far as I can tell we did not change anything for colcon recently.

This failure is related to https://github.com/ros-industrial/industrial_ci/pull/694#issuecomment-881206858.

Now `catkin_tools_isolated_devel` fails: https://github.com/ubi-agni/moveit/runs/3089772539?check_suite_focus=true
I triggered another build with `catkin_tools_isolated` as well: https://github.com/ubi-agni/moveit/actions/runs/1039708208

ipa-mdl:
For `catkin_tools_*` we can source `devel` as well.
Just not sure what to do with colcon.

ipa-mdl:
It looks like `--isolate-install` does not create the top-level setup.* files.

ipa-mdl:
@rhaschke: This version now sources the target space properly. Please try again.

industrial_ci/src/builders/catkin_make.sh
```diff
@@ -33,15 +39,15 @@ function builder_run_build {
...
 }
 
 function builder_run_tests {
     local extend=$1; shift
     local ws=$1; shift
-    local opts=()
+    local opts=(--force-cmake)
     _append_job_opts opts PARALLEL_TESTS 1
-    ici_exec_in_workspace "$extend" "$ws" catkin_make --make-args run_tests "${opts[@]}" "$@"
+    _run_catkin_make run_tests "$ws/devel" "$ws" "${opts[@]}" "$@"
```
rhaschke, line 50: Was this broken already before all these changes? How did the tests find there dependencies from the `$extend` upstream only?
Out of curiosity: Why don't you use `$(ici_extend_space "$ws")` here as well - as below for `catkin_tools`?

rhaschke (approved):
This works for `catkin_tools*` + `catkin_make*`: https://github.com/ubi-agni/moveit/actions/runs/1039708208
I didn't test `colcon` though.

rhaschke:
The `colcon` build is still failing. https://github.com/ubi-agni/moveit/runs/3094222224?check_suite_focus=true
But this might be related to a `colcon` issue / #694.

industrial_ci/src/builders/catkin_make.sh
ipa-mdl: > Was this broken already before all these changes? How did the tests find there dependencies from the $extend upstream only?

Perhaps. I am not sure, if the source space was added during the tests somehow.
It might be that we do not have to source the target space at all.

> Out of curiosity: Why don't you use $(ici_extend_space "$ws") here as well - as below for catkin_tools?

That's what `ros_buildfarm` is doing (https://github.com/ros-industrial/industrial_ci/pull/704#issuecomment-881493215).
`$(ici_extend_space "$ws")` might be better, I will give it a try.

ipa-mdl:
I factored out some commits, which are not related to sourcing the test workspace.
I am still not 100% convinces that this is need in general.

There might be some edges cases, which affect moveit. Or something is wrong with the moveit dependencies.
I will try to add a better test case.

> The colcon build is still failing

There seems to be an issue with the plugin loading.
It might make sense to compare the created environment locally.
However, `colcon` is not a priority for now.

ipa-mdl:
Open issues:

* needed for colcon?
* source devel as the time?
* is it even necessary? in which cases

rhaschke:
I was definitely necessary for `catkin_tools[_isolated]_devel` and `catkin_make_isolated_devel`.

rhaschke:
I merged this PR into master and triggered a new test: https://github.com/ubi-agni/moveit/runs/3098546328?check_suite_focus=true
Most of the jobs are failing :unamused: Maybe, we have a flaky test (pilz integration test)?!

ipa-mdl:
> Most of the jobs are failing unamused

I changed catkin_make* to  `ici_extend_space` instead of `devel`.
This might not work for some reason.

>  Maybe, we have a flaky test

Not sure, if it is a flaky test, but some dependencies seems to be missing between the different (test) packages.

ipa-mdl:
> but some dependencies seems to be missing between the different (test) packages.

See https://github.com/ros-planning/moveit_resources/issues/32 as well.
https://github.com/ros-planning/moveit_resources/pull/36/commits/c1cc11fa1dfebc3b9139bc0ca0bd8f4400dbe0a2 removed the circular dependencies, but disconnected them totally. The tested packages need to declare the `test_depends` now!

rhaschke:
It's completely possible that dependencies are not yet correctly defined in MoveIt. However, the current state of https://github.com/ros-planning/moveit/pull/2772 was passing for a previous state of this PR - see https://github.com/ros-industrial/industrial_ci/pull/704#pullrequestreview-708962654! What has happened since then?
I triggered a new job: https://github.com/ubi-agni/moveit/actions/runs/1046329344

rhaschke:
The latest MoveIt CI build [succeeded for most builders](https://github.com/ubi-agni/moveit/actions/runs/1046575587). However, `catkin_make_isolated` fails with lots of: `Test node [...] does not exist or is not executable.`
As not only C++ nodes but also python nodes are not found (e.g. python_moveit_commander.py), I don't think it's a build order issue.
I [reverted the changes to `catkin_make_isolated`](https://github.com/rhaschke/industrial_ci/commit/4fd154957f7f52badeb74cdac09727cb63c32240) and triggered a new build: https://github.com/ubi-agni/moveit/runs/3110799035

ipa-mdl:
> What has happened since then?

https://github.com/ros-industrial/industrial_ci/pull/704#issuecomment-882125717

> As not only C++ nodes but also python nodes are not found (e.g. python_moveit_commander.py), I don't think it's a build order issue.

If the dependencies are missing, the (isolated) paths will be missing everywhere.

rhaschke:
You are right: We are missing lots of `test_depend` declarations in MoveIt, which are explicitly hidden to avoid circular dependencies. You already pointed that out in https://github.com/ros-industrial/industrial_ci/pull/704#issuecomment-882571682!

If these would be correctly declared, this PR is not required anymore. Thus I think, this can be closed.

ipa-mdl:
> If these would be correctly declared, this PR is not required anymore.

You don't have to declare them transitively, just in the testted package. Or even better: Create a dedicated test package.

> Thus I think, this can be closed.

Okay, I will close it for now.

PR #707, opened 2021-07-16
Support non-isolated devel builds

ipa-mdl:


ipa-mdl:
@rhaschke: This completes the set of builders. It would be great, if you could test it as well.

industrial_ci/src/builders/catkin_tools_isolated_devel.sh
```diff
@@ -27,5 +27,5 @@ function ici_extend_space {
 function _catkin_config {
     local extend=$1; shift
     local ws=$1; shift
-    ici_exec_in_workspace "$extend" "$ws" catkin config --isolate-devel
+    ici_exec_in_workspace "$extend" "$ws" catkin config --install --link-devel
```
rhaschke, line 30: Why do you re-introduce `--install` here and remove `--isolate-devel`?
The builder name suggests an _isolated_ devel-space _only_.

industrial_ci/src/builders/catkin_tools_devel.sh
```diff
@@ -0,0 +1,25 @@
...
+# limitations under the License.
+
+# shellcheck source=industrial_ci/src/builders/catkin_tools.sh
+source "${ICI_SRC_PATH}/builders/catkin_tools.sh"
+
+ici_warn "BUILDER=catkin_tools_devel should only be used in addition to the other non-devel builders"
+
+function ici_extend_space {
+    echo "$1/devel"
+}
```
rhaschke, line 25: Shouldn't you drop the `--install` flag here as well?

rhaschke (commented):
Thanks for adding these. If merged, we can use a non-isolated build for MoveIt (which currently fails on isolated builds).

industrial_ci/src/builders/catkin_tools_isolated_devel.sh
ipa-mdl: The idea is to test the install tags anyway, even if they will not used in the chained workspaces.
`--isolate-devel` and `--install` together render the `devel` [unusable](https://github.com/catkin/catkin_tools/blob/f5a4a0ade82eb02891727b390c59cd1b6eef6f85/catkin_tools/verbs/catkin_build/build.py#L585).

industrial_ci/src/builders/catkin_tools_devel.sh
ipa-mdl: Again, install should be tested.

rhaschke: I agree here if you want to test `--install`, but not use it.

industrial_ci/src/builders/catkin_tools_isolated_devel.sh
rhaschke: I don't agree here. I think we definitely want `--isolate-devel`. If that's incompatible to `--install` (I don't yet understand why), we should drop `--install` (It's not used anyway as you claimed).

ipa-mdl: > I don't yet understand why

Looks like this is an open issue (https://github.com/catkin/catkin_tools/issues/674#issuecomment-879008090)
I removed it for now.

ipa-mdl: This feature (?) was implemented in https://github.com/catkin/catkin_tools/pull/24, so apparently sourcing the isolated devel would source the isolated install spaces anyway (except for the leaf packages, I guess).

PR #713, opened 2021-07-18
Fix clang-tidy checks

rhaschke:
`clang-tidy` checks ignored existing `.clang-tidy` config files, because clang-tidy was run from `target_ws`.
Now instead, clang-tidy is run from the folder of each individual source file separately. A corresponding unit test is added. Additionally, I enabled hooks `BEFORE_CLANG_TIDY_CHECKS` and `AFTER_CLANG_TIDY_CHECKS`. These could be used to e.g. show the list of applied checks etc. They are applied _once_ before/after all clang-tidy checks in the `target_ws`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -75,8 +75,9 @@ jobs:
           - {ROS_DISTRO: kinetic, TARGET_WORKSPACE: 'industrial_ci/mockups/format_tests/cpp/LLVM', CLANG_FORMAT_CHECK: 'WebKit', EXPECT_EXIT_CODE: 1}
 
           # Tidy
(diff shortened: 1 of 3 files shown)


rhaschke:
Currently, `.clang-tidy` config files are only considered in the root folder, `TARGET_REPO_PATH`. However, we should run clang-tidy for each individual file in its specific folder to handle any `.clang-tidy` file up the file hierarchy.

rhaschke:
This is finally ready for review :wink:

ipa-mdl (approved):
Looks good to me.
Should I just squash merge it with the PR title as a commit message?

rhaschke:
Yes, this can be squash-merged.

Issue #697, opened 2021-07-01
pre-release tests that more closely resemble what the buildfarm does for ros2


tylerjw:
We discovered recently is that the pre-release tests in industrial_ci use colcon but the official buildfarm does not.  This lead to builds passing in our pre-release tests and then failing on the buildfarm.

Quoting from here: https://github.com/ros-planning/moveit2/pull/532

> The issue is that ${PROJECT_NAME}_VERSION is not set automatically with colcon, it needs to be set explicitly using ament_package_xml(). Since the sonames are configured in all packages by setting the VERSION property, the actual file name ends up ending with a .. With this fix the sonames are set to the corresponding package versions for all libraries. In the future we should consider setting up explicit ABI versions.

If the pre-release test used `debuild` like the official buildfarm instead of colon it would have caught this issue.  I'm not sure how difficult it would be to implement that change.  Tagging @jspricke @henningkayser as they fixed the issue for our moveit2 release.

jspricke:
Note that you would also need to order the packages in one repo by dependency, run bloom, create a Debian source package and a minimal build environment (sbuild, pbuilder) on each. Not sure if it makes sense to do it in this project.

ipa-mdl:
> We discovered recently is that the pre-release tests in industrial_ci use colcon but the official buildfarm does not. This lead to builds passing in our pre-release tests and then failing on the buildfarm.

The `industrial_ci` pre-release test runs the pre-release test of `ros_buildfarm` (which uses colcon under the hood).
We just have some extra steps to setup docker for the different scenarios, make sure that it is not run as root and preparing `UPSTREAM_WORKSPACE` and `DOWNSTREAM_WORKSPACE`.

tylerjw:
> The `industrial_ci` pre-release test runs the pre-release test of `ros_buildfarm`

So maybe I should make this issue against the ros_buildfarm so hopefully they can have the official pre-release process more closely resemble the buildfarm?

jspricke:
I've written a simple script to do prerelease testing similar to the ROS buildfarm (i.e. without colcon), maybe you can integrate that:
```
#!/bin/sh -e
# run pre release tests for a repo or workspace
# setup sbuild: https://wiki.ubuntu.com/SimpleSbuild
# sudo apt install python3-colcon-package-information python3-bloom

REPO=$(mktemp -d)

# alternatively catkin_topological_order
for PKG_PATH in $(colcon list -tp); do
[...]

PR #719, opened 2021-07-22
Implement TEST=debians

ipa-mdl:


industrial_ci/src/tests/debians.sh
```diff
@@ -0,0 +1,64 @@
...
+}
+
+function build_and_install_debian() (
+    set -e
+    local pkg_path=$1
+    shift
+    cd "$pkg_path"
+    ici_quiet_echo bloom-generate rosdebian --ros-distro="$ROS_DISTRO" --debian-inc="ici~"
+    ici_quiet_echo dpkg-buildpackage -b -uc -us
+    mv ../*.{buildinfo,changes,deb} .
```
jspricke, line 31: superflous

industrial_ci/src/tests/debians.sh
```diff
@@ -0,0 +1,64 @@
...
+    ici_quiet_echo ici_asroot dpkg -i ./*.deb
```
jspricke, line 32:
```suggestion
    ici_quiet_echo ici_asroot apt-get install ./*.deb
```
(will resolve dependencies as well)

industrial_ci/src/tests/debians.sh
```diff
@@ -0,0 +1,64 @@
...
+    for name in upstream target downstream; do
+      local sources=()
+      local current="$sourcespace/$name"
+      ici_parse_env_array sources "${name^^}_WORKSPACE"
+      if [ -n "${sources[*]}" ]; then
+        ici_run "prepare_${name}_sourcespace" ici_prepare_sourcespace "$current" "${sources[@]}"
+        ici_run "install_${name}_dependencies" ici_quiet ici_install_dependencies "$UNDERLAY" "$ROSDEP_SKIP_KEYS" "$sourcespace"
+
+        while read -r -a pkg; do
+          ici_run "${pkg[0]}_debian" build_and_install_debian "$current/${pkg[1]}"
```
jspricke, line 60: this would still all be run in the same root filesystem, right?
The point of sbuild is to have a minimal chroot (container) (only essential packages installed) to find missing dependencies. With your setup if you have a repo with multiple ROS packages and package A ab B have a dependency X but only A declares it, packages B would still build if the dependency was installed for A before. That would not happen with sbuild nor work on the ROS build farm (and was the point in #697).

industrial_ci/src/tests/debians.sh
```diff
@@ -0,0 +1,64 @@
...
+    local sourcespace
+    ici_make_temp_dir sourcespace
+
+    for name in upstream target downstream; do
+      local sources=()
+      local current="$sourcespace/$name"
+      ici_parse_env_array sources "${name^^}_WORKSPACE"
+      if [ -n "${sources[*]}" ]; then
+        ici_run "prepare_${name}_sourcespace" ici_prepare_sourcespace "$current" "${sources[@]}"
+        ici_run "install_${name}_dependencies" ici_quiet ici_install_dependencies "$UNDERLAY" "$ROSDEP_SKIP_KEYS" "$sourcespace"
```
jspricke, line 57: I would recommend to move this after the bloom step and use 
```
apt-get build-dep .
```
The installation will be faster and you benefit from bloom resolving the dependencies.

jspricke (changes requested):

industrial_ci/src/tests/debians.sh
ipa-mdl: > (will resolve dependencies as well)

It should must resolve dependencies!
All dependencies should already be installed.

industrial_ci/src/tests/debians.sh
ipa-mdl: Why?

industrial_ci/src/tests/debians.sh
jspricke: you should only install build dependencies to build the package, but installing the build package could bring additional run dependencies.

industrial_ci/src/tests/debians.sh
jspricke: please give reasons why you need this (instead of just using `..` in the path for dpkg/apt below.

industrial_ci/src/tests/debians.sh
ipa-mdl: I see. Not sure how to solve this. `sbuild` is just too complicated to set-up (incl. some manual steps).
The ROS pre-release test builds each package in a clean environment already, so I wonder why it did not fail in the first place.

industrial_ci/src/tests/debians.sh
ipa-mdl: To only install a single *.deb file, and not all of them every time.

industrial_ci/src/tests/debians.sh
jspricke: which pre-release tests do you mean?
http://prerelease.ros.org/ does not, afaik.

industrial_ci/src/tests/debians.sh
ipa-mdl: > but installing the build package could bring additional run dependencies.

I just wanted to avoid pulling in the already built packages from the buildfarm.

jspricke: but then dpkg could fail and thus your CI job (I guess).

industrial_ci/src/tests/debians.sh
jspricke: hmkay, maybe add a comment then.

industrial_ci/src/tests/debians.sh
jspricke: Regarding sbuild being too complicated, you could spin up a new docker container for every package

industrial_ci/src/tests/debians.sh
ipa-mdl: > but then dpkg could fail and thus your CI job (I guess).

Yes, that's what I wanted to test for..

industrial_ci/src/tests/debians.sh
ipa-mdl: > which pre-release tests do you mean?

http://prerelease.ros.org/ does not, afaik.

This one, but with without the need to register it in rosdistro. 
I thought that it creates a new container (actually, new containers) for each package, but after digging into the log files this might not be the case :/

> Regarding sbuild being too complicated, you could spin up a new docker container for every package

Yes, I will try this next. Together with a shared, local repository.

ipa-mdl: The latest version does this, at least locally.
I will provide a test case as well later.

industrial_ci/src/tests/debians.sh
```diff
@@ -0,0 +1,118 @@
...
+    ici_exec bloom-generate rosdebian --ros-distro="$ROS_DISTRO" --debian-inc="ici~"
+    # https://github.com/ros-infrastructure/bloom/pull/643
+    echo 11 > debian/compat
+
+    ici_isolate debians build_debian "$current/${pkg[1]}" "$repo"
+    update_repo "$repo"
+)
+
+function run_debians() {
+    # shellcheck source=industrial_ci/src//isolation/docker.sh
```
jspricke, line 88: nit: '//'

industrial_ci/src/tests/debians.sh
```diff
@@ -0,0 +1,118 @@
...
+    source "${ICI_SRC_PATH}/isolation/docker.sh"
+    forward_mounts
+
+    export BUILDER=colcon
+    ici_source_builder
+    ici_run "${BUILDER}_setup" ici_quiet builder_setup
+    ici_run "setup_bloom" ici_quiet ici_install_pkgs_for_command bloom-generate python3-bloom debhelper
```
jspricke, line 95: why do you need debhelper here?

jspricke (approved):
I think this looks good (but didn't look to deep), Thanks for working on this!

Maybe a test with multiple packages would be a good addition (did you try it with Moveit already?)

industrial_ci/src/tests/debians.sh
ipa-mdl: It is not needed anymore, I will remove it!

ipa-mdl:
> Maybe a test with multiple packages would be a good addition

The testcase already builds two packages

> did you try it with Moveit already?

I am on it: https://github.com/ipa-mdl/industrial_ci/actions/runs/1059400937

industrial_ci/src/tests/debians.sh
jspricke: Just for completeness, I think it could happen that colcons topological sort builds a package A before B but A needs B to get installed. So it could be that a package is not installable after it was build. You could add an install test with all packages at the end, maybe.

jspricke:
> The testcase already builds two packages

Ah, missed the `industrial-ci` package. Maybe add a dependency between them?

> I am on it: https://github.com/ipa-mdl/industrial_ci/actions/runs/1059400937

Awesome, nice result :).

ipa-mdl:
> Awesome, nice result :).

Actually, it failed for a different reason (I believe because of repo mixups)
However, https://github.com/ipa-mdl/industrial_ci/actions/runs/1059855175 fails with:

```
dpkg-shlibdeps: error: no dependency information found for /opt/ros/galactic/lib/libmoveit_ros_occupancy_map_monitor.so. (used by debian/ros-galactic-moveit-ros-planning/opt/ros/galactic/lib/libmoveit_planning_scene_monitor.so.)
Hint: check if the library actually comes from a package.
```
Just like the build farm.

> Maybe add a dependency between them?

The two packages must be used separately.
But we should add a proper test case. Either an artificial one or a real pepository.

PR #727, opened 2021-07-25
Reduce overall verbosity

### LINKED ISSUE #712: Silence more steps
The package installation of utilities is very noise, use `ici_quiet` in more places.
Perhaps with an opt-out.

### LINKED ISSUE #728: Implement TRACE option
mostly for debugging, based on #727

ipa-mdl:
* list most commands (ici_cmd)
* filter apt-get output (ici_filtered)
* rename ici_run to ici_step

There is still room for improvement, but this can be optimized over time

@rhaschke: Please have a look at the new output format as this addresses your concerns about the verbosity in #452

closes #712 
closes #728 

includes #733 and #731


industrial_ci/src/util.sh
```diff
@@ -313,24 +314,81 @@ function ici_retry {
...
+    local err=${PIPESTATUS[0]}
+    if [ "$err" -ne 0 ]; then
+        ici_error "'$cmd' returned with $err" "$err"
+    fi
+}
+
+function ici_cmd_quiet {
+    local cmd; cmd=$(ici_get_log_cmd "$@")
+    ici_log
+    ici_color_output ${ANSI_BOLD} "$ $cmd* &> /dev/null"
```
ipa-mdl, line 371: Remove asterisk

rhaschke:
Looks good to me. But I noticed that coloring is missing for the build step, e.g. in https://github.com/ros-industrial/industrial_ci/runs/3156358865.

ipa-mdl:
>  But I noticed that coloring is missing for the build step, e.g. in https://github.com/ros-industrial/industrial_ci/runs/3156358865.

Where exactly?

rhaschke:
While the output of `catkin config` is still colored, the output of `catkin build` is not.

ipa-mdl:
> While the output of catkin config is still colored, the output of catkin build is not.

Ah, I see. It is not running in a tty anymore, because it is piped.
For now I might take the pipe away and get rid of `set -e`, which does not work reliably.

ipa-mdl:
@rhaschke: please have a look again. The "industrial_ci` have traces enables, the others not.

In a follow-up PR I will make the commands top-level folds.

PR #739, opened 2021-08-12
added special workspace setup for just one repo

dhled:
Hello, 

I added this parsing command for url, if you just want to workspace extend with one private repo using ssh without creating a ".repos" or ".rosinstall" file (very niche case I guess).

Maybe there was already a way to achieve that but as I couldn't, I added this 'git+ssh'.

Does it make sense ?

Thank you !

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -314,6 +314,11 @@ Or mixed:
 
   DOWNSTREAM_WORKSPACE="github:ros-simulation/gazebo_ros_pkgs@melodic-devel https://raw.githubusercontent.com/ros-controls/ros_control/melodic-devel/ros_control.rosinstall -ros_control additional.repos"
 
+To depend on a different repository of a private server using git and the SSH protocol:
+::
+
+  UPSTREAM_WORKSPACE='git+ssh://git@private.server.net/repository#branch'
+
 To filter the target workspace:
 ::
 
--- a/industrial_ci/src/workspace.sh
+++ b/industrial_ci/src/workspace.sh
@@ -37,7 +37,7 @@ function ici_parse_repository_url {
             gitlab | gl)
                 echo "${name%.git}" "git" "https://gitlab.com/$repo" "$fragment"
                 ;;
-            'git+file'*|'git+http'*)
+            'git+file'*|'git+http'*|'git+ssh'*)
                 echo "${name%.git}" "git" "${scheme#git+}:$repo" "$fragment"
                 ;;
             git+*)


ipa-mdl (changes requested):
It should just work right away: `git+ssh://git@github.com/ros-industrial/industrial_ci.git#master`.

Please try it out and then we should extend the documentation.

ipa-mdl:
Actually, `git+ssh` is deprecated. I would rewrite to `ssh` as it is done for `file`.

industrial_ci/src/workspace.sh
```diff
@@ -40,6 +40,9 @@ function ici_parse_repository_url {
             'git+file'*|'git+http'*)
                 echo "${name%.git}" "git" "${scheme#git+}:$repo" "$fragment"
                 ;;
+            'git+ssh')
+                echo "${name%.git}" "git" "git@$repo" "$fragment"
+                ;;
```
ipa-mdl, line 45:
```suggestion
```

dhled: Done

industrial_ci/src/workspace.sh
```diff
@@ -40,6 +40,9 @@ function ici_parse_repository_url {
             'git+file'*|'git+http'*)
```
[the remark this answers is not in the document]
dhled: Done

dhled:
Thank you ! It should work with this :). I have added an example in the documentation.

doc/index.rst
```diff
@@ -314,6 +314,13 @@ Or mixed:
 
   DOWNSTREAM_WORKSPACE="github:ros-simulation/gazebo_ros_pkgs@melodic-devel https://raw.githubusercontent.com/ros-controls/ros_control/melodic-devel/ros_control.rosinstall -ros_control additional.repos"
 
+To depend on a different repository of a private server using git and the SSH protocol:
+::
+
+  UPSTREAM_WORKSPACE='git+ssh://git@private.server.net/repository#branch'
+
+::
+
```
ipa-mdl, line 323:
```suggestion
```

?

ipa-mdl, line 323:
```suggestion
To depend on a different repository of a private server using git and the SSH protocol:
::

  UPSTREAM_WORKSPACE='git+ssh://git@private.server.net/repository#branch'

```

ipa-mdl:
Shellcheck is failing for some time now, but I did not have the time yet to fix these issues.
Please update/squash your commits for now. I will try to fix shellcheck.

dhled:
Thank you, updated and squashed

ipa-mdl (approved):

PR #743, opened 2021-09-21
Bundle industrial_ci, if run from act

ipa-mdl:


industrial_ci/src/isolation/docker.sh
```diff
@@ -155,6 +160,10 @@ function ici_run_cmd_in_docker() {
     docker_cp "$d" "$cid:${docker_query[*]:2}/" "${docker_query[0]}" "${docker_query[1]}"
   done
 
+  if [ "$_BUNDLE_ICI" = true ]; then
+    tar -c -f - "$ICI_SRC_PATH" | docker cp - "$cid:/"
```
ipa-mdl, line 164: should strip leading slash or silence stderr

destogl:
@ipa-mdl thanks for this this! It is working!

I am using the following configuration (example tested with ros-controls/ros2_control repository):
```
jobs:
  ci_binary:
    name: Galactic binary job
    runs-on: ubuntu-latest
    strategy:
      matrix:
        env:
          - {ROS_DISTRO: galactic, ROS_REPO: main}
          - {ROS_DISTRO: galactic, ROS_REPO: testing}
    steps:
      - uses: actions/checkout@v1
      - uses: ipa-mdl/industrial_ci@feature/act-support
        with:
          config: ${{toJSON(matrix.env)}}
```

gavanderhoorn:
If this works you might want to 'advertise' it a little @ipa-mdl.

Could save tenths of notifications each day by people pushing "fixup" and "test CI" commits (as they could now run it locally).

destogl:
@ipa-mdl do you want to write short “manual” or should I. I will be glad to review it at lest :robot:

ipa-mdl:
> do you want to write short “manual” or should I.

It would be great if you could draft one.

ipa-mdl:
> in what way? You mean configs need to be changed to no longer use the matrix: ${{ toJSON(matrix) }} syntax?

@gavanderhoorn: Away from `${{ matrix.env }}`, either by passing all env entries explicitly (recommended for simple cases) or by using `config: ${{toJSON(matrix)}}`

.github/action.sh
```diff
@@ -42,4 +42,9 @@ if [ "${ABICHECK_MERGE:-}" = "auto" ]; then
   [ "$GITHUB_EVENT_NAME" != "pull_request" ] || ABICHECK_MERGE=true
 fi
 
-env "$@" bash "$DIR_THIS/../industrial_ci/src/ci_main.sh"
+if [[ $ICI_SRC_PATH == *run/act/* ]]; then
```
ipa-mdl, line 45: I had to relax this test a little bit. Not sure why, but in one of my tests it was `/run/act` instead of `/var/run/act`.

PR #752, opened 2021-10-08
Update Github Actions template

### LINKED ISSUE #590: Update Github Action template

ipa-mdl:
closes #590 

Based on https://github.com/ros-industrial/fanuc/blob/melodic-devel/.github/workflows/ci_bionic.yml


ipa-mdl:
@gavanderhoorn: Please have a look

PR #755, opened 2021-10-19
Report exit codes of individual steps to github actions

### LINKED ISSUE #753: Provide feedback on outcome of individual steps
To automate various actions in GHA depending on the outcome of steps in `industrial_ci`, it would be great to have the exit code of individual steps as [output feedback](https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-an-output-parameter).

```yaml
steps:
  - id: industrial_ci
    uses: ros-industrial/industrial_ci@master
  - name: Upload test artifacts (on test failure)
     uses: actions/upload-artifact@v2
     if: steps.industrial_ci.outputs.run_target_test != 0
     with:
          name: test-results
          path: ${{ env.BASEDIR }}/target_ws/**/test_results/**/*.xml
```

rhaschke:
Report exit codes of (most) individual steps to github actions. 
Fixes #753.

This PR essentially introduces the new CI-specific [function `ici_report_result`](https://github.com/ros-industrial/industrial_ci/pull/755/files#diff-7750e2106e8f18fadcb9a46ba499ee6d1827d5c876167a39ac693710bfabf69aR28-R30), which is called at the end of each timed build step. To [expose the produced outputs to the composite github action](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#outputs-for-composite-actions), they need to be [declared in `action.yml`](https://github.com/ros-industrial/industrial_ci/pull/755/files#diff-1243c5424efaaa19bd8e813c5e6f6da46316e63761421b3e5f5c8ced9a36e6b6R9-R70). Finally, I added a simple unit test for this new functionality.


rhaschke:
@ipa-mdl, could you please have a look at this simple new feature and eventually merge?

ipa-mdl:
> could you please have a look at this simple new feature and eventually merge?

Looks like I did not submit my review back then for some reason..

action.yml
```diff
@@ -1,12 +1,78 @@
 name: 'ROS industrial_ci'
+description: "unning various CI checks for ROS package repositories"
```
ipa-mdl, line 2: Typo

industrial_ci/src/folding/github_actions.sh
```diff
@@ -24,3 +24,7 @@ function  ici_end_fold() {
     shift 4
     echo -e "##[endgroup]"
 }
+
+function ici_report_result() {
+    echo "::set-output name=$1::$2"
+}
```
ipa-mdl, line 30: Please add line endings

industrial_ci/src/folding/gitlab.sh
```diff
@@ -29,3 +29,7 @@ function  ici_end_fold() {
     local end=$1; shift
     ici_ansi_cleared_line "section_end:${end::-9}:$name"
 }
+
+function ici_report_result() {
+    true
+}
```
ipa-mdl, line 35: line ending

industrial_ci/src/folding/none.sh
```diff
@@ -24,3 +24,7 @@ function  ici_end_fold() {
     shift 4
     echo "<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<"
 }
+
+function ici_report_result() {
+    true
+}
```
ipa-mdl, line 30: line ending

ipa-mdl (dismissed):
In principle I am okay with this change.
However, I don't like listing the return code of the hooks (which are dynamic)..

Would it work to return the name (string..) and the exit code instead?

rhaschke:
> However, I don't like listing the return code of the hooks (which are dynamic).

That's the aim: I want to access the return codes of individual steps to steer subsequent GHA steps.
See [here](https://github.com/ros-planning/moveit/blob/27b4290f1c828192c98bc64ba31329c23834518a/.github/workflows/ci.yaml#L120) for an examplary use case.

> Would it work to return the name (string..) and the exit code instead?

Which name do you mean, `strerror()`? I would need to parse that name, later on, to access the exit code again. That doesn't make sense to me.

ipa-mdl:
> Which name do you mean, strerror()? I would need to parse that name, later on, to access the exit code again. That doesn't make sense to me.

```
failure() && (steps.ici.outputs.last_step == "run_target_test" || steps.ici.outputs.last_step == "target_test_results")
```

ipa-mdl:
We can still have a list with certain outputs, but only for a small selection of (new) checkpoints..
This would go more into the direction of stages.

* `ici_checkpoint setup_done`
* `ici_checkpoint upstream_done`
*  `ici_checkpoint target_done`
* `ici_checkpoint downstream_done`
* `ici_checkpoint all_done`

rhaschke:
Thanks for the clarification. Your idea of returning the name of the last executed stage is appealing. It easily allows checking for the name of the last executed (or failed) stage.
However it makes it hard to check that a certain (early) stage _succeeded_: You would need to test against all subsequent names.
I agree that the nasty point about my proposal is the explicit list of exports, which is hard to maintain and - as you pointed out - not comprehensive yet. But, it's a starting point :wink:

ipa-mdl:
> Return a dictionary (JSON..)

Should work: https://github.com/actions/javascript-action/blob/c1c60989a13ec68893bc5ea614991fba4aab8644/dist/index.js#L52

rhaschke:
> ```failure() && (steps.ici.outputs.steps.run_target_test || steps.ici.outputs.steps.target_test_results)```

I like this dictionary variant. Not sure, though, it will actually work. Your reference is _generating_ output not _parsing_, isn't it?
Also, we would need to collect processed steps (and their return values?) in a global array and generate the corresponding output after each step anew.

rhaschke:
As this blog post indicates, parsing of JSON outputs is not done automatically. However, doing it manually, we loose most of the benefits of your proposal :disappointed:

ipa-mdl:
> parsing of JSON outputs is not done automatically.

Hmm.
We could try to call `parseJSON` in `action.yml`.

ipa-mdl:
> generate the corresponding output after each step anew.

just in the exit handler..

I think it is worth a try.
If we cannot read JSON, then your current version is the way to go!

ipa-mdl:
Promising: https://docs.github.com/en/actions/learn-github-actions/expressions#fromjson

rhaschke:
Sure, but the test becomes somewhat uglier:
```failure() && (fromJSON(steps.ici.outputs.steps).run_target_test || fromJSON(steps.ici.outputs.steps).target_test_results)```

**Comment by ipa-mdl** [comment]
> Sure, but the test becomes somewhat uglier:

There is always a catch ;)
`fromJSON` works in `action.yml`, but gets rejected by some validator ("A mapping was not expected")

**Comment by ipa-mdl** [comment]
It [ain't stupid](https://github.com/ipa-mdl/industrial_ci/commit/f59a29bb88723ea72d3e3823da6e119e8c1c0d5c) if [it works](https://github.com/ipa-mdl/industrial_ci/runs/4098737496?check_suite_focus=true#step:4:4)

**Comment by rhaschke** [comment]
Directly running a JS file you aim to avoid the composite action and thus manually expose all outputs? Nice idea!

**Comment by ipa-mdl** [comment]
> Nice idea!

https://github.community/t/dynamic-outputs-for-action/178457/3

Luckily, I had the JS version before composite actions were available (#576).

**Comment by ipa-mdl** [comment]
https://github.com/rhaschke/industrial_ci/pull/1

**Review by ipa-mdl — APPROVED** [review]

**Comment by ipa-mdl** [comment]
@rhaschke: Could you clean-up the commits or should I squash-merge?

**Comment by rhaschke** [comment]
Please just squash-merge.

File: action.yml [hunk 5]
```diff
@@ -1,12 +1,10 @@
...
   config:
     description: 'Configuration in JSON format, e.g. from matrix'
     required: false
 runs:
-  using: "composite"
-  steps:
-    - run: $GITHUB_ACTION_PATH/.github/action.sh
-      shell: bash
-      env:
-        INPUT_CONFIG: ${{ inputs.config }}
```
rhaschke, line 12: You lost the `INPUT_CONFIG`, didn't you?

ipa-mdl: No.. it should still work.
If I recall it correctly, for composite actions you have to pass the inputs explicitly.
The tests use config, so I am sure that this works.

rhaschke:
Thanks for your effort on this! :tada:

PR #763, opened 2021-12-29
Fix folder removal

rhaschke:
Trying to keep a single folder from a larger source repository, I ran into several issues:
1. It's not easy to keep a single folder, but delete all others
2. Deleting folders as [described in the docs](https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst#examples) [doesn't work](https://github.com/rhaschke/test/runs/4654230318):
`UPSTREAM_WORKSPACE='github:ros-controls/ros_control#melodic-devel -rqt_controller_manager'`
The folder should be specified with the repository prefix: `-ros_control/rqt_controller_manager`. However, this is clumsy.

This PR
1. remembers the last imported repository name
2. tries to delete a folder from that repo (thus matching the docs)
3. adds syntax `+folder` to keep a specific folder

Example run: https://github.com/rhaschke/test/runs/4654479819

### DIFF
--- a/industrial_ci/src/workspace.sh
+++ b/industrial_ci/src/workspace.sh
@@ -268,6 +268,7 @@ function  ici_import_directory {
 function ici_prepare_sourcespace {
     local sourcespace=$1; shift
     local basepath=$TARGET_REPO_PATH
+    local last=""
 
     ici_guard mkdir -p "$sourcespace"
 
@@ -275,6 +276,10 @@ function ici_prepare_sourcespace {
         case "$source" in
         git* | bitbucket:* | bb:* | gh:* | gl:*)
             ici_import_repository "$sourcespace" "$source"
+            # Remember imported repository in variable last
+            local parsed; parsed=$(ici_parse_repository_url "$source")
+            IFS=" " read -r -a parts <<< "$parsed" # name, type, url, version
+            last=${parts[0]}
             ;;
         http://* | https://*) # When UPSTREAM_WORKSPACE is an http url, use it directly
             ici_import_url "$sourcespace" "$source"
@@ -285,10 +290,9 @@ function ici_prepare_sourcespace {
             ici_guard rm -r "${sourcespace:?}/$file"
             ;;
         -*)
-            local file="${source:1}"
-            if [ ! -e "${sourcespace:?}/$file" ]; then
-              file="$(basename "$basepath")/$file"
-            fi
+            for file in "${source:1}" "$last/${source:1}" "$(basename "$basepath")/${source:1}"; do
+                if [ -e "${sourcespace:?}/$file" ]; then break; fi
+            done
             ici_log "Removing '${sourcespace:?}/$file'"
             ici_guard rm -r "${sourcespace:?}/$file"
             ;;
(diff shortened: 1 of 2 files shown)


ipa-mdl:
> Deleting folders as described in the docs doesn't work:

Oh, the documentation needs to be fixed!

> remembers the last imported repository name

This only works with directly specified repositories, not with rosinstall files (which are recommended for larger setups)

> tries to delete a folder from that repo (thus matching the docs)

This adds ambiguity, especially, if you mix instructions.

industrial_ci/src/workspace.sh
```diff
@@ -285,13 +290,19 @@ function ici_prepare_sourcespace {
...
+            done
             ici_log "Removing '${sourcespace:?}/$file'"
             ici_guard rm -r "${sourcespace:?}/$file"
             ;;
+        +*)
+            for file in "${source:1}" "$last/${source:1}"; do
+                if [ -e "${sourcespace:?}/$file" ]; then break; fi
+            done
+            ici_log "Moving '${sourcespace:?}/$file' to '${sourcespace:?}'"
+            ici_guard mv "${sourcespace:?}/$file" "${sourcespace:?}"
```
ipa-mdl, line 304: This does not copy parent folders..

rhaschke: Which parent folder you are referring to? I'm assuming that you would move folders below `$sourcespace` only.

rhaschke:
I'm not quite sure how to interpret your feedback.
I guess you prefer keeping fully-specified paths only and fixing the docs appropriately?
What do you think about introducing `+folder syntax` to facilitate keeping a few folders only? Do you have other suggestions for this use case?

ipa-mdl:
> Trying to keep a single folder from a larger source repository,

I was told that this could be done easily with SVN (which is still supported by Github).
But it would still make sense to come up with a better filter syntax.

Something like:
`github:ros-controls/ros_control#melodic-devel(+rqt_controller_manager,-rqt_controller_manager/CHANGELOG.rst)`

rhaschke:
> I was told that this could be done easily with SVN (which is still supported by Github).

You can use git-svn to clone an svn repo with git, but not the other way around as far as I know.

> But it would still make sense to come up with a better filter syntax, e.g.
> `github:ros-controls/ros_control#melodic-devel(+rqt_controller_manager,-rqt_controller_manager/CHANGELOG.rst)`

That would clarify that the filter applies to the specific repo, which is good.
Do you plan to drop the generic `-folder` option then? This was much more generic, allowing to remove any file/folder from the source space.

ipa-mdl:
> You can use git-svn to clone an svn repo with git, but not the other way around as far as I know.

Apparently, it works the other way around as well.
I will investigate ;)

> Do you plan to drop the generic -folder option then?

No, the old options should stay (used with rosinstall).
The new syntax would be special for all cases in which the target folder is known (vcs, path).

rhaschke:
Closing as Mathias will suggest a more specific approach.

industrial_ci/src/workspace.sh
ipa-mdl: `repo/sub1/sub2/abc` would end up in `${sourcespace}/abc`

rhaschke: Yes. Does that harm?
Of course, your syntax proposal that filters folders within a repo parent folder, will be much better. No risk for conflicts ;-)

PR #768, opened 2022-01-06
Run clang_format_check in deterministic location

wxmerkt:
This is a proposal to implement #737. It enables to use the `AFTER_INSTALL_CLANG_FORMAT_CHECK` hook to download a clang-format file into `/tmp/clang_format_check/$PROJECT_NAME`


industrial_ci/src/tests/clang_format_check.sh
```diff
@@ -19,10 +19,10 @@ function prepare_clang_format_check() {
   true
 }
 
-function run_clang_format_check() {
-  local err=0
-  local path
-  ici_make_temp_dir path
+function install_clang_format_check() {
```
mathias-luedtke, line 22: This function never gets called..

mathias-luedtke (changes requested):
This is not working at all. That's why the tests fail.

industrial_ci/src/tests/clang_format_check.sh
wxmerkt: Indeed, thanks for pointing this out - I made a wrong assumption about the calling order. I've now simplified it. It works by e.g. setting `BEFORE_RUN_CLANG_FORMAT_CHECK="wget [my config] -O /tmp/clang_format_check/[project]/.clang-format"` along with `ADDITIONAL_DEBS=wget` in this case.

wxmerkt:
Friendly ping - any thoughts on this?

PR #776, opened 2022-02-18
[`abi_check`] Pass quoted `cflags` to colcon

aprotyas:
The `colcon` invocation in `abi_process_workspace` was only providing
the first of the two `cflags`, hence omitting `-Og`. To work around
this, the flags must be enclosed by literal quotation marks.

This commit escapes the flag specification to insert these literal
quotes around the flags.

Signed-off-by: Abrar Rahman Protyasha <aprotyas@u.rochester.edu>

### DIFF
--- a/industrial_ci/src/tests/abi_check.sh
+++ b/industrial_ci/src/tests/abi_check.sh
@@ -81,7 +81,7 @@ function abi_process_workspace() {
   local version=${1:-$tag}
 
   local cflags="-g -Og"
-  local cmake_args=(--cmake-args "-DCMAKE_C_FLAGS=$cflags" "-DCMAKE_CXX_FLAGS=$cflags")
+  local cmake_args=(--cmake-args "-DCMAKE_C_FLAGS=\"$cflags\"" "-DCMAKE_CXX_FLAGS=\"$cflags\"")
 
   ici_step "install_${tag}_dependencies" ici_install_dependencies "$extend" "$ROSDEP_SKIP_KEYS" "$workspace/src"
   ici_step "abi_build_${tag}" builder_run_build "$extend" "$workspace" "${cmake_args[@]}"


aprotyas:
Friendly ping for thoughts/review @mathias-luedtke

industrial_ci/src/tests/abi_check.sh
```diff
@@ -81,7 +81,7 @@ function abi_process_workspace() {
   local version=${1:-$tag}
 
   local cflags="-g -Og"
-  local cmake_args=(--cmake-args "-DCMAKE_C_FLAGS=$cflags" "-DCMAKE_CXX_FLAGS=$cflags")
+  local cmake_args=(--cmake-args "-DCMAKE_C_FLAGS=\"$cflags\"" "-DCMAKE_CXX_FLAGS=\"$cflags\"")
```
mathias-luedtke, line 84: Have you tried it with single quotes?

mathias-luedtke (changes requested):
Thank you very much!
Your patch seems to fix it.
Could you add a test as well, so that it will stay this way in the future?

industrial_ci/src/tests/abi_check.sh
aprotyas: Tried! It works: 6cd3837.

aprotyas:
@mathias-luedtke thanks for the review. I'm not sure how to test for this behavior though since its difficult to observe the abi_check interface behavior without parsing compiler command logs. Thoughts?

mathias-luedtke:
> I'm not sure how to test for this behavior though since its difficult to observe the abi_check interface behavior without parsing compiler command logs. Thoughts?

You wrote:
> This should generate a compile_commands.json containing the exact compiler calls for all translation units of the project in ros2_ws/build/pkg_name

So why not just check the compile_commands.json? For example with ' grep -q`. I might add this later.

PR #785, opened 2022-06-29
catkin tools should call `catkin test` for testing

rhaschke:
Calling `catkin build --catkin-make-args run_tests` is only correct for catkin packages, not for pure cmake packages, which require  `catkin build --catkin-make-args test`. `catkin test` selects the correct build target automatically:
https://catkin-tools.readthedocs.io/en/latest/verbs/catkin_test.html?highlight=run_tests#basic-usage

### DIFF
--- a/industrial_ci/src/builders/catkin_tools.sh
+++ b/industrial_ci/src/builders/catkin_tools.sh
@@ -59,7 +59,7 @@ function builder_run_tests {
         opts+=(-i)
     fi
     _append_job_opts opts PARALLEL_TESTS 1
-    ici_cmd ici_exec_in_workspace "$extend" "$ws" catkin build --catkin-make-args run_tests -- "${opts[@]}" --no-status
+    ici_cmd ici_exec_in_workspace "$extend" "$ws" catkin test "${opts[@]}" --no-status
 }
 
 function builder_test_results {
--- a/industrial_ci/src/builders/catkin_tools.sh
+++ b/industrial_ci/src/builders/catkin_tools.sh
@@ -59,7 +59,7 @@ function builder_run_tests {
         opts+=(-i)
     fi
     _append_job_opts opts PARALLEL_TESTS 1
-    ici_cmd ici_exec_in_workspace "$extend" "$ws" catkin test "${opts[@]}" --no-status
+    ici_cmd ici_exec_in_workspace "$extend" "$ws" catkin run_tests "${opts[@]}" --no-status
 }
 
 function builder_test_results {


rhaschke:
Looks like the mentioned feature is only available in Noetic (running catkin-tools 0.9.0). I originally ran into this issue with [`tesseract`](https://github.com/tesseract-robotics/tesseract), which uses plain cmake packages and [didn't actually run tests by default](https://github.com/tesseract-robotics/tesseract/runs/7096725952?check_suite_focus=true) (but only with an [additional `AFTER_SCRIPT` command](https://github.com/tesseract-robotics/tesseract/blob/98964bca335bfd29fcb66738dfe7bc46c191649a/.github/workflows/focal_build.yml#L98)).

rhaschke:
To be compatible between catkin tools 0.9 and older versions, now `catkin run_tests` is called.
For [recent versions this maps onto `catkin test`](https://github.com/ros-industrial/industrial_ci/runs/7131657117?check_suite_focus=true) and distinguishes between catkin and pure cmake packages.
For [older versions this maps onto `catkin build --verbose --catkin-make-args run_tests --`](https://github.com/ros-industrial/industrial_ci/runs/7131660341?check_suite_focus=true), i.e. the old call.

The [failing test](https://github.com/ros-industrial/industrial_ci/runs/7131653798?check_suite_focus=true) seems to be unrelated to this change.

industrial_ci/src/builders/catkin_tools.sh
```diff
@@ -59,7 +59,7 @@ function builder_run_tests {
         opts+=(-i)
     fi
     _append_job_opts opts PARALLEL_TESTS 1
-    ici_cmd ici_exec_in_workspace "$extend" "$ws" catkin build --catkin-make-args run_tests -- "${opts[@]}" --no-status
+    ici_cmd ici_exec_in_workspace "$extend" "$ws" catkin run_tests "${opts[@]}" --no-status
```
mathias-luedtke, line 62: There was a reason for using this instead of run_tests, but I can't recall it at the moment.
I guess it did not exists for indigo..

mathias-luedtke (commented):
Good catch and neat solution! :D

I just want to double check, why I could not use run_tests in the first place.

industrial_ci/src/builders/catkin_tools.sh
mathias-luedtke: I got introduced in https://github.com/ros-industrial/industrial_ci/pull/255, so I it should not be needed anymore.
I will run a local test against indigo to be sure.

mathias-luedtke (approved):

PR #789, opened 2022-07-23
Fix shellcheck issues

mathias-luedtke:
#788 and more


mathias-luedtke:
The galactic jobs are failing because of https://github.com/osrf/infrastructure/issues/12

PR #790, opened 2022-07-27
Support VERBOSE_OUTPUT for colcon

hoffmann-stefan:
This adds support for VERBOSE_OUTPUT to colcon.
It enables output for stdout (`console_cohesion+`) which would otherwise only be written to a logfile and not accessible to debug in CI.

cc @fmessmer

### DIFF
--- a/industrial_ci/src/builders/colcon.sh
+++ b/industrial_ci/src/builders/colcon.sh
@@ -28,11 +28,10 @@ function builder_setup {
 function builder_run_build {
     local extend=$1; shift
     local ws=$1; shift
-    local output_handler
+    local opts=(--event-handlers "${_colcon_event_handlers[@]}")
     if [ "${VERBOSE_OUTPUT:-false}" != false ]; then
-        output_handler="console_cohesion+"
+        opts+=("console_cohesion+")
     fi
-    local opts=(--event-handlers "${_colcon_event_handlers[@]}" "${output_handler}")
     local jobs
     ici_parse_jobs jobs PARALLEL_BUILDS 0
     if [ "$jobs" -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


industrial_ci/src/builders/colcon.sh
```diff
@@ -28,7 +28,11 @@ function builder_setup {
 function builder_run_build {
     local extend=$1; shift
     local ws=$1; shift
-    local opts=(--event-handlers "${_colcon_event_handlers[@]}")
+    local output_handler
+    if [ "${VERBOSE_OUTPUT:-false}" != false ]; then
+        output_handler="console_cohesion+"
+    fi
+    local opts=(--event-handlers "${_colcon_event_handlers[@]}" "${output_handler}")
```
mathias-luedtke, line 35: This will not work properly, because it adds an empty string, if VERBOSE_OUTPUT=false

Please try
```suggestion
    local opts=(--event-handlers "${_colcon_event_handlers[@]}")
    if [ "${VERBOSE_OUTPUT:-false}" != false ]; then
        opts+(="console_cohesion+")
    fi
```

mathias-luedtke (changes requested):
Thank you very for your first contribution!
Please fix the code as indicated and then let's cross fingers that all jobs will pass :D

industrial_ci/src/builders/colcon.sh
hoffmann-stefan: Thanks :) Commited the change (with small typo fix) and tested it in our internal repo, still worked as expected.

mathias-luedtke:
@hoffmann-stefan: All tests passed :)
Please confirm that the output is visible in you internal builds.

industrial_ci/src/builders/colcon.sh
mathias-luedtke: Ah, this got hidden already ;)
Just ignore my other comment..

hoffmann-stefan:
> Ah, this got hidden already ;)
> Just ignore my other comment..

No worries, thank you for the fast feedback and merge :)

PR #797, opened 2022-10-20
Fix GHA warnings

rhaschke:
Fix warnings about deprecated node.js 12 actions. Updates to latest versions.
Also shorten CI job names to better see the actual job description


mathias-luedtke (approved):
Great!

I was already wondering why https://github.com/ros-industrial/industrial_ci/pull/794 did not address all warning, but did not have time yet to look into it.

PR #800, opened 2022-12-14
ROS2 galactic is EOL

mathias-luedtke:


Nir-Az:
So now galactic cannot run anymore?
We get failures on running industrial-ci on it

Issue #807, opened 2023-01-10
Trouble cloning multiple private repositories from .repos file


TrevorGibson-SR:
Hi all,

I've been trying to get a GitHub Actions CI build to work with cloning more than one private repository, but have run into multiple issues and can't seem to resolve them.

### Overview
The current Action I have set up works fine specifically when only 1 private GitHub repository is listed in our `.repos` file. It breaks when either a second private GitHub repo is listed (issue 1), or any private BitBucket repositories are listed (issue 2).

Some additional information:
- Any number of public repositories are able to cloned successfully without issue
- Cloning multiple private repos from a local development machine (using `vcs import`) works fine, regardless of the source of those repos (GitHub or BitBucket)

Here is a snippet of the .yml action showing the steps being run (placeholders in place of actual repo URLs):

```
reusable_industrial_ci_with_cache:
    name: ${{ inputs.ros_distro }} ${{ inputs.ros_repo }} ${{ inputs.os_code_name }}
    runs-on: ubuntu-latest
    env:
      CCACHE_DIR: ${{ github.workspace }}/${{ inputs.ccache_dir }}
      BASEDIR: ${{ github.workspace }}/${{ inputs.basedir }}
      CACHE_PREFIX: ${{ inputs.ros_distro }}-${{ inputs.os_code_name }}-${{ inputs.ros_repo }}-${{ github.job }}
    strategy:
      fail-fast: false
    steps:
      - name: Checkout ${{ inputs.ref }} when build is not scheduled
        if: ${{ github.event_name != 'schedule' }}
        uses: actions/checkout@v3
      - name: Checkout ${{ inputs.ref }} on scheduled build
        if: ${{ github.event_name == 'schedule' }}
        uses: actions/checkout@v3
        with:
          ref: ${{ inputs.ref_for_scheduled_build }}
      - name: Setup SSH keys for private repositories 1
        uses: webfactory/ssh-agent@v0.7.0
        with:
          ssh-private-key: |
            ${{ secrets.KEY1 }}
            ${{ secrets.KEY2 }}
            ${{ secrets.KEY3 }}
      - name: Prepare git and ssh config for build context
        run: |
          mkdir root-config
          cp -r ~/.gitconfig ~/.ssh root-config/
      - name: cache target_ws
        if: ${{ ! matrix.env.CCOV }}
        uses: pat-s/always-upload-cache@v2.1.5
        with:
          path: ${{ env.BASEDIR }}/target_ws
[...]

PR #815, opened 2023-02-08
Add support for multiple GitHub deploy key mappings with GitHub Actions

TrevorGibson-SR:
Proposed as a solution to #807 

When cloning multiple private GitHub repositories with SSH, you can use the [deploy key mapping feature](https://github.com/webfactory/ssh-agent#support-for-github-deploy-keys) of the webfactory/ssh-agent GitHub action to map specific SSH keys to their respective GitHub repositories. This works as intended when cloning in the main GitHub runner environment, but fails to translate over to the industrial_ci docker environment.

Webfactory/ssh-agent [specifically calls this out](https://github.com/webfactory/ssh-agent#using-multiple-deploy-keys-inside-docker-builds) and suggests to copy the `~/.ssh` directory and `~/.gitconfig` file into the docker container to ensure the mappings are carried over. This change simply adds two new docker mount arguments to import the SSH configuration into the Docker container's build environment.


industrial_ci/src/isolation/docker.sh
```diff
@@ -95,6 +95,10 @@ function ici_isolate() {
   ici_forward_mount SSH_AUTH_SOCK rw # forward ssh agent into docker container
   ici_forward_mount GITHUB_OUTPUT rw # for ici_report_result in GitHub Actions
 
+  # For GitHub Actions, forward ssh key mappings for multiple github deploy keys generated by webfactory/ssh-agent
+  ici_forward_mount /home/runner/.gitconfig rw
+  ici_forward_mount /home/runner/.ssh rw
```
mathias-luedtke, line 100: We cannot hardcode Gihub-specific paths.

mathias-luedtke:
Please try adding .gitconfig here:
https://github.com/ros-industrial/industrial_ci/blob/26ac1310cddeeba0b545f629e4490c73bbe92830/industrial_ci/src/isolation/docker.sh#L20

And why would we mount them rw?
This is not advisable for running CI locally.

mathias-luedtke:
I think we could just add `.gitconfig`  to https://github.com/ros-industrial/industrial_ci/blob/3ed9846c96ed1e0bb36193e8e250632eaac980d0/industrial_ci/src/isolation/docker.sh#L20

mathias-luedtke:
@TrevorGibson-SR: The latest version should copy `.gitconfig` and `.ssh`. Is there anythings else needed for multiple SSH keys?

mathias-luedtke:
To answer my question:

> Is there anythings else needed for multiple SSH keys?

`webfactory/ssh-agent` does not configure the host keys, #902 just uses the known hosts from the GitHub runner.
In addition the generated config does not work inside docker, because of the hard-coded paths, so it needs some patch:
https://github.com/mathias-luedtke/sshtest/blob/main/.github/workflows/main.yml#L46

Issue #817, opened 2023-02-21
Support for ROS-Independent builds using ICI and the ROS toolchain


marip8:
I'm working on a few projects ([`tesseract`](https://github.com/tesseract-robotics/tesseract), [`tesseract_planning`](https://github.com/tesseract-robotics/tesseract_planning), [`descartes_light`](https://github.com/swri-robotics/descartes_light), etc.) that are independent of ROS but use the ROS dependency management and build tools (i.e. `vcstool`, `rosdep`, `colcon`). 

Is it possible to build a ROS-independent project using `industrial_ci` without sourcing `/opt/ros/<distro>` (given that all dependencies are system dependencies, not ROS package dependencies) and without referencing a specific ROS distro? If not where would I start looking in this repository to add this capability? 

At the end of the day, I'm interested in building/testing these repositories and creating a ROS-independent Docker image from the builds

For reference, I implemented [a custom job to do this for the `boost_plugin_loader` repository](https://github.com/tesseract-robotics/boost_plugin_loader/blob/d009f2bdbae2dd94c1ae32116ee1dbf1e98fb1a6/.github/workflows/ubuntu_focal.yml). I want to implement the same thing (and with chained workspaces and other `industrial_ci` features) for the packages mentioned above, but it seems like this custom approach is reinventing `industrial_ci` to some extent

PR #818, opened 2023-03-02
Support ROS-independent CI Builds

marip8:
Addresses #817 to allow CI builds that do not source `/opt/ros/<distro>` during the build.

Here is [an example CI run](https://github.com/marip8/opw_kinematics/actions/runs/4318010420/jobs/7535783448) where this capability was used


mathias-luedtke:
IMHO this is out of scope of industrial_ci.
The scripts assume a ROS environment to function properly.

industrial_ci/src/tests/source_tests.sh
```diff
@@ -172,7 +172,7 @@ function run_source_tests {
 
     ici_step "setup_rosdep" ici_setup_rosdep
 
-    extend=${UNDERLAY:?}
+    extend=${UNDERLAY:-}
```
mathias-luedtke, line 175: This knocks out one of the sanity tests.

industrial_ci/src/workspace.sh
```diff
@@ -406,7 +406,7 @@ function ici_source_setup {
     local extend=$1; shift
     if  [ ! -f "$extend/setup.bash" ]; then
         if [ "$extend" != "/opt/ros/$ROS_DISTRO" ]; then
-            ici_error "'$extend' is not a devel/install space"
+            ici_log "'$extend' is not a devel/install space; skipping source"
```
mathias-luedtke, line 409: Again, this is an important sanity check to test against misspelled parameters.

industrial_ci/src/ros.sh
```diff
@@ -88,6 +88,10 @@ function _set_ros_defaults {
         _ros2_defaults "jammy"
         ;;
     "false")
+        export BUILDER=${BUILDER:-colcon}
+        export ROS_VERSION=0
+        export ROS_VERSION_EOL=false
+        export ROS_PYTHON_VERSION=${ROS_PYTHON_VERSION:-3}
```
mathias-luedtke, line 94: Those are questionable defaults.

mathias-luedtke (commented):
I don't want to sacrifice the sanity checks for an edge case, which is not really supported..

marip8:
> IMHO this is out of scope of industrial_ci.

Maybe at the moment, but I'm proposing this PR to support these types of builds as a feature of `industrial_ci` rather than a convenient hack. 

At least for ROS-I Americas, I think ROS-independent builds are going to become more common as we separate the core functionality of our tools from the ROS interfaces that make them convenient to use. As stated in #817, I personally already have several ROS-I related projects that could leverage this type of build today, with more planned in the near future. We could certainly gather feedback from others if there is a question about general interest in this capability as a feature. In my opinion, there is a ton of value in the ICI pipeline: nested builds, docker image creation, testing, dependency installation, etc. I would rather not re-implement this whole pipeline elsewhere when it only requires very minor modifications to support the builds I am interested in. 

Regarding the sanity checks, I understand the concern. What would be an appropriate way to keep them while being able to keep these added features (i.e., no `UNDERLAY` required and no sourcing `/opt/ros/$ROS_DISTRO`)? An additional variable(s ) that must be explicitly set to enable a ROS-independent build?

industrial_ci/src/ros.sh
marip8: What are your suggested alternatives? Python2 is EOL, so Python3 seems reasonable. `colcon` is the only builder I am familar with that can build both ROS1 and ROS2 code. The ROS version cannot be EOL since this is for a ROS-independent build. `ROS_VERSION` is required by other parts of ICI, so anything < 1 seems okay

mathias-luedtke:
What is so bad about sourcing /opt/ros?
In your example (https://github.com/marip8/opw_kinematics/pull/1) you even install the ROS repositories manually..

mathias-luedtke:
https://github.com/ros-industrial/industrial_ci/pull/834 should be sufficient, it breaks ros_prerelease tests, but I will have to come up with a better implementation for those anyway

marip8:
> What is so bad about sourcing /opt/ros?
In your example (https://github.com/marip8/opw_kinematics/pull/1) you even install the ROS repositories manually..

Mostly because I want to do builds where ROS isn't actually installed, so there is no `/opt/ros/...` to source. In the `opw_kinematics` example, I don't actually install ROS. I'm just using some ROS ecosystem tools like `colcon` to build and `rosdep` to install dependencies that are distributed on Ubuntu (i.e. from `rosdep/base.yaml`). I still have to add the ROS source list and keys in order to install these tools, but ROS itself was never installed.

Ryanf55:
The whole point of colcon, ament, and vcs is they are separate from ROS. Perhaps an alternative method is to contribute the actions directly to those repositories?IE - VCS has a clone action, and colcon has a build action and a test action?

PR #823, opened 2023-04-25
Added support for iron irwini

fmauch:
This PR should add support for the upcoming Iron Irwini release. Unfortunately, this currently fails, since with an installed iron sourcing `/opr/ros/iron/setup.bash` currently results in `$ROSDISTRO=rolling` and then `rosdep install` fails, since the rosdistro is [explicitly passed](https://github.com/ros-industrial/industrial_ci/blob/4b78602d67127a63dce62926769d9ec4e2ce72e4/industrial_ci/src/workspace.sh#L335) to `rosdep update`.

I guess this should be everything that needs to be adapted here, hence I'm opening this PR.


fmauch:
Any comments on this? It would be nice to prepare Github workflows using Iron before the first Iron release :-)

130s (approved):
I'm not an expert but nothing in this PR raise a flag to me. @mathias-luedtke?

mathias-luedtke (approved):

mathias-luedtke:
Sorry, I missed this one..

PR #834, opened 2023-06-02
Support ROS_DISTRO=false

mathias-luedtke:


mathias-luedtke:
@marip8: please test it

marip8:
[Output from a test run](https://github.com/marip8/opw_kinematics/actions/runs/5182405969/jobs/9339158308?pr=1). Looks like `PYTHON_VERSION_NAME` needs to be set

image

mathias-luedtke:
>  Looks like PYTHON_VERSION_NAME needs to be set

Good catch!
I have not tested it with `DOCKER_IMAGE`.
I removed the special handling, lets see if the tests pass..

marip8:
Now it looks like the `ROS_VERSION_EOL` variable also needs to be set.

image

mpowelson:
Has there been any update on this and/or #818?

marip8:
We ended up creating a [new Github CI action](https://github.com/tesseract-robotics/colcon-action) to do the types of builds proposed #818. It turns out that we were previously only using a small fraction of the `industrial_ci` capabilities anyway, so it was much easier to create this custom Github action than to modify `industrial_ci` with the desired features

mpowelson:
I understand the reasoning, but having to roll your own seems like a non-ideal solution. @mathias-luedtke, do you plan to add this capability? I'm specifically interested in using this for Bitbucket, so I wouldn't be able to use the Tesseract github action directly.

PR #842, opened 2023-11-06
Use shallow import with VCS

marip8:
Changes per commit message to reduce overall setup time (and resultant docker image size) when working with large projects.

### DIFF
--- a/industrial_ci/src/workspace.sh
+++ b/industrial_ci/src/workspace.sh
@@ -177,7 +177,7 @@ function ici_setup_git_client {
 }
 
 function ici_vcs_import {
-    ici_guard vcs import --recursive --force "$@"
+    ici_guard vcs import --recursive --force --shallow "$@"
 }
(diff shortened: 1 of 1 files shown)


marip8:
Any ideas why the ABI check fails with `vcs import --recursive --force --shallow /root/base_ws/src`? I don't see anywhere in the `vcstool` docs that says the `recursive` and `shallow` options are mutually exclusive

mathias-luedtke:
> Any ideas why the ABI check test fails

That test checks the ABI against a previous commit and it does not exist in a shallow clone..

mathias-luedtke (changes requested):
This might break existing setups (just as it breaks the ABI test).
Not sure, if we should make it an opt-in or opt-out behavior..

`VCS_IMPORT_SHALLOW`?

It must get set to false, if `ABICHECK_MERGE=true`

PR #843, opened 2023-11-20
Updated ROS Snapshots GPG key

peci1:
Ref. https://github.com/osrf/docker_images/issues/697 .

I'm not sure this is the right ASC file. I've downloaded the 4th key from http://keyserver.ubuntu.com/pks/lookup?search=4B63CF8FDE49746E98FA01DDAD19BAB3CBF125EA&fingerprint=on&op=index .


mathias-luedtke:
Thank you very much for preparing this!

> I'm not sure this is the right ASC file. I've downloaded the 4th key 

I believe all the links download the same file / key :D

mathias-luedtke (approved):

peci1:
Thanks for merging. Melodic CI now works again for me.

mikaelarguedas:
FYI the ROS images on dockerhub have been rebuilt with the updated GPG key.
More details at https://github.com/osrf/docker_images/issues/697#issuecomment-1856463789

PR #854, opened 2024-03-19
Ensure that builder_test_results is exposed as an output

### LINKED ISSUE #853: Setting github output for failed `colcon test-result` broken
Hi,

after debugging a bit it seems to me that output reporting for a failed `colcon test-result` is broken.

I have setup a [minimal project](https://github.com/fmauch/simple_package) with a small C++-"library" and a gtest.

In the [workflow](https://github.com/fmauch/simple_package/blob/ce6f6cf173c86abec560dea9eabf07fd632fc898/.github/workflows/ici.yml#L23-L30) I run ICI and print some ici outputs:
```yaml
      - uses: 'ros-industrial/industrial_ci@master'
        env:
          ROS_DISTRO: ${{ matrix.ROS_DISTRO }}
          ROS_REPO: ${{ matrix.ROS_REPO }}
        id: ici
      - name: Print ICI outputs
        if: ${{ always() }}
        run: echo 'build_target_workspace is ${{ steps.ici.outputs.build_target_workspace }}, run_target_test is ${{ steps.ici.outputs.run_target_test }}, target_test_results is ${{ steps.ici.outputs.target_test_results }}'
```

When everything succeeds, this prints
```
build_target_workspace is 0, run_target_test is 0, target_test_results is 0
```

However, when I add a deliberate failure to the gtest the [result](https://github.com/fmauch/simple_package/actions/runs/8319027481/job/22761727148#step:4:5) is
```
build_target_workspace is 0, run_target_test is 0, target_test_results is 
```

I verified that running [the same commands](https://github.com/ros-industrial/industrial_ci/blob/6931750a64c88d171e23b5397c6c14231acc27cd/industrial_ci/src/workspace.sh#L402-L403) (stripped by ici wrapping) locally
```
err=0
colcon test
colcon test-result --verbose || err=$?
echo "Test result: $err"
```

prints `Test result: 1` in the end so I'm not sure whats going on here.

mathias-luedtke:
Fixes #853 

@fmauch: This should do the trick, but I would rather fix it in a cleaner way (tomorrow?)..

### DIFF
--- a/industrial_ci/src/util.sh
+++ b/industrial_ci/src/util.sh
@@ -233,6 +233,8 @@ function ici_teardown {
             local color_wrap=${ANSI_GREEN}
             if [ "$exit_code" -ne "0" ]; then color_wrap=${ANSI_RED}; fi  # Red color for errors
             ici_time_end "$color_wrap" "$exit_code"
+        elif [ -n "${ICI_RESULT_NAME:-}" ]; then
+            ici_report_result "$ICI_RESULT_NAME" "${exit_code}"
         fi
 
         exec {__ici_log_fd}>&-
--- a/industrial_ci/src/workspace.sh
+++ b/industrial_ci/src/workspace.sh
@@ -400,8 +400,7 @@ function ici_test_workspace {
     local err=0
 
     ici_step "run_${name}_test" builder_run_tests "$extend" "$ws"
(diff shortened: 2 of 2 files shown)


mathias-luedtke:
I might work on a cleaner fix, but for now it is more important to fix the issues

PR #860, opened 2024-04-25
Add jazzy

christophfroehlich:
Jazzy core packages are already available on testing repo.
Some workflows succeed already https://github.com/ros-controls/realtime_tools/actions/runs/8833673702/job/24253543049?pr=165


mathias-luedtke:
Thank you very much!

PR #871, opened 2024-08-04
Add jazzy to documentation and distro jobs

mathias-luedtke:


mathias-luedtke:
Extends https://github.com/ros-industrial/industrial_ci/pull/860

Issue #867, opened 2024-08-01
Prerelease error with Jazzy


johnwason:
I am seeing an error when using `PRERELEASE` set to true with Jazzy. The other distributions are working normally except for the problem described in #866. The non-prerelease is working normally.

Full log: https://github.com/johnwason/robotraconteur/actions/runs/10192639718/job/28195853746

Relevant log section:

```
Step 9/26 : RUN useradd -u 1000 -m buildfarm
   ---> Running in bc3de201b03a
  useradd: UID 1000 is not unique
   ---> Removed intermediate container bc3de201b03a
  The command '/bin/sh -c useradd -u 1000 -m buildfarm' returned a non-zero code: 4
```

mathias-luedtke:
Thanks for reporting!
If if I am not mistaken, this is actually an issue in https://github.com/ros-infrastructure/ros_buildfarm/
I will have a closer look!

mathias-luedtke:
See https://github.com/ros-infrastructure/ros_buildfarm/issues/1060 as well.

mathias-luedtke:
@johnwason: I deployed a work-around, please confirm that this fixes your build

johnwason:
The prerelease tests are working now. Thanks!

PR #872, opened 2024-08-04
Add CI job for jazzy prerelease

mathias-luedtke:


mathias-luedtke:
Work-around for https://github.com/ros-infrastructure/ros_buildfarm/issues/1060
Fixes #867

PR #877, opened 2024-08-20
Install llvm repository for newer versions of clang-format

dave992:
Using `CLANG_FORMAT_VERSION` we can specify which `clang-format` version to use. This assumes the version is available via apt.

I wanted to be able to use more recent versions of `clang-format` uniformly across multiple different ROS versions. Being limited to what is available for the base image is somewhat restrictive in this case. Furthermore, the packages that are included in Ubuntu are quite old.

The proposed changes check if the requested version is available via apt. If not, it will install the requested version using the _Automatic installation script_ recommended by https://apt.llvm.org/


mathias-luedtke:
This seems to be very specific to your use case.
It might be better to prepare a custom docker image or a script hook.

> the Automatic installation script recommended by https://apt.llvm.org/

This pulls more dependencies in, which are not needed for running `clang-format` and slow the process down.

PR #879, opened 2024-09-02
Update ros/actionlib

mosfet80:
updated ros actionlib from version 1.11.11 to  1.14

Python 3 support
https://github.com/ros/actionlib/pull/131

Some fix:
Noetic fixup spurious wakeup
roscpp ActionClient subscription queue size should be consistent with rospy. Fix python wait_for_server premature return


mathias-luedtke:
> Some fix:
Noetic fixup spurious wakeup
roscpp ActionClient subscription queue size should be consistent with rospy. Fix python wait_for_server premature return

It is getting tested on `kinetic` ;)

mathias-luedtke:
The whole point of this test case is the merge commit check.
No need to change it.

PR #895, opened 2024-12-29
Fix ros_canopen versions main.yml

mosfet80:
fixes in git  ros canopen version mismatches.


mathias-luedtke:
@mosfet80: why does this need patching? I chose those version on purpose..

PR #899, opened 2025-01-17
Update actions/cache@v2 -> v4

### LINKED ISSUE #898: Deprecated actions/cache@v2
Running something very similar to this github example

https://github.com/ros-industrial/industrial_ci/blob/master/doc/industrial_ci_action.yml

I get:

```

Your workflow is using a version of actions/cache that is scheduled for deprecation, actions/cache@v2. Please update your workflow to use either v3 or v4 of actions/cache to avoid interruptions. Learn more: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down

```

AndyZe:
Fixes #898


mathias-luedtke (approved):
Thanks for taking care of this!

PR #904, opened 2025-03-05
Add configurable docker entrypoint

pgehring:
This PR introduces two new environment variables: `DOCKER_ENTRYPOINT` and `DOCKER_ADDITIONAL_ENVS`. These variables enable the use of custom Docker images specified in `DOCKER_IMAGE` while ensuring their entrypoint.sh script is executed.

Additionally, `DOCKER_ADDITIONAL_ENVS` allows passing a list of custom environment variables to the running container. These variables are included in the docker create command.

Looking forward to feedback on this feature!


industrial_ci/src/isolation/docker.env
```diff
@@ -22,6 +22,8 @@ CPPFLAGS
 CXX
 CXXFLAGS
 DEBUG_BASH
+DOCKER_ADDITIONAL_ENVS
+DOCKER_ENTRYPOINT
```
mathias-luedtke: Docker variable should [not] be listed here..

industrial_ci/src/isolation/docker.sh
```diff
@@ -15,9 +15,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+export DOCKER_ADDITIONAL_ENVS=${DOCKER_ADDITIONAL_ENVS:-''}
 export DOCKER_COMMIT=${DOCKER_COMMIT:-}
 export DOCKER_COMMIT_MSG=${DOCKER_COMMIT_MSG:-}
 export DOCKER_CREDENTIALS=${DOCKER_CREDENTIALS-.docker .gitconfig .ssh .subversion}
+export DOCKER_ENTRYPOINT=${DOCKER_ENTRYPOINT:-''}
```
mathias-luedtke, line 22: To keep it in line with the other variables
```suggestion
export DOCKER_ENTRYPOINT=${DOCKER_ENTRYPOINT:-}
```

mathias-luedtke (changes requested):
Thanks for your contribution!

The support for DOCKER_ENTRYPOINT can be added right away if you address my comments and add it to https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst

However, I am not convinced of `DOCKER_ADDITIONAL_ENVS`.
The same can be achieved with `DOCKER_RUN_OPTS=-e MYENV`.
If you prefer the "syntactic sugar", then better implement it around https://github.com/ros-industrial/industrial_ci/blob/16472786c1ffc53596007da4318db98d1f677933/industrial_ci/src/isolation/docker.sh#L99

pgehring:
This PR has been completely forgotten. 
@mathias-luedtke if you still think this is a valid feature, would be happy if it will be merged. I've addressed your comment on `DOCKER_ENTRYPOINT`.
`DOCKER_ADDITIONAL_ENVS` is dropped for now. Haven't thought of the `DOCKER_RUN_OPTS`.

industrial_ci/src/isolation/docker.env
```diff
@@ -22,6 +22,7 @@ CPPFLAGS
 CXX
 CXXFLAGS
 DEBUG_BASH
+DOCKER_ENTRYPOINT
```
mathias-luedtke: No need to add it here. Not much use for DOCKER_ENTRYPOINT inside the docker containers..

mathias-luedtke (changes requested):

mathias-luedtke:
> if you still think this is a valid feature, would be happy if it will be merged

I do.
I will try to add a tests for this new feature.

mathias-luedtke:
> I will try to add a tests for this new feature.

I did not have time yet.. I will merge it as-is.

mathias-luedtke (approved):

PR #906, opened 2025-04-25
Add support for ROS Kilted Kaiju

fmauch:
I know, it might be a bit early given that there is no main repo and no documentation branch for kilted, but it would be nice to be able to run CI pipelines against kilted testing.


mathias-luedtke:
Thanks!

> it might be a bit early given that there is no main repo 

That's no problem at all! :)

PR #907, opened 2025-05-16
Set ubuntu distro into main.yml

mosfet80:


.github/workflows/main.yml
```diff
@@ -47,7 +47,7 @@ jobs:
           - distro: rolling
             ubuntu: jammy
           - distro: rolling
-            ubuntu:
+            ubuntu: ubuntu-latest
```
mathias-luedtke, line 50: This is not even a valid code name.
It should be an empty string, which selects the default..

PR #913, opened 2025-06-05
Update snapshots key

augustelalande:
Add the new snapshots key


mathias-luedtke (approved):
Thanks!
I just took the liberty to copy the file from https://keyserver.ubuntu.com/pks/lookup?op=get&search=0xad19bab3cbf125ea directly (which is how I imported it initially).

PR #914, opened 2025-06-05
Handle expired snapshots key

mathias-luedtke:


mathias-luedtke:
Does not work for noetic, but all legacy docker images (melodic and older)

PR #922, opened 2025-12-28
Fix canopen version into  main.yml

mosfet80:
Fix canopen version into  main.yml


mathias-luedtke:
This change would make the ABI tests check against the same version. We could consider this as an additional test, but please keep the original tests untouched.

PR #924, opened 2026-01-12
Fix cmake deprecation

mosfet80:
Cmake version < then 3.10 is deprecated.

### DIFF
--- a/industrial_ci/CMakeLists.txt
+++ b/industrial_ci/CMakeLists.txt
@@ -1,9 +1,7 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.10)
 project(industrial_ci)
 
 if("$ENV{ROS_VERSION}" EQUAL "2")
-    cmake_policy(VERSION 3.5)
-
     find_package(ament_cmake REQUIRED)
     find_package(ament_cmake_python REQUIRED)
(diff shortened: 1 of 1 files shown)


mathias-luedtke:
I am not sure why CMake < 3.10 would be deprecated..
CMake  2.8.12 is needed for Ubuntu trusty / ROS1 indigo.

mosfet80:
> .8.12 is needed for Ubuntu trusty / ROS1 indigo.
Cmake 3.9 and lower is no longer supported on newer operating systems.

Ros1 has been unsupported for a year. Here's the link.
https://www.ros.org/blog/noetic-eol/

In particular ros indigo is no longer supported since 2019

Maintaining backward compatibility with obsolete libraries exposes systems to vulnerabilities known and fixed years ago.
What can be done is to leave main in development and create a ros1 branch for deprecated software.

mathias-luedtke:
> Ros1 has been unsupported for a year. Here's the link.

But industrial_ci still [supports](https://github.com/ros-industrial/industrial_ci/blob/master/doc/index.rst#supported-ros-distributions) old distros on a best-effort basis..

> Maintaining backward compatibility with obsolete libraries exposes systems to vulnerabilities known and fixed years ago.
What can be done is to leave main in development and create a ros1 branch for deprecated software.

There is nothing to be maintained..
Technically the package.xml and CMakeLists.txt are only used by ROS1 tests.

PR #937, opened 2026-06-02
Override ROS_REPO=testin instead of ROS_REPO=main for rolling in resolute

mathias-luedtke:


industrial_ci/src/ros.sh
```diff
@@ -176,6 +181,7 @@ function _get_prefix() {
 function ici_set_ros_repository_path {
     local current_repository_path=$1; shift
     if [ -z "${ROS_REPOSITORY_PATH}" ]; then
+        echo ROS_REPO=${ROS_REPO}
```
mathias-luedtke, line 184:
```suggestion
```

PR #938, opened 2026-07-09
Fix cmake deprecation

mosfet80:
Update CMake minimum version to 3.10
cmake version < then 3.10 is deprecated


mosfet80:
@mathias-luedtke this is a fix for modern OS

